PR #462, opened 2020-11-17
Crypto payments

### LINKED ISSUE #389: Implement crypto payments
Users should be able to make subscription payments in predetermined stablecoin tokens

Flow:
1. Connect a wallet (if one is not yet connected)
2. Call Billing API to notify that a payment for User X is to be expected from the currently connected wallet (may be different from the wallet the user uses to log in to the system. This call should return some form of Id that can be used to query the status of the transcation (from the API perspective)
3. Make the transfer
4. Poll the API for transaction status.

Designs to be confirmed with @sweetpea22

RyRy79261:
closes #389


RyRy79261:
So, We're coming up on the 1 year anniversary of this PR :') 

I'm not sure if its usable when Crypto payments is ready

FSM1:
> So, We're coming up on the 1 year anniversary of this PR :')
> 
> I'm not sure if its usable when Crypto payments is ready

Yeah this can be nuked in my mind... the actual crypto implementation is entirely different from the hack that we had described in the original issue

PR #1048, opened 2021-05-21
Sharing file browser

### LINKED ISSUE #1043: Shared folder File Browser
The Share File Browser should be rendered at `/share/{bucket_id}/{path}`

We need to update the FileBrowser component to be cognisant of the user's ACL rights to the bucket (hide options that the user is not able to use). Owners and writers can perform any file system operations, whereas readers are only presented options such as `info` and `download`

### LINKED ISSUE #1157: Upload file to shared folder
In the shared folder navigation #1043 we should allow users with the `write` permission to upload new files.

### LINKED ISSUE #1162: Remove files from shared folder
This button in the menu should only visible for user with the `write` access

RyRy79261:
closes #1043 
closes #1157 
closes #1162


Tbaut:
Oops I meant the other one sorry.

Tbaut:
@RyRy79261 can you check the alert from lgtm?
image

RyRy79261:
@Tbaut Will take a look, need to update something with the avatars too so will update label later

packages/files-ui/src/Components/Modules/FileBrowsers/Sharing/ShareFileBrowser.tsx
```diff
@@ -0,0 +1,297 @@
...
+  const { filesApiClient } = useFilesApi()
+  const { addToastMessage } = useToaster()
+  const [loadingCurrentPath, setLoadingCurrentPath] = useState(false)
+  const [pathContents, setPathContents] = useState<FileSystemItem[]>([])
+  const { redirect } = useHistory()
+  const { pathname } = useLocation()
+
+  const currentPath = useMemo(() => {
+    const moduleRemoved = extractFileBrowserPathFromURL(pathname, ROUTE_LINKS.ShareExplorer("", "/"))
+    // TODO fetch contents for bucketId
```
Tbaut: Can we please create an issue for each of these 2 todos? I to understand what they really mean.

packages/files-ui/src/Components/FilesRoutes.tsx
```diff
@@ -17,15 +18,17 @@ export const ROUTE_LINKS = {
...
+  Search: (rawSearchTerm: string) => `/search/${rawSearchTerm}`,
   ApplyCryptography: "https://medium.com/chainsafe-systems/major-improvement-to-chainsafe-files-ab489d3e52a2",
   Settings: `${SETTINGS_BASE}/:path`,
   SettingsDefault: `${SETTINGS_BASE}`,
   PurchasePlan: "/purchase",
   UserSurvey: "https://shrl.ink/kmAL",
   GeneralFeedbackForm: "https://shrl.ink/gvVJ",
-  SharedFolders: "/shared"
+  SharedFolders: "/shared",
+  ShareBrowserRoot: "/share",
```
Tbaut: I think this should be `/shared` to be consistant. We agreed with @sweetpea22 that we should talk about shared folders (aka buckets) and then shared files (inside those folders). And never talk about `share` that is vague.

packages/files-ui/src/Contexts/FileBrowserContext.tsx
```diff
@@ -3,8 +3,12 @@ import React, { useContext } from "react"
 import { FileOperation, IBulkOperations, IFileBrowserModuleProps } from "../Components/Modules/FileBrowsers/types"
 import { FileSystemItem, UploadProgress } from "./FilesContext"
 import { Bucket } from "@chainsafe/files-api-client"
+
+export type AccessRole = "reader" | "owner" | "writer" | "none"
```
Tbaut: This is already appearing in the Files context as `BucketPermission` (there's no "none") there though, is that needed? Either case I'd export it and use this one.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemTableItem.tsx
```diff
@@ -236,6 +245,15 @@ const FileSystemTableItem = React.forwardRef(
             <TableCell align="left">
               {!isFolder && formatBytes(size)}
             </TableCell>
+            {
```
Tbaut: This can be removed as discussed.

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFolderRowWrapper.tsx
```diff
@@ -64,24 +17,19 @@ const useStyles = makeStyles(({ breakpoints, constants }: CSFTheme) => {
...
-      "& svg": {
-        fill: constants.fileSystemItemRow.dropdownIcon
-      }
     }
   })
 })
 
 interface Props {
   bucket: Bucket
+  openShare: (bucketId: string) => void
```
Tbaut: I suggest renaming to `openSharedFolder`

Tbaut (commented):
This is working damn well! Went through the code quickly, left a couple comments there, and what I came across:

- Getting this in the console
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `FileSystemTableItem`. See https://fb.me/react-warning-keys for more information.
    in span (at FileSystemTableItem.tsx:252)
    in FileSystemTableItem (at FileSystemItem.tsx:413)
    in FileSystemItem (at SharesList.tsx:905)
```

- When moving a file, the tree is not the right one (it's the one of not shared folders). (edit, made an issue out of it https://github.com/ChainSafe/files-ui/issues/1204)
- As discussed deleting a file should delete it altogether. (made an issue out of it https://github.com/ChainSafe/files-ui/issues/1203)

packages/files-ui/src/Components/FilesRoutes.tsx
FSM1: There were some issues with the routing failing to pick up the difference between `/shared/{id}/` vs `/shared/`, which is why I made them slightly different.

Tbaut: Please use `shared-folder-overview` or anything similar, as discussed in the design channel. "Share" is not a copy we should use.

RyRy79261: @Tbaut So you'd like the over view to be `shared-with-me` maybe and then when viewing a bucket go with `shared`?

FSM1: `shared-with-me` is also not 100% valid as all share buckets are shown in this view 🤔

Tbaut: Yeah, what about 'sharing-overview'?

FSM1: `shared-overview` and `shared/{bucketId}`

That works for me :)

Tbaut (approved):
Looking great, all is green, life is beautiful, thanks a lot for this huge one.
Created 2 small bug issues to tackle separately.

packages/files-ui/src/Components/Modules/FileBrowsers/Sharing/ShareFileBrowser.tsx
```diff
@@ -0,0 +1,284 @@
...
+      refreshContents,
+      deleteItems: moveItemsToBin,
+      downloadFile: handleDownload,
+      moveItems,
+      renameItem: renameItem,
+      viewFolder,
+      loadingCurrentPath,
+      showUploadsInTable: false,
+      sourceFiles: pathContents,
+      heading: t`Shared`,
```
FSM1: Should we not use the Bucket Name as the heading here?

packages/files-ui/src/Components/Modules/FileBrowsers/Sharing/ShareFileBrowser.tsx
```diff
@@ -0,0 +1,284 @@
...
+      controls: true,
+      allowDropUpload: access === "writer" || access === "owner",
+      itemOperations,
+      withSurvey: showSurvey && olderThanOneWeek
```
FSM1: Do we want the survey to also appear on the share browser?

FSM1 (approved):
Left 2 small comments, but looks great otherwise.

The SharesList component seems very much identical to the FilesList except for the menu options. We could potentially hoist the logic of getting these up to context, and just pass it in to the FilesList to avoid having the rest of that repeated.

packages/files-ui/src/Components/Modules/FileBrowsers/Sharing/ShareFileBrowser.tsx
Tbaut: good point, I think it can be removed

packages/files-ui/src/Components/Modules/FileBrowsers/Sharing/ShareFileBrowser.tsx
Tbaut: Makes sense, added it

PR #1074, opened 2021-05-31
Disabling hovers on common components

### LINKED ISSUE #297: Disable hover styling on Mobile for all Components
## Description 

On mobile, the hover state is not used and creates artifact styling when interacting with elements, Button for example

tanmoyAtb:
closes #297 

So here, I mainly disabled the `hover` and sometimes `active` attribute styles for below mobile screens,
made some changes so storybook adapts to CSS variables themeswitcher.


tanmoyAtb:
> Hey @tanmoyAtb, this PR has been open for some time now, is there any blocker or anything you need help with?

I had a talk with Ryan few days ago about some confusions, but they are clear now, I'm going to pick this up in a day or two again.

Tbaut (approved):
Just a little codacy error (we should move this to eslint as an error too, it's a warning currently).

Also I'm now questioning how useful it really is to do all this and remove the hover things from mobile. It adds a lot of code and logic, but what's the benefit? I originally thought this issue was open to make sure that we don't use important things with hover, or that we have a click behavior for mobile. But now I realize that this is different. I just don't see the benefit of this :shrug:

tanmoyAtb:
> Just a little codacy error (we should move this to eslint as an error too, it's a warning currently).
> 
> Also I'm now questioning how useful it really is to do all this and remove the hover things from mobile. It adds a lot of code and logic, but what's the benefit? I originally thought this issue was open to make sure that we don't use important things with hover, or that we have a click behavior for mobile. But now I realize that this is different. I just don't see the benefit of this 🤷

Thats still a bit how I feel right now,
But there are edge cases where `:hover` states and `:active` states appear in mobile and shows up a UI thats unintuitive in small screens.

Tbaut:
Can you point me to an example? I'd say we could have the hover disabled then for those specific cases.

tanmoyAtb:
> Can you point me to an example? I'd say we could have the hover disabled then for those specific cases.

One instance is, you can try clicking on a button on storybook and the `:active`. state persists! 
This has been disabled here for mobile,
otherwise @RyRy79261 may be able point us to more examples and why we need this!

Tbaut:
I just checked on a widely used https://react-bootstrap.github.io/components/buttons/ on mobile and the button is also active on mobile. I think if there's a specific behavior that we find weird/not nice, we can address it, but I don't see the value in doing this to all component blindly.

tanmoyAtb:
> I just checked on a widely used https://react-bootstrap.github.io/components/buttons/ on mobile and the button is also active on mobile. I think if there's a specific behavior that we find weird/not nice, we can address it, but I don't see the value in doing this to all component blindly.

I do agree here, 
Doing this for all components add code without adding much value in my opinion.
Particular weird behaviours can be handled separately!

Tbaut:
Ok, then let's close it, and the associated issue. We can still use any code later if needed.

PR #1078, opened 2021-05-31
Shared folder overview

### LINKED ISSUE #1041: Shared Overview page
This page should render at the `/shared` route and list all the `Share` buckets that a user is an `owner`, `writer` or `reader` in. 

The list of buckets can be obtained from `DriveContext.buckets` property

The `Shared Item` dropdown menu and `Create share` modal are excluded from scope of this ticket

image

Tbaut:
closes #1041


FSM1 (approved):

packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -94,29 +100,64 @@ const FilesProvider = ({ children }: FilesContextProps) => {
...
+          : b.writers.find(writer => writer === profile?.userId)
+            ? "writer" as BucketPermission
+            : b.readers.find(reader => reader === profile?.userId)
+              ? "reader" as BucketPermission
+              : undefined
+
+        let encryptionKey = ""
+
+        switch(b.type) {
+        case "csf":
```
tanmoyAtb, line 140: question! any particular reason, not getting the key for csf ?

tanmoyAtb:
One thing I noticed, I don't see uploads showing progress (downloads progress working). Not sure if its this PR!

tanmoyAtb (approved):
The PR looks smooth and ready to go :D,
There are some UI parts, specially the mobile view,
But thats for another issue.

packages/files-ui/src/Contexts/FilesContext.tsx
FSM1: Each case option will fall through to the nearest implementation, so syntactically this is equivalent to :
```
case 'a': {
  // do thing
  break;
}
case 'b': {
  // do thing
  break
}
```

Tbaut: yup, we do get the key for `csf`, but we get the key in a similar way for both `trash` and `csf` so I grouped them.

Tbaut:
Yup so just like I did for the Storage part, my goal was to have a working base PR for all the other issues to be tackled in parallel. We should def, add one for the mobile view because I didn't play with it at all and I def. expected things to break. For the upload, I guess you mean to add things to shared folder? This is also something that can be worked on in parallel. I thought there was an issue for that, but I'll add it now too.

PR #1133, opened 2021-06-14
Add username

### LINKED ISSUE #1091: Add Nickname field to the profile overview
Allow user to set nickname

tanmoyAtb:
closes #1091


packages/common-components/src/TextInput/FormikTextInput.tsx
```diff
@@ -21,14 +22,15 @@ const FormikTextInput = React.forwardRef(
...
+      hideLabel,
       ...rest
     }: FormikTextInputProps,
     forwardedRef: any
   ) => {
     const [field, meta, helpers] = useField(name)
     return (
       <TextInput
-        label={label ? label : field.name}
+        label={label ? label : hideLabel ? undefined : field.name}
```
Tbaut: This is convoluted. Would this make more sense?
```suggestion
        label={hideLabel ? undefined : label || field.name}
```

packages/common-components/src/TextInput/TextInput.tsx
```diff
@@ -396,7 +396,7 @@ export interface ITextInputProps
   inputVariant?: "default" | "minimal"
   size?: "large" | "medium" | "small"
   captionMessage?: string | ReactNode
-  onChange: (value: string | number | undefined) => void
+  onChange?: (value: string | number | undefined) => void
```
Tbaut: with ?
```suggestion
  onChange?: (value?: string | number) => void
```

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -197,9 +240,46 @@ const ProfileView = () => {
   const profileValidation = yup.object().shape({
     // email: yup.string().email("Email is invalid").required("Email is required"),
     firstName: yup.string(),
-    lastName: yup.string()
+    lastName: yup.string(),
+    username: yup.string()
   })
 
+  const onLookupUsername = async (username: string) => {
```
Tbaut: We're not catching errors here. Also with 1 await, a promise.then.catch may work just as well.

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -197,9 +240,46 @@ const ProfileView = () => {
...
+        loading: false,
+        error: ""
+      })
+    }
+  }
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const debouncedOnLookupUsername = useCallback(debounce(onLookupUsername, 200), [lookupOnUsername])
+
+  const onUsernameChange = async (value: any) => {
```
Tbaut: no need for async, need for useCallback. And we can avoid the `any` here.

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -197,9 +240,46 @@ const ProfileView = () => {
...
+      })
+    } else {
+      setUsernameData({
+        loading: false,
+        error: ""
+      })
+    }
+  }
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
Tbaut: Something wrong here, shouldn't need to silence it. I played with it and removing the useCallback by a `useMemo(() => debounce(onLookupUsername, 200), [onLookupUsername])` and adding a useCallback to `onLookupUsername` let it work pretty well, but pelase don't take my word for it and double-check that it's debouncing well

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -197,9 +240,46 @@ const ProfileView = () => {
...
+  const debouncedOnLookupUsername = useCallback(debounce(onLookupUsername, 200), [lookupOnUsername])
+
+  const onUsernameChange = async (value: any) => {
+    setUsername(value)
+    debouncedOnLookupUsername(value)
+  }
+
+  const onSubmitUsername = (e: any) => {
```
Tbaut: `(e: React.FormEvent<HTMLFormElement>)`

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -197,9 +240,46 @@ const ProfileView = () => {
...
+    e.preventDefault()
+    setUsernameData({ ...usernameData, loading: true })
+    addUsername(username).then(() => {
+      addToastMessage({ message: t`Username added` })
```
Tbaut: set successfully ?
```suggestion
      addToastMessage({ message: t`Username set successfully` })
```

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -212,6 +292,163 @@ const ProfileView = () => {
...
+                      variant="body1"
+                      component="p"
+                      className={classes.publicAddress}
+                    >
+                      {centerEllipsis(profile.publicAddress, 16)}
+                    </Typography>
+                    <CopyIcon className={classes.copyIcon} />
+                  </div>
+                </div>
+                : null}
```
Tbaut: the ternary should be replaced by `question && <Component>` instead of `question ? <Component> : null`

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -212,6 +292,163 @@ const ProfileView = () => {
...
+                      variant="body1"
+                      component="p"
+                      className={classes.publicAddress}
+                    >
+                      {centerEllipsis(publicKey, 16)}
+                    </Typography>
+                    <CopyIcon className={classes.copyIcon} />
+                  </div>
+                </div>
+                : null}
```
Tbaut: Same here, the `: null` should be removed and use a && instead.

packages/files-ui/src/Contexts/UserContext.tsx
```diff
@@ -14,16 +14,19 @@ export type Profile = {
...
   profile: Profile | undefined
   refreshProfile(): Promise<void>
-  updateProfile(
+  updateProfile: (
     firstName: string,
     lastName: string,
     // email: string,
-  ): Promise<void>
+  ) => Promise<void>
+  lookupOnUsername: (username: string) => Promise<boolean | undefined>
```
Tbaut: undefined not needed (see next comment)
```suggestion
  lookupOnUsername: (username: string) => Promise<boolean>
```

packages/files-ui/src/Contexts/UserContext.tsx
```diff
@@ -88,6 +93,43 @@ const UserProvider = ({ children }: UserContextProps) => {
...
+      return Promise.reject(
+        error && error.length
+          ? error[0].message
+          : "There was an error setting username."
+      )
+    }
+  }
+
+  const lookupOnUsername = async (username: string) => {
+    if (!profile) return
```
Tbaut: return false
```suggestion
    if (!profile) return false
```

packages/files-ui/src/Contexts/UserContext.tsx
```diff
@@ -88,6 +93,43 @@ const UserProvider = ({ children }: UserContextProps) => {
...
+      setProfile({
+        ...profile,
+        username
+      })
+      return Promise.resolve()
+    } catch (error) {
+      return Promise.reject(
+        error && error.length
+          ? error[0].message
+          : "There was an error setting username."
```
Tbaut: translation
```suggestion
          : t`There was an error setting username.`
```

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -212,6 +292,163 @@ const ProfileView = () => {
             className={classes.sectionContainer}
           >
             <div className={classes.profileBox}>
+              <Typography
+                variant="h3"
+                component="h3"
+                className={classes.header}
+              >
+                Profile settings
```
Tbaut: Trans
```suggestion
                <Trans>Profile settings</Trans>
```

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -229,91 +466,42 @@ const ProfileView = () => {
...
-                    </div>
-                  ) : null}
-                  <div className={classes.boxContainer}>
+                      <Trans>Only you can see this.</Trans>
+                    </Typography>
                     <FormikTextInput
                       placeholder={t`First name`}
                       name="firstName"
                       size="medium"
+                      hideLabel={true}
```
Tbaut, line 486: Didn't check deeper but can't we have empty label rather?

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -136,16 +151,44 @@ const useStyles = makeStyles(({ constants, breakpoints, palette, typography }: C
       fontWeight: 400,
       marginTop: 25,
       marginBottom: 14
+    },
+    buttonLink: {
+      color: palette.primary.main,
```
Tbaut: black please, see my comment int he issue. There's no such buttonLink anywhere else in the UI. We need to be consistent :)

packages/files-ui/src/Contexts/UserContext.tsx
```diff
@@ -88,6 +93,43 @@ const UserProvider = ({ children }: UserContextProps) => {
...
+      )
+    }
+  }
+
+  const lookupOnUsername = async (username: string) => {
+    if (!profile) return
+    try {
+      await filesApiClient.lookupUser({ username })
+      return true
+    } catch (error) {
```
Tbaut, line 129:
```suggestion
    } catch (error) {
      console.error(error)
```

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -136,16 +151,44 @@ const useStyles = makeStyles(({ constants, breakpoints, palette, typography }: C
...
-  const { profile, updateProfile } = useUser()
+  const { profile, updateProfile, addUsername, lookupOnUsername } = useUser()
   const { publicKey } = useThresholdKey()
   const [updatingProfile, setUpdatingProfile] = useState(false)
+  const [showUsernameForm, setShowUsernameForm] = useState(false)
+  const [username, setUsername] = useState("")
+  const [usernameData, setUsernameData] = useState({
+    error: "",
+    loading: false
+  })
```
Tbaut:
```suggestion
  const [usernameData, setUsernameData] = useState({error: "", loading: false})
```

Tbaut (changes requested):
Working well but I'm getting some 404 on `https://stage.imploy.site/api/v1/user/lookup` when typing something (although the lookup does work).

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
tanmoyAtb: the `label` is actually optional, but the `FormikTextInput` passes the `field.name` and I had no way of removing that since its being used else where in code.

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
tanmoyAtb: This works just as perfect! thanks!

tanmoyAtb:
> Working well but I'm getting some 404 on `https://stage.imploy.site/api/v1/user/lookup` when typing something (although the lookup does work).

Yes, the `lookup` api responds with 404 when it doesn't find a username we are sending,
not sure if thats a standard behaviour but it works.

FSM1 (approved):

Tbaut:
> api responds with 404 when it doesn't find a username we are sending,
not sure if thats a standard behaviour but it works.

ok a 404 is definitely not a standard for an expected answer, this is showing red things all over (network and console tab), it is handled as an error in the code and in the browser. I guess it should be a 200 with `null` or something. I'll check with the api team

Tbaut (approved):
:+1: nice one! Just the cla signing is missing :man_shrugging:

tanmoyAtb:
> > api responds with 404 when it doesn't find a username we are sending,
> > not sure if thats a standard behaviour but it works.
> 
> ok a 404 is definitely not a standard for an expected answer, this is showing red things all over (network and console tab), it is handled as an error in the code and in the browser. I guess it should be a 200 with `null` or something. I'll check with the api team

Yes a 200 with null, sounds right!

PR #1143, opened 2021-06-15
UI Tests: bin page - delete a file

### LINKED ISSUE #1086: UI Tests: Add new identifiers, populate the bin page object with element declarations
As part of our test framework development we need  populate the bin page object with all the element selectors 

Acceptance criteria: 

- Applicable new identifiers are added for bin page elements
- Add element declarations to the bin page object
- Add basic test using bin page elements

asnaith:
closes #1086

*What I changed* 

- Added new element identifiers and updated page objects
- Added a basic delete file ui test
- Added the ability to clear the `trash` bucket


asnaith:
Having issues setting up identifier for the delete file confirmation box element, they are currently using weak selectors in homePage.ts : 

deleteFileDialog: () => cy.get(".makeStyles-inner-186 > .makeStyles-root-35"),
deleteFileCancelButton: () => cy.get(".makeStyles-dashed-97"),
deleteFileConfirmButton: () => cy.contains("Confirm")

Screen Shot 2021-06-15 at 9 32 01 AM

I'll also need to add one to this toast message too (at the moment I haven't declared it on the homePage.ts) and I couldn't seem to find out where in BinFileBrowser.tsx to add it. 

Screen Shot 2021-06-15 at 9 35 20 AM

packages/common-components/src/Button/Button.tsx
```diff
@@ -318,6 +320,7 @@ const Button: React.FC<IButtonProps> = ({
       )}
       disabled={disabled || loading}
       {...rest}
+      data-testid={`${testId}-button`}
```
Tbaut: Super nit, I think for consistency we should have 
```suggestion
      data-testid={`button-${testId}`}
```

packages/files-ui/cypress/support/page-objects/homePage.ts
```diff
@@ -8,9 +8,9 @@ export const homePage = {
   uploadFileForm: () => cy.get("[data-cy=upload-file-form] input"),
   moveSelectedButton: () => cy.get("[data-cy=move_selected_file_button]"),
   deleteSelectedButton: () => cy.get("[data-cy=delete_selected_file_button]"),
-  deleteFileDialog: () => cy.get("[data-testid=file-deletion-modal]"),
-  deleteFileCancelButton: () => cy.get(".makeStyles-dashed-97"),
-  deleteFileConfirmButton: () => cy.contains("Confirm"),
+  deleteFileDialog: () => cy.get("[data-testid=modal-container-file-deletion]"),
+  deleteFileCancelButton: () => cy.get("[data-testid=cancel-deletion-button]"),
+  deleteFileConfirmButton: () => cy.get("[data-testid=confirm-deletion-button]"),
   uploadStatusToast: () => cy.get("[data-cy=upload_status_toast_message]", { timeout: 10000 }),
```
Tbaut: I see this a little late but for consistency we should use `-` instead of `_` (that's just a suggestion for here but it would need to be adapted in the code too)
```suggestion
  uploadStatusToast: () => cy.get("[data-cy=upload-status-toast-message]", { timeout: 10000 }),
```

packages/common-components/src/Button/Button.tsx
asnaith: @Tbaut Ok 👍. Agreed, it's good to keep consistency

packages/files-ui/cypress/support/page-objects/homePage.ts
asnaith: @Tbaut Oh! That was a legitimate mistake, thanks for catching it 👍

Tbaut:
Any reason to close it? We could just iterate until it's good to merge wdyt?

asnaith:
@Tbaut It's incomplete and I thought I'd just create a new one when it was finished instead of triggering test runs each time I push changes. I can re-open if we want to keep it all in here.

packages/files-ui/cypress/support/page-objects/homePage.ts
```diff
@@ -36,7 +41,8 @@ export const homePage = {
     this.uploadButton().click()
     this.uploadFileForm().attachFile(filePath)
     this.fileUploadList().should("have.length", 1)
-    this.startUploadButton().click()
+    this.fileListRemoveButton().should("be.visible")
```
asnaith, line 44: By checking for the appearance of the remove button here, we can be sure that the modal has finished adjusting the layout after attaching a file. It's better to do this before cypress proceeds to click the start upload button.

This should hopefully resolve the intermittent issue described in #1146. I haven't seen that failure since I added this.

Tbaut: Awesome :)

packages/files-ui/cypress/support/utils/apiTestHelper.ts
```diff
@@ -0,0 +1,37 @@
...
+        .then((tokens) => {
+          apiClient.setToken(tokens.access_token.token)
+          apiClient.listBuckets(bucketType).then((buckets) => {
+            apiClient
+              .getBucketObjectChildrenList(buckets[0].id, { path: "/" })
+              .then((items) => {
+                const toDelete = items.map(
+                  ({ name }: { name: string }) => `/${name}`
+                )
+                console.log(`Deleting bucket ${bucketType} ${JSON.stringify(toDelete)}`)
```
Tbaut, line 30: should we change that to a `cy.log`?

FSM1: its not possible to call a `cy.*` method inside a callback. Cypress complains about this very loudly...

packages/files-ui/cypress/support/commands.ts
```diff
@@ -182,18 +160,20 @@ Cypress.Commands.add(
...
       cy.reload()
-      cy.get("[data-cy=files-app-header]", { timeout: 20000 }).should(
-        "be.visible"
-      )
+      cy.get("[data-cy=files-app-header]", { timeout: 20000 }).should("be.visible")
+    }
+
+    if (clearTrashBucket) {
+      cy.clearTrashBucket(apiUrlBase)
+      cy.reload()
```
Tbaut, line 176: if we have clearCSF and clearTrash, we're doing 2 reloads needlessly. I guess we should have a dedicated 
```
if(clearTrashBucket || clearCSFBucket){
    cy.reload()
    cy.get("[data-cy=files-app-header]", { timeout: 20000 }).should("be.visible")
}
```

packages/files-ui/cypress/support/commands.ts
```diff
@@ -182,18 +160,20 @@ Cypress.Commands.add(
...
+      cy.get("[data-cy=files-app-header]", { timeout: 20000 }).should("be.visible")
```
Tbaut, line 177: Should this one be abstracted in the homepage? I've seen it in many placed in the command.ts

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -66,5 +70,37 @@ describe("File management", () => {
       homePage.fileRenameInput().should("not.exist")
       homePage.fileItemName().contains(newName)
     })
+
+    it("can delete a single file", () => {
+      cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })
+
+      // upload a file 
+      homePage.uploadFile("../fixtures/uploadedFiles/text-file.txt")
```
Tbaut, line 80: we need to replace this with a programmatic upload. Creating an issue

Tbaut (approved):
Looks great, a couple nits found but feel free to merge when you're happy with it :)

packages/files-ui/cypress/support/utils/apiTestHelper.ts
Tbaut: ah well, then we'll let it be a console :man_shrugging:

packages/files-ui/cypress/support/page-objects/homePage.ts
asnaith: @Tbaut Party over :( I spoke too soon, I saw it happen again after this. ~~I'll look into it out of the context of this PR~~

packages/files-ui/cypress/support/commands.ts
asnaith: @Tbaut Ah yeah, that's a good point. Thanks for noticing that. I've removed the reload and header check from the other two statements and created the dedicated `if` for it like you suggested

packages/files-ui/cypress/support/commands.ts
asnaith: Yes for sure, all of these direct selector calls need to be replaced with page object properties. I'll create an issue

packages/files-ui/cypress/tests/file-management-spec.ts
asnaith: @Tbaut Yeah, that will be more efficient, happy to skip the UI when that functionality is not what we are testing directly and the file presence is just a pre-requisite for the rest of the test(s)

packages/files-ui/cypress/support/page-objects/homePage.ts
asnaith: Update: I just pushed a small change with waits as we discussed this morning. Including them in this PR as they are only two lines of code.

FSM1 (approved):

Tbaut:
Wasn't it possible to wait for the button to be enabled instead of waiting purely?

asnaith:
> Wasn't it possible to wait for the button to be enabled instead of waiting purely?

@Tbaut Unfortunately not. I have used `.should("not.be.disabled").click()` on the start upload button so in theory cypress should only performed the click once enabled however it's had no effect on the reliability. 

I've just experimented with removing the waits and using `should("be.enabled").click()` too (just in case the detection behaves different from `not.be.disabled`) but no luck there either.

Tbaut:
OK I'll check it out, pretty sure there is something to be done. I guess the disabled isn't checked correctly because cypress expects some specific class or attributes that the button doesn't have.

PR #1160, opened 2021-06-21
Update commands.ts to use page object properties

### LINKED ISSUE #1156: UI Tests: Update commands.ts to use page object properties
We have many selectors directly included in `commands.ts` which is a bit less friendly to read and harder to maintain as things change. Let's update them to use page object properties. 

Example 

`cy.get("[data-cy=files-app-header]", { timeout: 20000 }).should("be.visible")` 

becomes something like 

`homePage.appHeader.should("be.visible")` 

Acceptance Criteria: 

- Direct selectors are replaced with page object properties!

asnaith:
closes #1156


Tbaut (approved):
:100:

RyRy79261 (approved):

Issue #1131, opened 2021-06-11
 CypressError: Timed out after waiting `60000ms` for your remote page to load


asnaith:
We have been experiencing sporadic page load timeouts whilst running ui tests in CI. It doesn't seem to be an issue with a specific test. 

Below is an example: 

https://github.com/ChainSafe/files-ui/pull/1127/checks?check_run_id=2801773985
![Settings -- mobile -- can navigate to the settings security page on a phone -- before all hook (failed)](https://user-images.githubusercontent.com/3982665/121755637-89dc7e00-cacc-11eb-8b79-74e4cc64b781.png)

I have looked into this but haven't found a fix just yet.

Suggestions:

**Improvement A**

I _think_ cypress is sometimes trying to perform an action (page visit) too quickly, before the browser has had chance to respond to the previous commands (window object creation for ethereum wallet) and this is seen more in CI because it's not as powerful as our development machines.  

Perhaps we could add a guard statement, with logging, that checks for the created window object before proceeding to the page visit (not too sure how to write that yet) or at least we could a slight wait statement between the two commands but the former would be better. 

I think it could be a similar situation with #1070 too. Cypress commands running too quickly before the browser can respond to a previous one.

**Improvement B**

After reviewing test.yml / package.json it seems that we are only using `"cypress run"` to execute the tests. According to [this doc](https://docs.cypress.io/guides/guides/launching-browsers#Browsers) that will mean we are using Electron by default in CI. 

We can specify it to use chrome using the command `cypress run --browser chrome` instead. I read online that there are other teams experiencing similar issues with a electron / cypress / ci combo that were resolved by doing this.

Tbaut:
I think we can close this one, we got rid of the readFiles tasks thanks to `cy.session` in #1421

PR #1164, opened 2021-06-21
Make cypress tests use chrome version 90 in ci

asnaith:
Related to "Improvement B" mentioned in #1131

*What I changed* 

- Updated the docker container being used in ci
- Specified cypress to run tests in the chrome browser instead of the default (electron)


asnaith:
Ughhh github automatically closed this because I'd changed a typo in the branch name

PR #1169, opened 2021-06-22
Fix lgtm alerts

Tbaut:
Fixing those alerts https://lgtm.com/projects/g/ChainSafe/files-ui/alerts/?mode=list


packages/common-theme/src/utils/colorManipulator.ts
```diff
@@ -124,20 +124,7 @@ export function decomposeColor(color: string): IColorObject {
...
-
-  if (type.indexOf("rgb") !== -1) {
-    // Only convert the first 3 values to int (i.e. not alpha)
-    valuesArray = values.map((n, i) => (i < 3 ? n : n)) as ColorValues
-  } else if (type.indexOf("hsl") !== -1) {
-    valuesArray[1] = `${values[1]}%`
-    valuesArray[2] = `${values[2]}%`
-  }
-
+export function recomposeColor({ type, values }: IColorObject) {
```
RyRy79261, line 128: This looks like a regression

Tbaut: check it out in details, `valuesArray` is actually never used.

edit: I'm fine to remove this function altogether, I don't think it's used either anyway.

tanmoyAtb (approved):
lgtm!

RyRy79261 (approved):

Issue #1159, opened 2021-06-21
Update lookup for user (will return null instead of 404)


Tbaut:
Following https://github.com/ChainSafe/files-api/issues/1216#event-4915213852
The /user/lookup will now answer a 200 with `null` in case no user was found with this username. We should update it in the UI

Tbaut:
I estimated it unilaterally since I just tackled most of it as part of another issue.

Tbaut:
done in  #1180

PR #1180, opened 2021-06-23
Fix lookup and validation

### LINKED ISSUE #1170: Settings: first character of username text input cannot be changed
**Describe the bug**
User is unable to backspace the first character of username text input after initial entry

**To Reproduce**
1. Log in to Files with an account that has no username set
2. Navigate to settings
3. Navigate to the profile tab
4. Click "Add a username" 
5. Enter a long unique string as input
6. Attempt to remove the input that you typed

**Observed behavior**
It's not possible to remove the first character that you typed, you have to navigate away and back to the tab in order to set the username to something else. 

**Expected behavior**
It should be possible to remove all prior input text 

 - OS: macOS
 - Browser Chrome
 - Version 91

**Additional context**
Dev branch only, set username feature is not available on production.

Tbaut:
closes #1170 and also most of #1159 since the change was merged on the api, just waiting for the api-spec update.
edit: api-spec doesn't need change.


packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -242,42 +242,48 @@ const ProfileView = () => {
...
-          error: ""
-        })
-      }
-    }).catch(console.error)
+    lookupOnUsername(username)
+      .then((doesUsernameExist) => {
+        if (doesUsernameExist) {
+          setUsernameData({
+            loading: false,
+            error: "This username is already taken"
```
tanmoyAtb: missed trans here!
```suggestion
            error: t`This username is already taken`
```

packages/files-ui/src/Contexts/UserContext.tsx
```diff
@@ -116,16 +116,16 @@ const UserProvider = ({ children }: UserContextProps) => {
...
       )
     }
   }
 
   const lookupOnUsername = async (username: string) => {
     if (!profile) return false
     try {
-      await filesApiClient.lookupUser({ username })
-      return true
+      const alreasdyExists = await filesApiClient.lookupUser({ username })
```
tanmoyAtb:
```suggestion
      const alreadyExists = await filesApiClient.lookupUser({ username })
```

packages/files-ui/src/Contexts/UserContext.tsx
```diff
@@ -116,16 +116,16 @@ const UserProvider = ({ children }: UserContextProps) => {
...
+      return !!alreasdyExists
```
tanmoyAtb:
```suggestion
      return !!alreadyExists
```

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
```diff
@@ -242,42 +242,48 @@ const ProfileView = () => {
...
   const debouncedOnLookupUsername = useMemo(
     () => debounce(onLookupUsername, 300),
     [onLookupUsername]
   )
 
-  const onUsernameChange = (value: string) => {
-    setUsername(value)
-    debouncedOnLookupUsername(value)
+  const onUsernameChange = (value: string | number | undefined) => {
+    const sanitizedValue = value?.toString() || ""
```
tanmoyAtb, line 283: This makes me wonder, if we made the same mistake in other inputs!

tanmoyAtb (approved):
Make sure to check the suggestions!

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
Tbaut: Not sure, we can check.

packages/files-ui/src/locales/en/messages.po
```diff
@@ -448,8 +451,8 @@ msgstr "Remind me later"
 msgid "Rename"
 msgstr "Rename"
 
-msgid "Rename File/Folder"
-msgstr "Rename File/Folder"
+msgid "Rename ${0}"
```
Tbaut: pretty sure this won't fly, I'll check it out.

packages/files-ui/src/Components/Modules/Settings/Profile.tsx
asnaith: I checked these inputs in the UI:

- change password inputs
- first and last name inputs
- file rename input (list and grid view)
- folder rename input (list and grid view)
- create folder input

I didn't have a problem removing anything that was entered, none of them exhibited the same issue that was happening on the username input.

RyRy79261 (approved):

PR #1181, opened 2021-06-23
Cancel editing when there's a click outside

### LINKED ISSUE #1171: File Management: User can drag and drop file whilst it is being renamed
**Describe the bug**
User can drag and drop a file before the rename changes are saved

**To Reproduce**
1. Login to files
2. Create a folder in your drive root
3. Upload a file to your drive root
4. Start to rename your file but do not save changes
5. Try to move the file to the folder you created

**Observed behavior**
It is possible to drag and drop the file into the folder whilst the file rename components are still being displayed. If you proceed to move it to the folder then the file will be moved with the original name in place. 

After the file has been moved if you navigate to that folder you will see the rename components on that file are still visible. 

**Expected behavior**
It should not be possible to drag and drop the file whilst it is being renamed (or if we want to allow this perhaps we should  prompt the user to save their changes first). 

 - OS: macOS
 - Browser: Chrome
 - Version : 91

**Additional context**
Happening on dev and current production release

Tbaut:
closes #1171


tanmoyAtb (approved):
👍

asnaith (approved):

RyRy79261 (approved):

PR #1197, opened 2021-06-28
Rename shared folder

### LINKED ISSUE #1130: Shared folder name editing
Allow an Owner to Rename a shared bucket using `FilesApiClient.updateBucket()`

In the Shared Overview page, the buckets that a user owns should have a "Rename" which when clicked will allow the user to rename the share (same as renaming files and folders in the File Browser)

tanmoyAtb:
closes #1130 

rebased to the create shared folder modal PR!


Tbaut (changes requested):
Found a bug when renaming a second time without refreshing, the old name is showing up in the field:

https://user-images.githubusercontent.com/33178835/123783651-e4e8e000-d8d6-11eb-8997-7b636c6adc15.mp4

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/SharedFolderRow.tsx
```diff
@@ -99,9 +109,12 @@ interface Props {
   bucket: Bucket
   onFolderClick: (e?: React.MouseEvent) => void
   menuItems: IMenuItem[]
+  editing: boolean
```
Tbaut: can we call it `isEditing` and the handler `setIsEditing`?

packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -117,10 +117,10 @@ const FilesProvider = ({ children }: FilesContextProps) => {
     return decrypted || ""
   }, [decryptMessageWithThresholdKey, userId])
 
-  const refreshBuckets = useCallback(async () => {
+  const refreshBuckets = useCallback(async (showLoading?: boolean) => {
     if (!personalEncryptionKey || !userId) return
 
-    setIsLoadingBuckets(true)
+    if (showLoading) setIsLoadingBuckets(true)
```
Tbaut: Maybe `showLoading && setIsLoadingBuckets(true)`
```suggestion
    showLoading && setIsLoadingBuckets(true)
```

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/SharedFolderRow.tsx
tanmoyAtb: sure thing

tanmoyAtb:
> Found a bug when renaming a second time without refreshing, the old name is showing up in the field:
> 
>  rename.mp4

good catch! looking into it

tanmoyAtb:
> Found a bug when renaming a second time without refreshing, the old name is showing up in the field:
> 
>  rename.mp4

Cool, had to add `reinitializeOnProps` for this formik, its working now!

Tbaut (approved):
Awesome stuff

FSM1 (changes requested):
Works perfectly for those buckets where the user is an owner.

Just need to add a check on the menu items to ensure that the current user is an owner of the bucket. I think that for other users, the menu dropdown can just be disabled, as I don't think there are currently any actions a reader/writer can perform on a bucket level.

Also found that renaming a shared bucket currently resets the list of readers and writers to empty arrays

tanmoyAtb:
> Works perfectly for those buckets where the user is an owner.
> 
> Just need to add a check on the menu items to ensure that the current user is an owner of the bucket. I think that for other users, the menu dropdown can just be disabled, as I don't think there are currently any actions a reader/writer can perform on a bucket level.
> 
> Also found that renaming a shared bucket currently resets the list of readers and writers to empty arrays

I added the owner check to show menu,
also commented out delete bucket for now

RyRy79261 (approved):

FSM1 (approved):

FSM1 (approved):

PR #1220, opened 2021-07-01
Bump cypress to v1.7

Tbaut:
self explanatory


FSM1 (approved):

tanmoyAtb (approved):
👍 
was wondering why we have dependabot turned off now,
would get automatic updates that way!

Tbaut:
> was wondering why we have dependabot turned off now,

isn't this only for deps that have security vulns? In any case I think it'd be cool.

PR #1221, opened 2021-07-01
Use BucketKeyPermission instead of missleading Bucket

Tbaut:
Renaming the `Bucket` type (that extends `Bucket` from the api-spec) so that our VSC aren't mislead, and so that we can use the `permission` that `BucketKeyPermission` actually has.

Our FilesContext is actually enhancing the `Bucket` type from the api-spec, and adds some more fields such as the encryption key, and the permission for the current user. I've replaced it here with a clearer type `BucketKeyPermission`

small fix along the way, prevent the overkill use of regex.


Tbaut:
Codacy seems to have internal issues :shrug:

FSM1 (approved):

tanmoyAtb (approved):

PR #1223, opened 2021-07-02
Fix bucket filtering in Share browser

FSM1:


tanmoyAtb (approved):
💯 

I'm wondering why aren't we using route params
sth like `/shared/:bucketId`, and then `const { bucketId } = route.params`

FSM1:
> 💯
> 
> I'm wondering why aren't we using route params
> sth like `/shared/:bucketId`, and then `const { bucketId } = route.params`

React Router was being super weird about the complex url structure with the `/{bucketId}/{bucket path}`, so I just went for manually parsing the string

Tbaut (approved):
Those path fiddling are scaring me, we've had many bugs in these, it's a little like cuisine every single time I feel. Wondering if we could have something nice with React router like `{ bucketId, path } = useParams`. Probably not super simple to figure out but I don't see why that shouldn't work.

FSM1:
> Those path fiddling are scaring me, we've had many bugs in these, it's a little like cuisine every single time I feel. Wondering if we could have something nice with React router like `{ bucketId, path } = useParams`. Probably not super simple to figure out but I don't see why that shouldn't work.

Yeah agreed, it would be really nice to have a better defined structure here, but unfortunately params dont play along nicely.

For example if the route we are working with is structured as `/:bucketId/:path` - `https://files/bucket_id/a/path/somewhere` destructuring with useParams will result  `bucketId=bucket_id` and `path=a` with the rest of it being ignored.
image

PR #1225, opened 2021-07-02
Owner column in shared folder overview

### LINKED ISSUE #1129: Be able to spot the shared folders I am the owner of
Shared Folders overview Page

@sweetpea22 we should find out how to let users know right away for what shared folder they are the owner. A different Icon would prob work.

image

tanmoyAtb:
closes #1129


packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/SharedFolderRow.tsx
```diff
@@ -216,6 +216,11 @@ const SharedFolderRow = ({ bucket, onFolderClick, menuItems, isEditing, setIsEdi
           </FormikProvider>
         }
       </TableCell>
+      {desktop &&
+        <TableCell align="left">
+          {isOwner && t`me`}
```
Tbaut: We should show here who's the owner when it's not us. This info can be found on the `bucket.owners` afaik

Tbaut (changes requested):
We should show the actual owner when it's not us.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/SharedFolderRow.tsx
tanmoyAtb: sounds good!

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/SharedFolderRow.tsx
```diff
@@ -216,6 +217,13 @@ const SharedFolderRow = ({ bucket, onFolderClick, menuItems, isEditing, setIsEdi
           </FormikProvider>
         }
       </TableCell>
+      {desktop &&
+        <TableCell align="left">
+          {isOwner
+            ? t`me`
```
RyRy79261, line 223: Shouldn't this be in a bubble as well? Seems a bit awkward just the left aligned word there

Tbaut: Problem with a bubble is that you need an additional action (mouse over) to know if you're the owner. The original issue is about knowing right away https://github.com/ChainSafe/files-ui/issues/1129 and it contains the design which Cindy spent quite some time on. If we have issues with the original design, I think it'd be better to discuss it there. I don't think the "me" is awesome, but I personally prefer this than a bubble that needs an action

Tbaut (approved):

FSM1 (approved):

PR #1231, opened 2021-07-05
Fix bulk operations for csf buckets

Tbaut:
This is fix for a bug I introduced in https://github.com/ChainSafe/files-ui/pull/1219/ 
How to reproduce, select at least 1 File in a csf bucket (normal, non shared bucket) and see that the bulk operation buttons don't show up.

edit: Github shows a complex diff but the reality is just a `if (!!permission && isShared) {` that was closed too late, it should be closed right after the `switch`.

### DIFF
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
@@ -487,34 +487,23 @@ const FilesList = ({ isShared = false }: Props) => {
         fileOperations = baseOperations
         break
       }
+    }
 
-      for (let i = 0; i < selectedCids.length; i++) {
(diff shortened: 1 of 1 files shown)


packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -597,6 +597,7 @@ const FilesList = ({ isShared = false }: Props) => {
     setIsDeleteModalOpen(true)
   }, [])
 
+  console.log("validBulkOps.indexOf(move)", validBulkOps.indexOf("move"))
```
Tbaut:
```suggestion
```

RyRy79261 (approved):

tanmoyAtb (approved):

PR #1232, opened 2021-07-05
Fix shared folder creation path

Tbaut:
This PR prevents sending the `/shared/{bucket-id}` to the `api` and tries to create paths similarly to what is done for csf bucket.


FSM1 (approved):
this is great :)

RyRy79261 (approved):

PR #1235, opened 2021-07-05
Fix refresh token flow

### LINKED ISSUE #1224: Access Token Refresh flow not working
**Describe the bug**
Once the access token expires, there is an error thrown while attempting to refresh the expired token and further operation is not possible until the page is refreshed

**To Reproduce**
1. Navigate to the app
2. Log in
3. Wait till the access token expires (10 minutes)
4. Attempt any action that requires authentication
5. Observe an error thrown when the request is made, and 401 response is received

**Expected behavior**
The refresh token should be utilized to obtain a fresh access token and the original request that failed should be retried

**Console error**
image

**Additional notes**
Initial investigation seems to indicate that the refresh token is not being read from session storage correctly, causing the cascading failure.

FSM1:
Closes #1224 

Replace call to the session storage helper with direct browser Api calls to `sessionStorage.getItem()` inside the refresh token handler.

### DIFF
--- a/packages/files-ui/src/Contexts/FilesApiContext.tsx
+++ b/packages/files-ui/src/Contexts/FilesApiContext.tsx
@@ -43,7 +43,7 @@ const FilesApiProvider = ({ apiUrl, withLocalStorage = true, children }: FilesAp
 
   const { wallet, onboard, checkIsReady, isReady } = useWeb3()
   const { localStorageRemove, localStorageGet, localStorageSet } = useLocalStorage()
-  const { sessionStorageRemove, sessionStorageGet, sessionStorageSet } = useSessionStorage()
+  const { sessionStorageRemove, sessionStorageSet } = useSessionStorage()
 
   // initializing api
   const initialAxiosInstance = useMemo(() => axios.create({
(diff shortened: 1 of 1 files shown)


Tbaut (approved):
Looks good, I had a look at the code, and I'm not sure I understand the `withLocalStorage` altogether. If you recall what this is @FSM1 and let me know, I'd be happy to add some comments about it.

RyRy79261 (approved):

FSM1:
> Looks good, I had a look at the code, and I'm not sure I understand the `withLocalStorage` altogether. If you recall what this is @FSM1 and let me know, I'd be happy to add some comments about it.

When the FilesApiContext was originally built, it was meant to be more generic component, and a flag could be passed in that would determine where the refresh token was stored (local or session). 

The real issue this ticket solves is that the initialize useEffect (which we really only want to run once) did not have the `sessionStorageGet` and `localStorageGet` functions as part of the dependency array, and as such these were always resolving to undefined (in the context of the refresh token handling flow), and no token refresh was happening. Using the browser API directly solves this issue.

Since Files is only using session storage for the refresh token, I just removed the localStorage branch of the ternary.

PR #1236, opened 2021-07-05
Storage-ui: Add identifiers for page objects and basic bucket test 

### LINKED ISSUE #1187: Testing the pinning of an IPFS file

### LINKED ISSUE #1188: Testing the creation of a bucket

asnaith:
*closes #1188* and closes #1187

*What I changed* 
- Added lots of new identifiers for storage-ui page objects (updated some files ones too where it made sense). 
- Updated some pre-existing identifiers to better match consistency used elsewhere
- Added basic tests for bucket creation
- Added basic tests for CID pinning
- Added ability to clear pins prior to test launch (needs confirming once product issues are resolved)


asnaith:
This is a draft only at the moment as I'm having issues using the newly created `cy.web3Login({ clearPins: true })` in the test. The pinning clearing is not working and throwing an unhandled promise error

Tbaut:
Friendly reminder that many small PRs > one big. So that we can take the time to look at every detail and avoid conflicts as much as possible. Should we take a look at it now?

asnaith:
> Friendly reminder that many small PRs > one big. So that we can take the time to look at every detail and avoid conflicts as much as possible. Should we take a look at it now?

@Tbaut I hear you, this one is dragging along. It made sense to do it all at the same time but also taking longer to complete than expected because I keep running into legitimate issues in the product (trying to write tests before all the app issues are resolved) and then cypress (nuking local storage on cy.reload). 

I've just brought in the latest dev changes and dealt with the merge conflicts if we are happy to merge as is (I've left some comments explaining a few things that are not as complete as I'd like). 

I've also marked the newly added tests as `it.skip` to make sure they don't get triggered in CI. I can't reliably "tests the tests" at the moment until product issues are resolved.

packages/storage-ui/cypress/support/commands.ts
```diff
@@ -39,13 +40,16 @@ export interface Web3LoginOptions {
...
+  clearPins?: boolean
 }
 
 const SESSION_FILE = "cypress/fixtures/storage/sessionStorage.json"
 const LOCAL_FILE = "cypress/fixtures/storage/localStorage.json"
 
+Cypress.Commands.add("clearPins", (apiUrlBase: string) => {
+  apiTestHelper.clearPins(apiUrlBase)
+})
+
```
asnaith, line 52: The new `clearPins` command is still not working but I think it's because we have other unresolved promise issues in the actual app so I was planning on looking into it more once they are resolved.

packages/storage-ui/cypress/support/commands.ts
```diff
@@ -126,13 +131,15 @@ Cypress.Commands.add(
...
+          "there is something in local storage ---> direct login"
       )
 
       if (local.length === 0) {
-        cy.log("nothing in session storage, --> click on web3 button")
+        cy.log("nothing in local storage, --> click on web3 button")
         authenticationPage.web3Button().click()
         authenticationPage.metaMaskButton().click()
+        // eslint-disable-next-line cypress/no-unnecessary-waiting
+        cy.wait(1000)
```
asnaith, line 142: I haven't found a reliable way to get past "detached from DOM" issues, when clicking the `web3SignInButton()`,  without this hardcoded wait being in place.

packages/storage-ui/cypress/support/commands.ts
```diff
@@ -141,6 +148,23 @@ Cypress.Commands.add(
...
+      cy.reload({ timeout: 50000 }).then(() => {
+        if (local.length === 0) {
+          // Temp work around for local storage being cleared after the reload          
+          cy.log("nothing in local storage after reload, --> click on web3 button")
+          authenticationPage.web3Button().click()
+          authenticationPage.metaMaskButton().click()
+          // eslint-disable-next-line cypress/no-unnecessary-waiting
+          cy.wait(1000)
+          authenticationPage.web3SignInButton().click()
+        }
```
asnaith, line 163: This is a temp workaround to commands.ts that checks the storage after the reload and clicks through if it is blank (which it always seems to be if it's the first reload of the first test). 

It's not great but could be improved upon in another PR if we want to get this merged sooner.

RyRy79261 (approved):

packages/storage-ui/src/Components/Layouts/AppHeader.tsx
```diff
@@ -180,14 +180,17 @@ const AppHeader = ({ navOpen, setNavOpen }: IAppHeader) => {
...
                       onClick: () => signOut(),
                       contents: (
                         <div className={classes.menuItem}>
                           <PowerDownSvg />
                           <Typography>
-                            <Trans>Sign Out</Trans>
+                            <span data-cy="menu-sign-out">
+                              <Trans>Sign Out</Trans>
+                            </span>
                           </Typography>
```
Tbaut: Why not adding the data-cy to the `Typography` element, looks like you did it that way in this PR for things. If that doesn't work there's still the `div` above rather than adding a `span`.

packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
```diff
@@ -127,7 +127,9 @@ const AddCIDModal = ({ modalOpen = false, close }: IAddCIDModuleProps) => {
         onSubmit={onSubmit}
       >
         <Form>
-          <div className={classes.root}>
+          <div className={classes.root}
+            data-cy="form-pin-cid"
+          >
```
Tbaut: nit
```suggestion
          <div
            className={classes.root}
            data-cy="form-pin-cid"
          >
```

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -171,7 +174,9 @@ const BucketsPage = () => {
         }}
         closePosition="none"
       >
-        <div className={classes.modalRoot}>
+        <div className={classes.modalRoot}
+          data-cy="form-create-bucket"
+        >
```
Tbaut: nit
```suggestion
        <div
          className={classes.modalRoot}
          data-cy="form-create-bucket"
        >
```

packages/storage-ui/cypress/tests/cid-management-spec.ts
```diff
@@ -0,0 +1,32 @@
...
+describe("CID management", () => {
+
+  context("desktop", () => {
+
+    it.skip("can pin a CID", () => {
+      cy.web3Login({ clearPins: true })
+
+      // pin a cid and see it in the pinned items table
+      cidsPage.pinButton().click()
+      cidsPage.cidInput().type("QmZEE7Ymh2mRMURLnFLipJTovb44AoUYDzx7aipYZwvxX5")
```
Tbaut: Just a comment for the future, I see it's skipped, but ideally this cid should be in the fixture rather than hardcoded

packages/storage-ui/cypress/tests/bucket-management-spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      bucketsPage.createBucketCancelButton().click()
+      bucketsPage.createBucketForm().should("not.exist")
+    })
+
+    it.skip("can delete a bucket", () => {
+      cy.web3Login({ clearPins: true })
+
+      // delete a bucket and ensure it's row is removed
+      navigationMenu.bucketsNavButton().click()
+      bucketsPage.createBucket("Delete Me")
```
Tbaut: Same here, this should be a fixture or a shared variable so that we can check it later on.

packages/storage-ui/cypress/support/commands.ts
Tbaut: We had the same problem in Files, and I solved it with black magic.. and something I still don't understand today. @FSM1  had integrated the fix here https://github.com/ChainSafe/ui-monorepo/pull/1096/files#diff-02266cd4a001cf379134518735d94aaa81a29482fb341d2a12424eeebcb89140L205

We are importing this part of the UI from a separate component. Doing this broke the tests. Getting it back in and it solves it. We can check it out in details later, I don't want to stop this PR for this only.

packages/files-ui/src/Components/Layouts/AppHeader.tsx
```diff
@@ -205,14 +205,17 @@ const AppHeader = ({ navOpen, setNavOpen }: IAppHeader) => {
...
+                  testId="sign-out"
                   menuItems={[
                     {
                       onClick: () => signOut(),
                       contents: (
                         <div className={classes.menuItem}>
                           <PowerDownSvg />
                           <Typography>
-                            <Trans>Sign Out</Trans>
+                            <span data-cy="menu-sign-out">
```
Tbaut: Same comment here, why not the div or the Typography element?

Tbaut (approved):
Thanks! I found a couple nits. Approving so as not to block you once you've addressed the comments and conflicts.

packages/storage-ui/cypress/support/commands.ts
asnaith: 👍 @Tbaut Created a separate issue to improve this later https://github.com/ChainSafe/ui-monorepo/issues/1282

packages/storage-ui/cypress/support/commands.ts
asnaith: Will address later here: https://github.com/ChainSafe/ui-monorepo/issues/1283

packages/storage-ui/src/Components/Layouts/AppHeader.tsx
asnaith: @Tbaut Good point, they shouldn't be on the typography element at all. I've just pushed an update to change all the identifiers to use the relevant div or link. That makes more sense, closer to the element that a real user would click.

packages/storage-ui/cypress/tests/cid-management-spec.ts
asnaith: @Tbaut Agreed, I've updated this.

PR #1241, opened 2021-07-06
Delete files & folders from shared folders directly.

### LINKED ISSUE #1203: Shared files and folder should be deleted directly and not go in the bin
After https://github.com/ChainSafe/files-ui/pull/1048 the deleted shared files end-up in the bin.
We should have a dedicated error message and have them deleted.

tanmoyAtb:
closes #1203


Tbaut (approved):
:100:

FSM1 (approved):

Issue #1207, opened 2021-06-30
Update OAuth and add email login for Storage


FSM1:
- ~~Integrate new API client after the API updates are made~~
- Migrate OAuth flows to use the Torus DirectAuth SDK instead of getting the OAuth redirect from the API.
- Allow login with MM, email, Github/Google

See the detailed login flows here: https://drive.google.com/file/d/1h09MiNOdiDopqkxnrDnWVAF1nukpU8cv/view?usp=sharing

PR #1242, opened 2021-07-06
Fix web3 login flow for storage

FSM1:
This PR addresses some of the changes described in #1207 and necessitated by [API PR #1257](https://github.com/ChainSafe/files-api/pull/1257), namely
- Integrate new API client after the API updates are made
- Update Storage Auth flow for web3 login to use the new API.


packages/storage-ui/src/Contexts/StorageApiContext.tsx
```diff
@@ -226,99 +215,35 @@ const StorageApiProvider = ({ apiUrl, withLocalStorage = true, children }: Stora
...
-      return Promise.resolve()
-    } catch {
-      return Promise.reject("There was an error logging in")
+    if (!isReady) {
+      const connected = await checkIsReady()
+      if (!connected) return Promise.reject("You need to allow the connection")
     }
-  }
+    const signer = provider.getSigner()
+    const address = await signer.getAddress()
```
Tbaut: Can we keep this inside the try/catch?

packages/storage-ui/src/Contexts/StorageApiContext.tsx
```diff
@@ -226,99 +215,35 @@ const StorageApiProvider = ({ apiUrl, withLocalStorage = true, children }: Stora
...
-          signature: signature,
-          token: token,
-          public_address: address
-        })
-        setTokensAndSave(access_token, refresh_token)
-        setReturningUser()
-        return Promise.resolve()
-      }
     } catch (error) {
       return Promise.reject("There was an error logging in.")
```
Tbaut, line 249: In case it helps future us debug.
```suggestion
      console.error(error)
      return Promise.reject("There was an error logging in.")
```

Tbaut (approved):
We expect to implement the Google/Github login shortly so that we don't want to hide the buttons right?

FSM1:
> We expect to implement the Google/Github login shortly so that we don't want to hide the buttons right?

Yeah Ryan is on the oauth/email login

packages/storage-ui/src/Contexts/StorageApiContext.tsx
FSM1: done :)

Tbaut:
@asnaith this is the one that will unblock you for web3 login. Feel free to approve and merge

RyRy79261 (approved):

PR #1245, opened 2021-07-07
Clean locally stored cache files before running tests

### LINKED ISSUE #1149: Delete storage fixture before all the tests are ran
When launching the test suite, if we have ever ran it in the past, we may have outdated `packages/files-ui/cypress/fixtures/storage` stored locally. The `web3Login` will find these, and expect the browser to login the user directly. Now because the data contained in those files is outdated (refresh token expired) the browser will not log the user in and the test will fail.

Although it's not a problem on CI because the tests are started afresh every single time, it's annoying when testing locally.

Ideas how to solve this:
- There may be a "beforeAll" hook in Cypress to run before all the test where we'd use `fs` in a new task (in `cypress/plugins`) to remove those files
- Add a `rimraf packages/files-ui/cypress/fixtures/storage` in the test script.

Tbaut:
closes #1149


asnaith:
Awesome! Thanks for this very useful improvement @Tbaut 🙏

asnaith (approved):

FSM1 (approved):

PR #1247, opened 2021-07-07
Analytics Integration

priom:


Tbaut (changes requested):
As far as I can see, this would use the hosted version of Posthog, effectively sharing personal data from our users* with a 3rd party service. 
*those that do not use an ad/analytics blocker.

While it is stated in Files privacy policies that ChainSafe could share personal data with 3rd parties, I personally do not encourage the use of analytics, even less if it's not self hosted, because this would be data harvesting without clear goals of what to do with these.

From our privacy policies:
> In keeping privacy top of mind, our goal is to make sure that any information we require from you is necessary for the limited purposes you consent to.

Actually keeping privacy top of my mind, I do not see how these analytics should be required, or how they would benefit our users. I am of the opinion that server analytics, carefully crafted user surveys and user interviews already allow to answer any question one could have to help shape the product strategy.

Tbaut:
On the legal part, for European users thanks to GDPR, the Files application would [have to show a consent banner](https://www.cookieyes.com/gdpr-compliant-cookie-consent-banner/) with an opt-in button before enabling cookies related to 3rd party analytics.

sweetpea22:
> Actually keeping privacy top of my mind, I do not see how these analytics should be required, or how they would benefit our users. 

I am also concerned about the social costs that collecting as much data as possible to profile our users. I feel that this approach is at odds with our value proposition and users may also feel the same way. If we give people reason to distrust our product, this can have cascading consequences on product success and badly reflect chainsafe as a whole even if we didn't do anything illegal or in bad faith. 

I do believe that there is a need to understand our user needs deeply, but whether it's worth undermining customer trust and perceived safety is a big question.

One solution we could use is to decide which analytics are most critical to fit our market and be completely transparent with what we're using it for, allow users to opt-in, and make it known we self-host the info. The only thing is, the people who opt in to analytics may represent a biased fraction of the real user pool, and sometimes biased is worse than blind.

In the near term let's think about separating out what we can find out by asking our users, and what we can't find out.

GregTheGreek:
hey! 

Just came across this, I think theres good points on both sides here and we should definitely discuss how we can create better analytics without exposing ourselves to third party tools.

Maybe we can create a [github discussion](https://github.com/ChainSafe/files-ui/discussions) to discuss the different options.

cc: @sweetpea22 @Tbaut @priom

priom:
Thanks, everyone for bringing up these valid points and got confirmation from legal that we need to make some changes so, I have closed this PR for now till we complete the prerequisites.

cc: @Tbaut @sweetpea22

PR #1255, opened 2021-07-09
Copy file to shared folder

### LINKED ISSUE #1161: Sharing an existing document
This should be a new menu item where users should
- chose a shared folder, or create a new one
- chose who they want to share it with
-  re-encrypt and re-upload it to the right bucket

It may be interesting to start downloading the file in the background on the first step, so that the only steps needed once the user clicks on "share" is to re-encrypt and upload 

Design https://www.figma.com/file/wRMGqc11iQLVpDVmqXVXWK/ChainSafe-Files---Dev?node-id=2349%3A59153

image
image
image
image
image

Tbaut:
closes #1161


packages/files-ui/src/Components/Modules/FileBrowsers/ShareFileModal.tsx
```diff
@@ -0,0 +1,506 @@
...
+    () => isCreatingSharedFolder || isDownloading || isUploading
+    , [isCreatingSharedFolder, isDownloading, isUploading]
+  )
+
+  const bucketsOptions = useMemo(() => {
+    if (!profile) {
+      return []
+    }
+
+    // todo extract in helper for isOwner, isWritr, isReader
```
Tbaut: FYI I'll do that in a separate PR now for an easier review process

FSM1 (changes requested):
This is amazing work Thibaut.

The button layout on the Share modal needs to be fixed: 
image

Additionally, The `ApiClient.ShareBucket(...)` API call should be made once the upload to the shared folder has been made. This will remove the original file from the user's file system and create the "symlink" to the shared bucket in the users file system. This change will also require some modifications to the FileBrowser to handle the new content type, with an option to "view in shared folder"

Tbaut:
> The button layout on the Share modal needs to be fixed

:see_no_evil: I've seen this for so long now that I didn't even think about it any more!
edit: done.

> The ApiClient.ShareBucket(...) API call should be made once the upload to the shared folder has been made. This will remove the original file from the user's file system and create the "symlink"

Ah, I didn't think about that one, this would mean that someone with write access would actually be able to delete that file, I'm not sure what would be the implications. I'll ask in the chat to discuss about this.

RyRy79261 (commented):
An awesome clean up and implementation, just testing render quick

RyRy79261:
Just some minor styling points: 
Some spacing between the bottom two elements:
image

Not sure about the view folder link here, maybe some spacing: 
image

FSM1 (approved):
Really amazing work @Tbaut 🥇 Since the "Move to shared folder" can be added later I think we can merge this, and figure out which way we are going to tackle the rest in another issue

Tbaut:
Here we go @RyRy79261 , should be better now

image
image

RyRy79261 (approved):

PR #1257, opened 2021-07-12
Sorted by dates by default

### LINKED ISSUE #1215: Storage UI - Pins should be sorted by Created Date
The default sorting for the CIDs Overview should be by Created Date descending (Newest items first)

tanmoyAtb:
closes #1215 

I kept the states in place, in case we want to have sorting functionality on the table columns in future


packages/storage-ui/src/Components/Pages/CidsPage.tsx
```diff
@@ -53,6 +54,28 @@ const CidsPage = () => {
...
+    }
+    case "size": {
+      temp = pins.sort((a, b) => (a.info?.size < b.info?.size ? -1 : 1))
+      break
+    }
+    case "date_uploaded": {
+      temp = pins.sort((a, b) => (a.created < b.created ? -1 : 1))
+      break
+    }
+    }
```
Tbaut: This is duplicated, we should have default at the end, and remove the "date_uploaded" part.
```suggestion
    case "size": {
      temp = pins.sort((a, b) => (a.info?.size < b.info?.size ? -1 : 1))
      break
    }
    default: {
      temp = pins.sort((a, b) => (a.created < b.created ? -1 : 1))
      break
    }
    }
```

Tbaut (commented):
I just checked the UI, there's no sorting in the columns? Could we add it? Otherwise all the logic for the sorting direction etc isn't used.

tanmoyAtb:
> I just checked the UI, there's no sorting in the columns? Could we add it? Otherwise all the logic for the sorting direction etc isn't used.

I did not put the sorting buttons into the column heads, as this was not in the issue scope,
The issue only asked for default sorting on dates,
But if thats something we want now, I could quickly add that in!

Tbaut:
My main concern is that this PR is a bit in between. Either we just build what the issue wants, and sort things per name in a hard-coded manner, or we copy/paste some older code, but then we make sure it's actually fully used. Right now all these `useState` have no actual use-case because there's no setter.

I guess it's a quick win to add the sorting in the header, but I personally wouldn't mind if you go the other route.

tanmoyAtb:
> My main concern is that this PR is a bit in between. Either we just build what the issue wants, and sort things per name in a hard-coded manner, or we copy/paste some older code, but then we make sure it's actually fully used. Right now all these `useState` have no actual use-case because there's no setter.
> 
> I guess it's a quick win to add the sorting in the header, but I personally wouldn't mind if you go the other route.

All right, I totally feel you, 
And the unused states were bugging me as well, so I added sorting on dates and size !

FSM1 (approved):
looking good 🎉

Tbaut (approved):

PR #1258, opened 2021-07-12
Create bucket modal changes

### LINKED ISSUE #1237: ChainSafe Storage: Input field retains the bucket name of previously created bucket
**Describe the bug**
The input field on the "Create Bucket" modal retains the text of the previously created bucket. 

**To Reproduce**
1. Login to ChainSafe storage
3. Click on "Create Bucket" and make a new bucket called "ABC"
4. Click on "Create Bucket" again to create an additional bucket
5. Observe the input field is not blank 

**Observed behaviour**
The bucket name input field on the modal retains the text of the previously created bucket. 

**Expected behavior**
The input field on the "Create Bucket" modal should be blank when the modal opens

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

tanmoyAtb:
Closes #1237 

Added formik, focus and loaders as well


packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -101,27 +103,48 @@ const useStyles = makeStyles(({ breakpoints, animation, constants, typography }:
...
 const BucketsPage = () => {
   const classes = useStyles()
   const { storageBuckets, createBucket } = useStorage()
-  const [createBucketModalOpen, setCreateBucketOpen] = useState(false)
-  const [newBucketName, setNewBucketName] = useState("")
+  const [createBucketModalOpen, setCreateBucketModalOpen] = useState(false)
+  const inputRef = useRef<any>(null)
+
+  useEffect(() => {
+    if (createBucketModalOpen) {
```
Tbaut: This is a dirty hack. we should use the `autoFocus` props on the Formik field

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -101,27 +103,48 @@ const useStyles = makeStyles(({ breakpoints, animation, constants, typography }:
...
+      setTimeout(() => inputRef.current?.focus(), 100)
+    }
+  }, [createBucketModalOpen])
+
+  const folderNameValidator = yup.object().shape({
```
Tbaut: I think this can be defined outside of the component because it has no dependency. This way it doesn't get computed for each render.

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -101,27 +103,48 @@ const useStyles = makeStyles(({ breakpoints, animation, constants, typography }:
...
     }
   })
 )
 
 const BucketsPage = () => {
   const classes = useStyles()
   const { storageBuckets, createBucket } = useStorage()
-  const [createBucketModalOpen, setCreateBucketOpen] = useState(false)
-  const [newBucketName, setNewBucketName] = useState("")
+  const [createBucketModalOpen, setCreateBucketModalOpen] = useState(false)
```
Tbaut: Could we have a `isCreateBucketModalOpen` ? for consistency, any boolean state should be is*

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -172,52 +195,72 @@ const BucketsPage = () => {
...
-            item
-            xs={12}
-            sm={12}
+          <Formik
+            initialValues={{
+              name: ""
+            }}
+            validationSchema={folderNameValidator}
+            validateOnChange={false}
+            onSubmit={async (values) => {
```
Tbaut: Let's use Promise.then.catch. so that we log thing and don't need this try/catch + async

packages/storage-ui/src/Contexts/StorageContext.tsx
```diff
@@ -150,15 +150,18 @@ const StorageProvider = ({ children }: StorageContextProps) => {
     return storageApiClient.addPin(({ cid }))
   }, [storageApiClient])
 
-  const createBucket = useCallback((name: string) => {
-    storageApiClient.createBucket({
```
Tbaut, line 154: Instead of changing this all to try/catch, we can just return the promise I think, `return storageApiClient.createBucket(...`

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
tanmoyAtb: great suggestion! thanks!

packages/storage-ui/src/Contexts/StorageContext.tsx
tanmoyAtb: I wanted to do that, but I also had to call `refreshBuckets` after the `createBucket`, and I didn't want refreshBucket in the BucketsPage

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
tanmoyAtb: using `useFormik` now with `then catch`

packages/storage-ui/src/Contexts/StorageContext.tsx
Tbaut: I'm not sure I understand what you mean. From what I understand you want to make sure the `createBucket` returns a `Promise<void>` so that you can chain some other logic (in the `BucketsPage`).

What I would suggest is to keep the whole .then.catch, and just return the whole promise, so that you can chain things in the same way. This should do the trick

```diff
  const createBucket = useCallback((name: string) => {
-    storageApiClient.createBucket({
+    return storageApiClient.createBucket({
      name,
      type: "fps",
      public: "read",
      encryption_key:""
    }).then(() =>
      refreshBuckets()
    ).catch(console.error)
  }, [storageApiClient, refreshBuckets])

```

tanmoyAtb: Sorry, I was thinking of importing the refreshBuckets as well,
Anyways I updated it to your suggestions,
good to go!

asnaith:
Hey @tanmoyAtb, I checked this out on Render and I was still able to reproduce the problem by following the repro steps mentioned on #1237

tanmoyAtb:
> Hey @tanmoyAtb, I checked this out on Render and I was still able to reproduce the problem by following the repro steps mentioned on #1237

Thanks for checking in on this, I pushed a change which should fix this!

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -101,27 +103,61 @@ const useStyles = makeStyles(({ breakpoints, animation, constants, typography }:
...
+      fontSize: 14,
+      lineHeight: "22px"
     }
   })
 )
 
+const bucketNameValidator = yup.object().shape({
+  name: yup
+    .string()
+    .required(t`Folder name is required`)
```
FSM1:
```suggestion
    .required(t`Bucket name is required`)
```

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -101,27 +103,61 @@ const useStyles = makeStyles(({ breakpoints, animation, constants, typography }:
...
+    .test(
+      "Invalid name",
+      t`Folder name cannot contain '/' character`,
```
FSM1:
```suggestion
      t`Bucket name cannot contain '/' character`,
```

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -101,27 +103,61 @@ const useStyles = makeStyles(({ breakpoints, animation, constants, typography }:
...
+      (val: string | null | undefined) => !!val && !val.includes("/")
```
FSM1: Not sure if this rule is valid for bucket names

FSM1 (changes requested):
Just a few minor comments, but otherwise, works really well.

packages/storage-ui/src/Contexts/StorageContext.tsx
```diff
@@ -150,15 +150,13 @@ const StorageProvider = ({ children }: StorageContextProps) => {
...
+  const createBucket = useCallback(async (name: string) => {
+    return storageApiClient.createBucket({
       name,
       type: "fps",
       public: "read",
       encryption_key:""
-    }).then(() =>
-      refreshBuckets()
-    ).catch(console.error)
+    }).then(refreshBuckets).catch(console.error)
```
Tbaut:
```suggestion
    })
    .then(refreshBuckets)
    .catch(console.error)
```

Tbaut (approved):
Good to go after the suggestions

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
tanmoyAtb: Nevertheless, its a good rule to have I feel, not allowing '/'

Tbaut: I'd say so too, the / would only bring mess. I would reconsider only if that's something annoying some users.

FSM1 (approved):

PR #1269, opened 2021-07-14
Edit shared folder by owner

### LINKED ISSUE #1141: Add/remove users with read/write permission to a shared folder
An owner should be able to add/remove users the shared folder is shared with.
An owner should be able to change the permission level of a current user (read/write).

It looks like it's possible with `updateBucket` from `FilesApiClient`

The UI should be a modal, with 2 text fields, populated with tags (each tag is a user that is added to the share). The user lookup should happen when typing into either of the text fields.

It is not possible to have a user withing the "writers" and the "readers". We should make sure that a user doesn't appear with both right.

tanmoyAtb:
closes #1141


Tbaut (commented):
Quick check as I played with it, I was able to add the same user to both fields
image

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFolderRowWrapper.tsx
```diff
@@ -29,9 +30,10 @@ interface Props {
   bucket: BucketKeyPermission
   handleRename: (bucket: BucketKeyPermission, newName: string) => void
   openSharedFolder: (bucketId: string) => void
+  onUpdateSharedFolder: () => void
```
Tbaut: I think the word "update" is misleading. The icon and the wording make me think of upgrade, or security update. I think this should rather be "update permissions", or even "edit permission"?

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFoldersOverview.tsx
```diff
@@ -228,7 +231,13 @@ const SharedFolderOverview = () => {
       </article>
       <CreateSharedFolderModal
         modalOpen={createSharedFolderModalOpen}
-        close={() => setCreateSharedFolderModalOpen(false)}/>
+        close={() => setCreateSharedFolderModalOpen(false)}
+      />
+      <UpdateSharedFolderModal
+        isModalOpen={!!bucketToUpdate}
+        onClose={() => setBucketToUpdate(undefined)}
```
Tbaut: Mixing concerns doesn't scale well, this is a big part of the refactoring I did in the past days. `bucketToUpdate` should not be used to know whether or not this modal should be open. We should have (like the other modal above) a variable `isUpdatePermissionModalOpen` which sole purpose is to trigger the modal. The bucketToUpdate is an additional prop that has 1 job then.

tanmoyAtb:
> Quick check as I played with it, I was able to add the same user to both fields
> image

What do you think should be the course of action here? should I remove the user from the previous box and add to this one ?

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFoldersOverview.tsx
Tbaut: Also if you don't want to bother, you can have
```
{isUpdatePermissionModalOpen && <UpdateSharedFolderModal
        onClose={() => setBucketToUpdate(undefined)}}
        ...
```

so that you don't need to care inside the modal about whether or not it's open.

packages/files-ui/src/Components/Modules/FileBrowsers/UpdateSharedFolderModal.tsx
```diff
@@ -0,0 +1,296 @@
...
+  const classes = useStyles()
+  const { updateSharedFolder } = useFiles()
+  const { filesApiClient } = useFilesApi()
+  const { profile } = useUser()
+  const [isUpdatingSharedFolder, setIsUpdatingSharedFolder] = useState(false)
+  const [sharedFolderWriters, setSharedFolderWriters] = useState<{label: string; value: string; data: SharedUser}[]>([])
+  const [sharedFolderReaders, setSharedFolderReaders] = useState<{label: string; value: string; data: SharedUser}[]>([])
+
+  useEffect(() => {
+    setSharedFolderWriters(
```
Tbaut: I'd add an early return for `if (!bucket) return`

packages/files-ui/src/Components/Modules/FileBrowsers/UpdateSharedFolderModal.tsx
```diff
@@ -0,0 +1,296 @@
...
+      })
+      ) || []
+    )
+  }, [bucket])
+
+  const { desktop } = useThemeSwitcher()
+
+  console.log(bucket)
+
+  const handleLookupUser = useCallback(async (inputVal: string, permission: "reader" | "writer") => {
```
Tbaut: This is duplicated right? Can we please extract it to a custom context?

packages/files-ui/src/Components/Modules/FileBrowsers/UpdateSharedFolderModal.tsx
```diff
@@ -0,0 +1,296 @@
...
+          >
+            <Trans>Cancel</Trans>
+          </CustomButton>
+          <Button
+            size={desktop ? "medium" : "large"}
+            variant="primary"
+            className={classes.okButton}
+            loading={isUpdatingSharedFolder}
+            onClick={handleUpdateSharedFolder}
+            disabled={!isValid}
```
Tbaut: Any way we can disable it if nothing was changed?

Tbaut (dismissed):
> What do you think should be the course of action here? should I remove the user from the previous box and add to this one ?

We should let the user do it. Let's show an error similar to the one in the moveModal, and also disable the update button.

I went super quickly though the code , will go more in depth later but i left a couple comments.

Tbaut (commented):
I made some code suggestions in a separate PR so that it's easy to review, feel free to merge if you're ok with them. https://github.com/ChainSafe/ui-monorepo/pull/1292

Another thing that I found is that we should be allowed to have empty readers and writers arrays. Right now the `isValid` would be false. Say I'm sharing with someone, just 1 person, and I want to remove them, right now I can't

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFolderRowWrapper.tsx
FSM1: Agree that `update` isnt the right word here. Let's call this `Edit`

packages/files-ui/src/Components/Modules/FileBrowsers/UpdateSharedFolderModal.tsx
tanmoyAtb: I'll extract this in the merging PR to reuse same function!

RyRy79261 (approved):

RyRy79261 (approved):

FSM1 (approved):
Working great :)

PR #1277, opened 2021-07-15
Fix to prevent from moving folders in sub-folders

### LINKED ISSUE #1267: ChainSafe Storage: "Ghost" bucket appearing for all users on first row
**Description**
A blank bucket row is being displayed for all users on Row 1 

**Steps to Reproduce**
1. Log into ChainSafe Storage
2. Navigate to Buckets screen

**Observed behavior**
The first row (appearing for all users) is blank. 

**Expected behavior**
The only buckets being displayed should be ones created by the user

**Screenshots**
Screen Shot 2021-07-14 at 8 17 12 AM

 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

### LINKED ISSUE #1276: ChainSafe Storage: Folder can be moved into it's subfolder
**Description**
It's possible to lose complete access to a folder (at least in the UI) by moving a folder to a subfolder within itself 

**Steps to Reproduce**
1. Login to Storage
2. Create bucket
3. Navigate to newly created bucket 
4. Create a "Folder1" in the bucket
5. Navigate inside "Folder1" and create another folder inside it called "Folder2" 
6. Return to the main bucket view and select "Folder1"
7. Click the "Move selected" button 
8. In the move modal, expand Folder1 so that Folder2 can be selected
9. Select "Folder2" as the destination and click "Move"

**Observed behavior**
It's possible to move a folder into it's subfolder and then lose complete access to both folders in the UI

**Expected behavior**
It should not be possible to move the folder into a subfolder within itself 

**Screenshots**
Screen Shot 2021-07-14 at 3 18 28 PM

**Video**
https://user-images.githubusercontent.com/3982665/125700569-672a2758-66c7-4dfd-8e81-76cc61fa979b.mov

**Desktop**
 - OS: macOS
 - Browser: Firefox
 - Version : 89.02

Tbaut:
closes #1276 

My helper function involves some JS tricks that I commented, let me know if it's not clear.


packages/files-ui/src/Utils/pathUtils.ts
```diff
@@ -57,4 +57,33 @@ export function extractSharedFileBrowserPathFromURL(browserUrl: string, modulePa
...
+  }
+
+  // we loop over every section of path1
+  path1.every((section, index) => {
+
+    // so far every paths were identical
+    // if paths1 is longer than paths2
+    if (index === path2.length){
+      result = true
+      //break out of the every look
```
Tbaut:
```suggestion
      //break out of the loop
```

packages/storage-ui/src/Utils/pathUtils.ts
```diff
@@ -54,4 +54,33 @@ export function getParentPathFromFilePath(filePath: string) {
...
+  }
+
+  // we loop over every section of path1
+  path1.every((section, index) => {
+
+    // so far every paths were identical
+    // if paths1 is longer than paths2
+    if (index === path2.length){
+      result = true
+      //break out of the every look
```
Tbaut:
```suggestion
      //break out of the loop
```

FSM1 (changes requested):
It seems like there is something going wrong with the logic here:
image

See the screenshot above where I am trying to move `/another parent` into `/parent` but not being allowed to even though there is nothing that should be preventing the move. i.e. after the move the path should be 

```
|-parent
|--Another parent
|--child of parent
```

Tbaut:
Ah right, nice catch, I don't allow to move the folder into *any* subfolder of the current path. I'll adapt.

Tbaut:
Ok this is turning into some dark magic.. I will probably not understand half of it any more in 2 weeks, hence the comments.
I tested the weirdest things and it look all right to me now.

packages/files-ui/src/Utils/pathUtils.ts
```diff
@@ -57,4 +57,32 @@ export function extractSharedFileBrowserPathFromURL(browserUrl: string, modulePa
...
+  // the path1 can only be a subfolder
+  // if it's longer than path2
+  if (path1.length < path2.length){
+    return result
+  }
+
+  // we loop over every section of path1
+  path1.every((section, index) => {
+    // so far every paths were identical
+    // if paths1 is longer than paths2
```
Tbaut:
```suggestion
    // if we reach the end of paths2, then paths1 is a subfolder of paths2
```

packages/storage-ui/src/Utils/pathUtils.ts
```diff
@@ -54,4 +54,32 @@ export function getParentPathFromFilePath(filePath: string) {
...
+  // the path1 can only be a subfolder
+  // if it's longer than path2
+  if (path1.length < path2.length){
+    return result
+  }
+
+  // we loop over every section of path1
+  path1.every((section, index) => {
+    // so far every paths were identical
+    // if paths1 is longer than paths2
```
Tbaut:
```suggestion
    // if we reach the end of paths2, then paths1 is a subfolder of paths2
```

asnaith:
I just tested out the fix on Render (for Storage). It looks good for subfolders now, it's not possible to move a parent into a subfolder. 

However, on a related note I just noticed you can move the parent into itself, "Folder1" to "Folder1", and loose access in the UI. Do you want to address in this PR or should I create a separate issue? 

*Video:*

https://user-images.githubusercontent.com/3982665/125825458-17b5cccf-dd78-4040-a14f-b4cba6dca4e3.mov

packages/storage-ui/src/Components/Modules/MoveFileModal/MoveFileModal.tsx
```diff
@@ -187,29 +212,48 @@ const MoveFileModule = ({ filesToMove, modalOpen, onClose, onCancel, mode }: IMo
...
-          loading={movingFile}
-          disabled={!movePath}
-          onClick={onMoveFile}
+        {!!movePath && !isAllowedToMove && (
+          <Typography
+            component="p"
+            variant="body1"
+          >
+            {
+              folders.length
```
tanmoyAtb, line 225: The error coming up wrong,
```suggestion
              movePath !== currentPath
```

packages/files-ui/src/Components/Modules/FileBrowsers/MoveFileModal.tsx
```diff
@@ -196,17 +221,16 @@ const MoveFileModule = ({ filesToMove, modalOpen, onClose, onCancel, mode }: IMo
...
             variant="body1"
           >
-            <Trans>
-              {plural(filesToMove.length, {
-                one: "The file is already in this folder",
-                other: "The files are already in this folder"
-              })}
-            </Trans>
+            {
+              folders.length
```
tanmoyAtb, line 230: The right error comes up this way
```suggestion
              movePath !== currentPath
```

packages/files-ui/src/Components/Modules/FileBrowsers/MoveFileModal.tsx
```diff
@@ -135,11 +137,34 @@ const MoveFileModule = ({ filesToMove, modalOpen, onClose, onCancel, mode }: IMo
...
+    filesToMove.filter((f) => f.isFolder)
+  , [filesToMove])
+
+  const isSubFolderOfAnySelectedFolder = useMemo(() => {
+    if(!movePath){
+      return false
+    }
+
+    return folders.some((folder) => {
+      const currentPathWithFolder = `${currentPath}/${folder.name}`
```
tanmoyAtb: This should also solve the problem of moving into itself that @asnaith reported
Don't forget to add the import of `getPathWithFile`
```suggestion
      const currentPathWithFolder = getPathWithFile(currentPath, folder.name)
```

packages/storage-ui/src/Components/Modules/MoveFileModal/MoveFileModal.tsx
```diff
@@ -135,6 +136,30 @@ const MoveFileModule = ({ filesToMove, modalOpen, onClose, onCancel, mode }: IMo
...
+    filesToMove.filter((f) => f.isFolder)
+  , [filesToMove])
+
+  const isSubFolderOfAnySelectedFolder = useMemo(() => {
+    if(!movePath){
+      return false
+    }
+
+    return folders.some((folder) => {
+      const currentPathWithFolder = `${currentPath}/${folder.name}`
```
tanmoyAtb: same here
```suggestion
      const currentPathWithFolder = getPathWithFile(currentPath, folder.name)
```

tanmoyAtb (changes requested):
Amazing, 
A few comments to resolve some of the problems.

packages/files-ui/src/Components/Modules/FileBrowsers/MoveFileModal.tsx
Tbaut: I don't think the error is wrong, in which case do you have it wrong? To have better error, we should check whether we are talking about folders, files, or mixed. In that case, I'm checking whether there's at least one folder selected, in which case I talk about folders, otherwise about files. I think the error are half wrong then.

packages/files-ui/src/Components/Modules/FileBrowsers/MoveFileModal.tsx
Tbaut: Thanks a lot, it's def. the way to go and solves Andres's problem from what I can see.

asnaith:
Top level seems good now, I cannot move a parent into a child folder or itself. 

I know this is expanding the scope of the original ticket but I've found a couple of additional issues that are related to this, regarding child folders. You can cause a child folder to become unreachable in two scenarios

- When moving it into the parent folder that it is already in 
- When when moving it to itself

Can we show the same error when trying to move child folders in this way too?

Tbaut:
>    When moving it into the parent folder that it is already in
    When when moving it to itself

I can't reproduce this, it's not allowed for me. Can you do another round of check @tanmoyAtb and @FSM1 when you have the chance? I'd love to get to the bottom of it today :)

Here's what I'm testing, at the root, and in a sub-folder, and it looks good.

https://user-images.githubusercontent.com/33178835/125917900-add45f06-d037-48ab-88af-2569583f2adc.mp4

packages/files-ui/src/Components/Modules/FileBrowsers/MoveFileModal.tsx
tanmoyAtb: You're right, this change is not needed as I think about it now,
I was going for `The files are already in this folder`, when you have folders selected,
but `You can't move into this folder` works for me.

tanmoyAtb: Now i feel whats going on here, And the change is not needed.
I was going for `The files are already in this folder` even for folder,
but `You can't move into this folder` works just fine!

tanmoyAtb:
For me both of these cases are working as expected,
I'm not allowed to,
- move into one's parent folder
- move onto itself

tanmoyAtb (approved):
Tested out all the cases I could think of,
working for me 💯

FSM1 (approved):
Also tested all the scenarios I could think of and seem unable to break it 🎉

PR #1290, opened 2021-07-16
Trim before creation and renaming file, folders and buckets

### LINKED ISSUE #1270: ChainSafe Storage: User can create a bucket with space character as the name
**Description**
It is possible to create a bucket with special characters in the name, whilst this is not so bad in most instances it's particularly weird when we allow a single space character as the name because it looks like a blank bucket in the bucket rows. 

**Steps to Reproduce**
1. Login to Storage
3. Click the "Create Bucket" button
4. Enter a single space character as the name

**Observed behavior**
A bucket is created with the space character as the name 

**Expected behavior**
It should not be possible to create a bucket with only a space character as the name

**Screenshots**
Screen Shot 2021-07-14 at 2 24 03 PM

 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

### LINKED ISSUE #1271: ChainSafe Storage: Create bucket modal error is retained after closing the dialog
**Description**
After invoking an error when trying to create a bucket if you close the modal and re-open it the same error message will be displayed. 

**Steps to Reproduce**
1. Login to Storage
3. Click the "Create Bucket" button
4. Without entering a name click the "Create" button to invoke "Bucket name is required" error message
5. Click cancel on the create bucket modal 
6. Click the "Create Bucket" button

**Observed behavior**
When re-opened the modal will be displaying the previous error that was invoked but as the user has not yet entered a name this should not be the case. 

**Expected behavior**
No previously invoked errors should be displayed when the modal is opened 

 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

Tbaut:
closes #1270 
closes #1271


asnaith:
Using the original repro steps I was still able to create a blank bucket by using just a single space as the name on render. 

It is not possible to highlight the space character in the bucket name row so it looks like it is being trimmed successfully but don't we want to prevent people creating blank buckets? 

Screen Shot 2021-07-16 at 2 52 07 PM

Tbaut:
I read your issue too quickly. As the PR originally stated, it trimmed for *rename* but the problem is actually there for any folder create/rename and file rename, bucket creation/rename. Should be good now also I found that there's no check for new shared folder. I'll create an issue.

Tbaut:
I saw that the button is still enabled but the bucket creation is prevented at least. I'll leave this PR in this state for now as it's late for me. Someone can pick it up or we can make an issue out of it.

RyRy79261 (approved):

tanmoyAtb (approved):
looks good!

packages/storage-ui/src/Utils/validationSchema.ts
```diff
@@ -2,22 +2,28 @@ import { t } from "@lingui/macro"
 import { object, string } from "yup"
 import CID, { isCID  } from "cids"
 
+// eslint-disable-next-line 
+const spacesOnlyRegex = new RegExp(`^\s+$`)
+
 export const renameSchema = object().shape({
   fileName: string()
-    .trim()
```
Tbaut: Why did this trim (and the others) get removed @FSM1?
The input is trimmed when sent to the api, hence they should be trimmed for the check as well isn't it?

FSM1: We should not be manipulating strings that a user enters unnecessarily. If a user wants to have a bucket, folder or file named ```      whatever``` this should be their prerogative.

Tbaut: > If a user wants to have a bucket, folder or file named whatever this should be their prerogative.

We explicitly prevent that with a trim before making the api call though! If we trim the name sent to the api, I don't see the problem in applying the same in validation rather than adding specific edge-cases.

> We should not be manipulating strings that a user enters unnecessarily.

If you're talking about performances, I guess a regex check is more costly, on top of being less easy to read. I just don't see the point in this edge-case, where a trim just solves it all, in a simpler manner.

PR #1300, opened 2021-07-20
S3 API Keys 

### LINKED ISSUE #1294: Chainsafe Storage: Allow users to create S3 API keys
Users require API keys compatible with AWS S3

A user would click an `Add S3 API Key`  button to create an S3 compatible API key, and the API Key and Secret would be displayed to the user.

FSM1:
Closes #1294


asnaith:
Nice! Looks good on render 🙂.

Only feedback I have are more UX considerations. 

- I thought it might perhaps be better to make the "Make sure to save the secret, as it can only be displayed once." text (on the new modal) more prominent to the user (different colour or bold). 

- Maybe it makes sense to add sorting options to the columns now that we can create more than one key (future ticket?).

tanmoyAtb (approved):
Looks all good to go!

FSM1:
> * I thought it might perhaps be better to make the "Make sure to save the secret, as it can only be displayed once." text (on the new modal) more prominent to the user (different colour or bold).

Can try play around with this.

> * Maybe it makes sense to add sorting options to the columns now that we can create more than one key (future ticket?).

Dont see the value in this right now. Let's wait and see if any users actually ask for this, or we can add it to Cindy's user group questions in future.

RyRy79261 (approved):

PR #1302, opened 2021-07-20
Ensure bucket names are unique

### LINKED ISSUE #1272: ChainSafe Storage: No character limit enforced on bucket name
**Describe the bug**
Currently we are not restricting the amount of characters that can be entered for a bucket name

**Steps to Reproduce**
1. Login to Storage
3. Click the "Create Bucket" button
4. Enter a ridiculously large string for the bucket name
6. Click the "Create Bucket" button

**Observed behavior**
The bucket with huge name will be created and whilst the bucket browser handles this well (it just truncates the string) when you navigate inside the bucket it will try and display the large name and cause a horizontal scroll bar to appear. This can cause the name to overlap over the side navigation bar. 

**Expected behavior**
It would be best to enforce a limit on the amount of characters that can be entered to prevent other weird display (and perhaps even API) issues that might arise from using a huge string as the name. 

**Screenshots**
Bucket Browser view
Screen Shot 2021-07-14 at 2 30 07 PM

Inside bucket view
Screen Shot 2021-07-14 at 2 29 25 PM

**Desktop**
 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

### LINKED ISSUE #1297: ChainSafe Storage: No error / info when creating a bucket with the same name as an existing bucket
**Description**
There is no feedback given to the user when attempting to create a bucket with the same name as a pre-existing bucket. 
Upon clicking "Create" the dialog closes but nothing happens because the bucket is already there. 

**Steps To Reproduce**
1. Navigate to the buckets page
2. Create a new bucket with a unique  #name
3. Create another bucket using the same name that was used previously 

**Expected behavior**
Inform the user on the modal that a bucket with the same name already exists

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

FSM1:
Closes #1297

### DIFF
--- a/packages/storage-ui/src/Components/Elements/BucketRow.tsx
+++ b/packages/storage-ui/src/Components/Elements/BucketRow.tsx
@@ -72,7 +72,7 @@ const BucketRow = ({ bucket }: Props) => {
     >
       <TableCell className={classes.name}
         onClick={() => redirect(ROUTE_LINKS.Bucket(bucket.id, "/"))}>
-        {bucket.name}
+        {bucket.name || bucket.id}
       </TableCell>
       <TableCell>
         {formatBytes(bucket.size)}
--- a/packages/storage-ui/src/Components/Pages/BucketsPage.tsx
+++ b/packages/storage-ui/src/Components/Pages/BucketsPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react"
+import React, { useMemo, useState } from "react"
 import { makeStyles, createStyles } from "@chainsafe/common-theme"
 import {
   Button,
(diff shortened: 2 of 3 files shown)


tanmoyAtb (approved):
looks great!

On a sidenote:
I feel these types of errors should come from the API!

FSM1:
> looks great!
> 
> On a sidenote:
> I feel these types of errors should come from the API!

The API does throw on these cases, but it makes sense to duplicate the validation logic in UI, to prevent the user from making requests which we know will fail.

RyRy79261 (approved):

PR #1307, opened 2021-07-21
Resolve move folder glitches

### LINKED ISSUE #1301: ChainSafe Storage: Child folder disappears from UI after illogical move allowed
**Description**
It is possible to move a child folder into itself or the parent folder it is already in and after doing this it will disappear from the UI and all access to it is lost. 

**Steps to Reproduce**
1. Login to ChainSafe storage
2. Create a new bucket and inside the bucket create a "parent" folder
3. Inside the "parent" folder create a "child" folder
4. Attempt to move the child folder into the parent folder it is already in / or two to itself 

**Expected behavior**
When trying to move a child folder into it's parent folder or itself the modal should display a message informing the user "You can't move folders to this path" and the "Move" button should be greyed out. 

**Video**

https://user-images.githubusercontent.com/3982665/126390716-c7045929-330d-4411-83a9-f04912339a50.mov

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114 

**Additional context**
Probably an additional and currently uncounted for scenario following on from the work done in #1276

FSM1:
Closes #1301

### DIFF
--- a/packages/storage-ui/src/Components/Modules/FilesList/FilesList.tsx
+++ b/packages/storage-ui/src/Components/Modules/FilesList/FilesList.tsx
@@ -38,7 +38,7 @@ import UploadProgressModals from "../UploadProgressModals"
 import DownloadProgressModals from "../DownloadProgressModals"
 import CreateFolderModal from "../CreateFolderModal/CreateFolderModal"
 import UploadFileModule from "../UploadFileModal/UploadFileModal"
-import MoveFileModule from "../MoveFileModal/MoveFileModal"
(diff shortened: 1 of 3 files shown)


asnaith (approved):

asnaith:
Nice fix Mike, I've just tested out the 4 scenarios below and all is looking good: 

User cannot move a parent folder into a child folder 
User cannot move a parent folder into itself
User cannot move a child folder into the parent folder it is already in
User cannot move a child folder into itself

RyRy79261 (approved):

PR #1310, opened 2021-07-22
Fixing storage empty folder select

### LINKED ISSUE #1295: ChainSafe Storage: Multiple folders selected in the UI after only clicking one
**Description**
It's not possible to select a single folder if you have multiple empty folders at the root of the bucket. If you try to select a single folder then you will see multiple being selected. As soon as you add content to at least one of the folders then you can select single folders again. 

This only affects folder selection, no problems with file selection. 

**Steps to Reproduce**
1. Login to ChainSafe Storage
2. Create a new bucket
3. Create two empty folders at the root of the bucket
4. Observe that when you try to click a single folder that both are selected
5. Navigate into one of the folders and add something to it (can be folder or file) 
6. Return to the bucket root
7. Observe that now when you try to click a single folder only that one is selected

**Observed behavior**
Multiple folders will be selected when the user is only clicking one (unless one of the folders has content). 

**Expected behavior**
Only one folder should be selected regardless of wether it has contents or not. 

**Video** [Updated]

https://user-images.githubusercontent.com/3982665/126386536-3974ecce-bbe6-4e9b-aa4a-995ae2d43e47.mov

**Desktop:**
 - OS: macOS
 - Browser: Chrome / Firefox
 - Version : 91.0.4472.114 / 89.0.2

RyRy79261:
closes #1295


asnaith:
Nice fix Ryan! I just tested it out on Render and all is looking good with folder selection now wether they are empty or with contents.

asnaith (approved):

FSM1 (approved):
great work 🥇

PR #1316, opened 2021-07-23
Escape backslash in "white space error message" regex

### LINKED ISSUE #1314: ChainSafe Storage: User shown incorrect error after entering "sss" as bucket name
**Description**
The regex for detecting white-space only is not working and instead matching the "S" character

**Steps to Reproduce**
1. Login
2. Create a bucket with the name "s"

**Expected behavior**
This error should only be displayed when the user has entered one or more spaces as the bucket name

**Screenshots**
Screen Shot 2021-07-22 at 3 11 23 PM

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

asnaith:
closes #1314

**Before** 
Screen Shot 2021-07-22 at 9 01 59 AM
(and no error was being shown when entering just whitespace)

**After**

Screen Shot 2021-07-22 at 4 31 58 PM

Screen Shot 2021-07-22 at 4 32 24 PM
(showing as intended now)

### DIFF
--- a/packages/files-ui/src/Utils/validationSchema.ts
+++ b/packages/files-ui/src/Utils/validationSchema.ts
@@ -1,8 +1,7 @@
 import { t } from "@lingui/macro"
 import { object, string } from "yup"
 
-// eslint-disable-next-line 
-const spacesOnlyRegex = new RegExp(`^\s+$`)
+const whitespaceOnlyRegex = new RegExp("^\\s+$")
 
 export const renameSchema = object().shape({
   fileName: string()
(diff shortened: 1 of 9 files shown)


FSM1 (approved):
great catch @asnaith, thanks for taking the initiative to fix this.

tanmoyAtb (approved):
Great work !

Tbaut:
I think we should systematically trim anything the user type (appart from the password probably). This would make this specific check not needed. I had started in https://github.com/ChainSafe/ui-monorepo/pull/1290/files but missed many as it ended up being a bigger task that I originally thought.

edit: ok I see this got modified in https://github.com/ChainSafe/ui-monorepo/pull/1290#pullrequestreview-714715500 adding a regex that I think could be avoided.

PR #1323, opened 2021-07-23
Dynamic limits added

### LINKED ISSUE #1319: Intergrate dynamic user limits
The buckets summary endpoint needs to be called when the user logs in and when refreshing a path

RyRy79261:
closes #1319


packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -117,9 +118,14 @@ const FilesProvider = ({ children }: FilesContextProps) => {
   } = useFilesApi()
   const { publicKey, encryptForPublicKey, decryptMessageWithThresholdKey } = useThresholdKey()
   const { addToastMessage } = useToaster()
-  const [spaceUsed, setSpaceUsed] = useState(0)
   const [personalEncryptionKey, setPersonalEncryptionKey] = useState<string | undefined>()
   const [buckets, setBuckets] = useState<BucketKeyPermission[]>([])
+  const [storageSummary, setBucketSummary] = useState<BucketSummaryResponse>({
```
tanmoyAtb: We could have this value initialized as `undefined`,
otherwise getting a flash of 0 of 0 bytes!
Screenshot 2021-07-23 at 8 58 40 PM

tanmoyAtb:
I have a feeling the `formatBytes` is using base 10 but the back end is sending numbers in base 2,
Seeing 5.37GB here, which is 5GB if we use base 2!

Screenshot 2021-07-23 at 9 00 32 PM

packages/files-ui/src/Contexts/FilesContext.tsx
FSM1:
```suggestion
  const [storageSummary, setStorageSummary] = useState<BucketSummaryResponse>({
```

packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -193,20 +200,16 @@ const FilesProvider = ({ children }: FilesContextProps) => {
...
+    const getStorageSummary = async () => {
       try {
-        const totalSize = buckets.filter(b => b.type === "csf" ||
-          b.type === "trash" ||
-          (b.type === "share" && !!b.owners.find(u => u.uuid === profile?.userId)))
-          .reduce((totalSize, bucket) => totalSize += bucket.size, 0)
-
-        setSpaceUsed(totalSize)
+        const bucketSummaryData = await filesApiClient.bucketsSummary()
+        setBucketSummary(bucketSummaryData)
```
FSM1:
```suggestion
        setStorageSummary(bucketSummaryData)
```

packages/files-ui/src/Contexts/FilesContext.tsx
FSM1: Agreed that using `undefined` would work well here, and then if Summary is is undefined, just don't render the progress bar at all.

FSM1 (approved):

FSM1 (approved):

tanmoyAtb (approved):

PR #1342, opened 2021-07-26
Make sure to always use helpers with paths

Tbaut:
- enforce the use of `getPathWithFile`
- create a `getSafePAthUrlWithFiles` to prevent duplication


RyRy79261 (approved):

FSM1 (approved):
Thanks for neatening this up @Tbaut 🥇

PR #1351, opened 2021-07-27
Remove submit button for renaming and useFormik everywhere

### LINKED ISSUE #1345: Files: Rename changes are not saved when clicking the checkmark icon
**Description**
Attempts to rename files and folders do not save when clicking the checkmark icon

(If the user presses [Enter] when in edit mode the changes will save)

**Steps to Reproduce**
1. Login to Files
2. Create a folder and upload a file
3. Click the "rename" option for a file or folder
4. Enter a new name and click the checkmark icon. 

**Observed behavior**
After selecting the rename option the user can enter a name but the changes will not save after clicking the checkmark icon. 

**Expected behavior**
The user should be able to rename a file or folder by clicking the checkmark icon 

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Version : 91.0.4472.114

Tbaut:
closes #1345

The reason for the bug was a conflict with the blur action taking precedence.
I removed the little submit button altogether, (cc @sweetpea22 ) to have a consistent UI between grid and list view.
Submitting a rename works with hitting Enter just like on a desktop file explorer. Clicking anywhere outside cancels the action, unlike the behavior on a desktop computer.


RyRy79261 (approved):

FSM1:
Any specific reason for removing the button from the desktop view instead of wiring it up correctly so that it actually works as intended? It might not be obvious to users that they need to hit the `enter` key to submit, which is why we originally included the button there.

Tbaut:
It's a mix between being lazy, trying to be consistent, and trying to match with what users are used to.
1. I couldn't manage to make both work, the blur and the click on the submit button
2. I realized that the grid view doesn't have a submit button
3. I realized that on a desktop OS, you don't have a submit button, you just hit enter or click outside.

So maybe we should actually submit the change (rename) on blur to match with the behavior on an OS. I found that weird so I didn't go for it, and kept the "cancel" behavior on blur, to play it safe.

FSM1:
> It's a mix between being lazy, trying to be consistent, and trying to match with what users are used to.
> 
> 1. I couldn't manage to make both work, the blur and the click on the submit button
> 2. I realized that the grid view doesn't have a submit button
> 3. I realized that on a desktop OS, you don't have a submit button, you just hit enter or click outside.
> 
> So maybe we should actually submit the change (rename) on blur to match with the behavior on an OS. I found that weird so I didn't go for it, and kept the "cancel" behavior on blur, to play it safe.

You raise a lot of valid points. I also feel that submit on blur is a bit too bold. 

Maybe the solution here is to add some text at the end of the input, to make sure users know what to do: `press ENTER` or `ENTER to confirm`

Tbaut:
I'm not sure how a helper text would look like for this, I thought about a tooltip but we use those for errors already so it shouldn't be used as an info box as well. Do you think that typing "enter" would feel unusual for our users?

FSM1:
> I'm not sure how a helper text would look like for this, I thought about a tooltip but we use those for errors already so it shouldn't be used as an info box as well. Do you think that typing "enter" would feel unusual for our users?

Yeah agreed that pressing enter does feel quite natural. If we have users asking about how to rename things, we can always solve that problem when it happens. I will look for some good examples of what I mean by helper text, and screenshot them if I find something that looks good.

FSM1 (approved):

Tbaut:
Ah I think I need to update a test, it's failing for legitimate reasons, I'll merge after that.

asnaith:
I noticed that after you have clicked away to cancel the rename operation when you attempt to rename a file again you will see that the cancelled name was retained in the input field despite it never being set. 

https://user-images.githubusercontent.com/3982665/127208676-eada8f75-2b1e-45c1-b5fe-cc86ce69ad73.mov

Tbaut:
Ah thanks, I'll address that.

Tbaut:
omg it's passing. I added a test explicitly for what you mentioned @asnaith

asnaith:
@Tbaut Nice! Thank you for doing that.

PR #1353, opened 2021-07-27
create and edit shared folder modals merge

### LINKED ISSUE #1279: Combine Create and update Shared folder modal
Currently there are 2 separate modals for creating shared folder modal and update shared folder modal.
We want one modal with both input tags for `readers` and `writers`

tanmoyAtb:
closes #1279


packages/files-ui/src/Components/Modules/FileBrowsers/SharedFoldersOverview.tsx
```diff
@@ -224,24 +225,21 @@ const SharedFolderOverview = () => {
...
-        isModalOpen={isEditBucketModalOpen}
+      <CreateOrEditSharedFolderModal
+        mode={createOrEditSharedFolderMode}
+        isModalOpen={!!createOrEditSharedFolderMode}
         onClose={() => {
-          setIsBucketModalOpen(false)
           setBucketToEdit(undefined)
+          setCreateOrEditSharedFolderMode(undefined)
         }}
         bucket={bucketToEdit}
```
Tbaut: Could we name it "bucketToEdit"?

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFoldersOverview.tsx
```diff
@@ -153,7 +151,10 @@ const SharedFolderOverview = () => {
           </Typography>
           <div className={classes.controls}>
             <Button variant='outline'
```
Tbaut, line 153: new line

packages/files-ui/src/Components/Modules/FileBrowsers/CreateOrEditSharedFolderModal.tsx
```diff
@@ -111,7 +116,8 @@ const useStyles = makeStyles(
   }
 )
 
-interface IUpdateSharedFolderModalProps {
+interface ICreateOrEditSharedFolderModalProps {
+  mode?: "create" | "edit"
```
Tbaut: 100% bonus to prevent duplication, we could have some fancy typescript, without going too crazy.
```
// exported in SharedFoldersOverview.tsx
export type = SharedFolderModalMode = "create" | "edit" | undefined

// imported here and used on this line
mode: Exclude<SharedFolderModalMode, undefined>
```

Tbaut (changes requested):
I started reviewing but quickly realized that we're duplicating here a lot of logic that was extracted in `useCreateSharedFolder` and `useLookupUser` explicitly for this purpose. We should reuse them. edit: take a look at the file that got deleted here, `CreateSharedFolderModal.tsx` was using them.

Also while I tested the functionality to create a shared folder, I found out about the following
- type in a user "abc" as reader
- type in a user "abc" as writer
- the "reader" is automatically removed and a text appears.
- I didn't realize this user was removed right away

We should not make choices for our users. Instead we should have the "create" button disabled, and the error text say `User "abc" (maybe truncated) cannot be added both as reader and writer`. And let our user chose where this one should go.

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFoldersOverview.tsx
Tbaut: variant should be on a new line (I can't suggest because it's not an edited line)

packages/files-ui/src/Components/Modules/FileBrowsers/CreateOrEditSharedFolderModal.tsx
tanmoyAtb: Placed the types in the types file!

FSM1 (approved):
works great 💯

tanmoyAtb:
> I started reviewing but quickly realized that we're duplicating here a lot of logic that was extracted in `useCreateSharedFolder` and `useLookupUser` explicitly for this purpose. We should reuse them. edit: take a look at the file that got deleted here, `CreateSharedFolderModal.tsx` was using them.
> 
> Also while I tested the functionality to create a shared folder, I found out about the following
> 
> * type in a user "abc" as reader
> * type in a user "abc" as writer
> * the "reader" is automatically removed and a text appears.
> * I didn't realize this user was removed right away
> 
> We should not make choices for our users. Instead we should have the "create" button disabled, and the error text say `User "abc" (maybe truncated) cannot be added both as reader and writer`. And let our user chose where this one should go.

Okay, so we have updates on checking the error on intersecting users. 
I made a number of changes to reuse `useLookupUser` and `useCreateOrEditSharedFolder`. 

So this PR would effect creating shared folder, update permissions and copy to shared folder.

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
```diff
@@ -29,9 +39,10 @@ export const useLookupSharedFolderUser = () => {
 
       if (!result) return []
 
-      const currentUsers = Array.isArray(sharedFolderUsers)
-        ? sharedFolderUsers.map(su => su.value)
-        : []
+      if (!result) return []
```
Tbaut: duplicate
```suggestion
```

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
```diff
@@ -1,15 +1,25 @@
...
 import { useCallback, useState } from "react"
 import { useFilesApi } from "../../../../Contexts/FilesApiContext"
 import { useUser } from "../../../../Contexts/UserContext"
-import { SharedFolderCreationUser } from "../types"
+import { centerEllipsis } from "../../../../Utils/Helpers"
+import { SharedUserTagData, SharedFolderUserPermission } from "../types"
+import {
+  ITagOption,
+  ITagValueType
+} from "@chainsafe/common-components"
```
Tbaut: nit
```suggestion
import { ITagOption, ITagValueType } from "@chainsafe/common-components"
```

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
```diff
@@ -29,9 +39,10 @@ export const useLookupSharedFolderUser = () => {
 
       if (!result) return []
 
-      const currentUsers = Array.isArray(sharedFolderUsers)
-        ? sharedFolderUsers.map(su => su.value)
-        : []
+      if (!result) return []
+      const usersList = permission === "read" ? sharedFolderReaders : sharedFolderWriters
+      const currentUsers = Array.isArray(usersList) ? usersList.map(su => su.value) : []
```
Tbaut: No need to check if it's an array, the type indicated that it must be.
```suggestion
      const currentUsers = usersList.map((su) => su.value)
```

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
```diff
@@ -29,9 +39,10 @@ export const useLookupSharedFolderUser = () => {
...
+      if (currentUsers.includes(result.uuid) || result.uuid === profile?.userId) return []
```
Tbaut: Also duplicated
```suggestion
```

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
```diff
@@ -1,15 +1,25 @@
...
+  ITagOption,
+  ITagValueType
+} from "@chainsafe/common-components"
 
 export const useLookupSharedFolderUser = () => {
   const { filesApiClient } = useFilesApi()
-  const [sharedFolderUsers, setSharedFolderUsers] = useState<SharedFolderCreationUser[]>([])
+  const [sharedFolderReaders, setSharedFolderReaders] = useState<SharedUserTagData[]>([])
+  const [sharedFolderWriters, setSharedFolderWriters] = useState<SharedUserTagData[]>([])
+  const [hasPermissionsChanged, setHasPermissionsChanged] = useState(false)
```
Tbaut: I wonder if this needs to be here. This seems to be related to a form that should be checked with dirtiness.
edit: After a check, unless I'm mistaken, it's only used in the `CreateOrEdit..Modal.tsx`, and should be added to the `TagsInput` onChange. This can be a `useState` in this component only and reduce a little the size and complexity of this hook

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
```diff
@@ -41,7 +52,53 @@ export const useLookupSharedFolderUser = () => {
       console.error(e)
       return Promise.reject(e)
     }
-  }, [filesApiClient, sharedFolderUsers, profile])
+  }, [filesApiClient, sharedFolderReaders, sharedFolderWriters, profile])
+
+  const onNewReaders = (val: ITagValueType<ITagOption, true>) => {
```
Tbaut: onNewReaders and onNewWriters are super similar, what do you think about unifying them and have an additional `permission` parameter, just like the handleLookUp?

packages/files-ui/src/Components/Modules/FileBrowsers/CreateOrEditSharedFolderModal.tsx
```diff
@@ -111,143 +113,82 @@ const useStyles = makeStyles(
...
+  }, [onClose, setSharedFolderReaders, setSharedFolderWriters])
 
-    const readers = getUserPermission(sharedFolderReaders)
-    const writers = getUserPermission(sharedFolderWriters)
-
-    setIsEditingSharedFolder(true)
-    editSharedFolder(bucket, writers, readers)
+  const onCreateSharedFolder = useCallback(() => {
+    handleCreateSharedFolder(sharedFolderName, sharedFolderReaders, sharedFolderWriters)
       .then(handleClose)
```
Tbaut:
```suggestion
      .catch(console.error)
      .finally(handleClose)
```

packages/files-ui/src/Components/Modules/FileBrowsers/CreateOrEditSharedFolderModal.tsx
```diff
@@ -111,143 +113,82 @@ const useStyles = makeStyles(
...
-    }
-    setHasPermissionsChanged(true)
-    val && val.length > 0
-      ? setSharedFolderWriters(val?.map(v => ({ label: v.label, value: v.value, data: v.data })))
-      : setSharedFolderWriters([])
-  }
+  const onEditSharedFolder = useCallback(() => {
+    if (!bucketToEdit) return
+    handleEditSharedFolder(bucketToEdit, sharedFolderReaders, sharedFolderWriters)
+      .then(handleClose)
```
Tbaut:
```suggestion
      .catch(console.error)
      .finally(handleClose)
```

Tbaut (changes requested):
There a lot of serioulsy good things in there, the mix of `.filter` and `.some` is :exploding_head: 
Found a couple nits while looking at the code, will play with it now.

Tbaut:
We should check with https://github.com/ChainSafe/ui-monorepo/pull/1356 this is going to conflict (I checked the dark mode and it doesn't adapt well indeed)

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
tanmoyAtb: Yeap, moved the logic back into `createOrEdit..` file!

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
tanmoyAtb: Yes! merged the functions into one!

tanmoyAtb:
> We should check with #1356 this is going to conflict (I checked the dark mode and it doesn't adapt well indeed)

going to wait till the dark mode comes through! going to be quite some conflicts !

tanmoyAtb:
So looks like we're good to go with the merge with dark mode components!
Removed a couple of warnings in the PR!

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useLookupUser.tsx
Tbaut: you forgot about this one, committing it now.

Tbaut (approved):
I've removed the external `setSharedFolderWriters` and writers because these were redundant with the `setNewUsers`. Moved the logic to check for errors in a `useEffect` and added a `useCallback` to avoid infinite loop. 

I've tested it as much as I could and it looked good.

PR #1358, opened 2021-07-27
Fixed Storage summary updating

### LINKED ISSUE #1355: Storage used should be updated whenever the value changes
The current `storage used` value is only updated when the user logs in to the app. This creates inconsistencies between what the user see's as their space used until they log in to the app again, or refresh the page.

Any operation that impacts the amount of storage used should query the API for this value once the operation has been completed.

Currently I can think of the following operations:
- Delete bucket
- Add pin
- Remove pin
- upload file
- delete file from Bin
- upload file to share you own
- delete file from share you own

Alternative approaches:
- Poll for this data - 👎🏼 Lots of unnecessary requests
- Return the new space used value with any request that modifies the value - 👍🏼 minimise api requests 👎🏼 Could make API architecture more complex than necessary 🤷🏼

RyRy79261:
closes #1355


asnaith:
Hey Ryan, I checked this out on the render links for storage and files. The below is a list of all the data changing actions I could think of. The storage summary is updated after some of these actions but not all at the moment. 

**Storage** 

🟩 - User should see storage summary adjust after uploading a file to bucket
🟩 - User should see storage summary adjust after deleting a file from bucket
🟥 - User should see storage summary adjust after deleting a bucket containing files
🟥 - User should see storage summary adjust after adding a pin
🟥 - User should see storage summary adjust after removing a pin

**Files**

🟩 - User should see storage summary adjust after uploading a file in home
🟥 - User should see storage summary adjust after deleting a file in home *(possibly an invalid scenario, it goes to the bin?)
🟩 - User should see storage summary adjust after deleting file in bin
🟩 - User should see storage summary adjust after uploading file to owned shared folder
🟥 - User should see storage summary adjust after deleting file in owned shared folder
🟥 - User should see storage summary adjust after deleting shared folder containing files

RyRy79261:
@asnaith 

So for the storage side of things, when ever a refresh is triggered, like in the case of deleting buckets, the get summary is called as it is part of the refresh bucket call that updates the list, removing the bucket in question

However the buckets haven't been removed immediately, they are in state of `Deleting` by the time the summary is updated. Since it hasn't been removed immediately, the summary wouldn't update immediately 

image

RyRy79261:
@asnaith I've updated the Files section to call the get summary for more functions, should cover most of it now.

asnaith:
@RyRy79261 Great work! Thanks for clarifying, what you describe with bucket deletion is exactly what I'm seeing (and pinning / unpinning because of the queue system). The storage summary will update after a change but only after a browser refresh. I presume for Storage this is the way it will be staying too as the only other work around would be to constantly request status? 

As for files, all of those test scenarios look good now. So to summarize 

**Storage** 

🟩 - User should see storage summary adjust after uploading a file to bucket*
🟩 - User should see storage summary adjust after deleting a file from bucket*
🟩  - User should see storage summary adjust after deleting a bucket containing files*
🟩 - User should see storage summary adjust after adding a pin*
🟩  - User should see storage summary adjust after removing a pin*

*All after browser refresh

**Files**

🟩 - User should see storage summary adjust after uploading a file in home
🟩 - User should see storage summary adjust after deleting file in bin
🟩 - User should see storage summary adjust after uploading file to owned shared folder
🟩  - User should see storage summary adjust after deleting file in owned shared folder
🟩  - User should see storage summary adjust after deleting shared folder containing files

asnaith:
@RyRy79261 Great work! Thanks for clarifying, what you describe with bucket deletion is exactly what I'm seeing (and when pinning / unpinning because of the queue system and deleting entire shared folders containing files). The storage summary will still update after a change but only after a browser refresh for those scenarios. I presume for Storage this is the way it will be staying too as the only other work around would be to constantly request status? 

To summarize these are the scenarios that all look good now and I've left an asterix beside the scenarios which require a browser refresh. 

**Storage** 

🟩 - User should see storage summary adjust after uploading a file to bucket
🟩 - User should see storage summary adjust after deleting a file from bucket
🟩  - User should see storage summary adjust after deleting a bucket containing files*
🟩 - User should see storage summary adjust after adding a pin*
🟩  - User should see storage summary adjust after removing a pin*


**Files**

🟩 - User should see storage summary adjust after uploading a file in home
🟩 - User should see storage summary adjust after deleting file in bin
🟩 - User should see storage summary adjust after directly uploading file to owned shared folder
🟩 - User should see storage summary adjust after copying file to owned shared folder
🟩 - User should see storage summary adjust after deleting file in owned shared folder
🟩 - User should see storage summary adjust after deleting shared folder containing files* 

*After browser refresh

Tbaut (approved):
looks :100:

asnaith (approved):

PR #1360, opened 2021-07-27
[STORAGE] Added filtering for filesystem type

### LINKED ISSUE #1354: Update file browser to handle IPFS & FPS buckets
Items uploaded to `ipfs` buckets do not have a `created at` value, since there is nowhere to store this data using the IPFS file system. Due to this the Storage file browser always displays blank value for `created at`.

While we do not currently allow users to create `fps` buckets using the UI, buckets created using the S3 API will be created with the `fps` file system, and as such items uploaded would have a `created at` value.

The storage file browser should be extended to be aware of the bucket file system type, and if the bucket file system type is `chainsafe`, then the column for Created At should be displayed. For `ipfs` file system buckets, the column should be hidden.

RyRy79261:
closes #1354


Tbaut (changes requested):
We're doing quite some props drilling. Could we get the file system type directly where we use it? I'm thinking about a small `useIsIPFSFileSystem` custom hook hat would get the current bucket, and return `fileSystemType === "ipfs"` (a boolean), and use this directly where it's needed. What do you think?

FSM1 (approved):
While the prop drilling isn't ideal, I dont feel that adding an extra hook is right here either. Could the value not be retrieved from the useFileBrowser `bucket` property?

Either case, I think given the mass of props already being passed down, one more in this case should not hurt too much.

Tbaut:
Absolutely, the context is there, let's use it and prevent those walls of props.
I am always a little afraid to see that we copy/paste and generally duplicate more than we should, but I agree that for this one a custom hook is more the sugar on the cake.

RyRy79261:
Yeah the main idea avoiding using a hook since it felt like adding another method of getting variables in addition to the brick wall of props was more inconsistent. Can swap to the hook if preferred

packages/storage-ui/src/Contexts/FileBrowserContext.tsx
```diff
@@ -38,6 +38,7 @@ interface FileBrowserContext extends IFileBrowserModuleProps {
   getPath?: (cid: string) => string
   isSearch?: boolean
   withSurvey?: boolean
+  fileSystemType: FileSystemType | undefined
```
Tbaut, line 41: This can be removed
```suggestion
```

packages/storage-ui/src/Components/Pages/BucketPage.tsx
```diff
@@ -221,7 +221,8 @@ const BucketPage: React.FC<IFileBrowserModuleProps> = () => {
       controls: true,
       allowDropUpload: true,
       itemOperations,
-      withSurvey: showSurvey
+      withSurvey: showSurvey,
+      fileSystemType: bucket?.file_system_type
```
Tbaut, line 225: And this one too.
```suggestion
```

Tbaut (approved):
looks :100:

Issue #1381, opened 2021-07-30
Resolve issues with local session being cleared after calling cy.reload


asnaith:
**Task:**
Investigate why the local session is being cleared after `cy.reload` is called. 

**Acceptance Criteria**

- Figure out why the local session is being cleared after calling `cy.reload` (or if we have no luck resolving the problem then replace the use of it all together). It wasn't doing this originally though and we had been using that command with no issues in the past so not sure if it's 100% a cypress-only issue. 

- Remove the work around in `commands.ts` that's currently in place after `cy.reload`

Tbaut:
Removing the cy.reload altogether sounds like a good idea, I think we added it because sometimes the UI was quicker than our script and we didn't login correctly because the data wasn't picked up in the local storage. But this was certainly not the best thing to do.

Tbaut:
no more `cy.reload` since #1421, hopefully no more problem.

PR #1383, opened 2021-07-30
Add new ui tests for folder creation and deletion 

### LINKED ISSUE #1380: Create a Cypress test to add and delete folders
pre-requisite before #1374 can be done

asnaith:
closes #1380

- Needed to do this before I add the file / folder recovery work in another ticket 

- Had to add the same temp workarounds to commands.ts that I'd previously added for storage. That will be improved in a separate ticket (#1381).

- Added `safeClick` command to resolve timing issues between app / cypress when clicking elements (detached from DOM / false successful click etc). Just a temp solution until Cypress add their own ability to retry on the native click command.


FSM1 (approved):
working really well :)

Tbaut (approved):
Looking good. 
Just like the file creation, it'd be good to test the validation as well (with a /)
I'll make sure to create an issue to have the folder creation automated with a command. We shouldn't do like for upload and do it using the ui.

PR #1384, opened 2021-08-02
update bnc-onboard to 1.32.0

FSM1:
Updates WalletConnect dependency to 1.5.2


RyRy79261 (approved):

Tbaut (approved):
I thought I had approved this before. The tests aren't happy but I checked locally and it looked good

PR #1385, opened 2021-08-02
Allow recovering without any folder

### LINKED ISSUE #1382: File recovery not possible without pre-existing folders in home
**Description**

A follow on from the work done in #1365. 

If no folders exist in the home directory then it's not possible to choose the home folder itself to recover a file. If you have at least one folder in home then there is no issues recovering a file.  

**To Reproduce**
1. Login to files
2. Remove all files and folders in home directory 
3. Attempt to recover a file from the bin

**Expected behavior**
User should be able to recover a file regardless of wether they have a folder in the home directory or not. 

**Screenshots**
Screen Shot 2021-07-29 at 10 20 41 PM

Tbaut:
- closes #1382
- fix an infinite loop in the modal
- better UX with a loader right away and then showing the folder tree


FSM1 (approved):
looking great 🎉

packages/files-ui/src/Components/Modules/FileBrowsers/MoveFileModal.tsx
```diff
@@ -97,40 +102,36 @@ const MoveFileModule = ({ filesToMove, modalOpen, onClose, onCancel, mode }: IMo
...
-        setFolderTree([])
-      }
-    }).catch(console.error)
+      const folderTreeNodes = [
+        {
+          id: "/",
+          title: bucketForFolderTree?.name || "Home",
+          isExpanded: true,
+          expandable: true,
+          tree: mapFolderTree(newFolderTree.entries || [])
```
Tbaut, line 122: @FSM1 I had to add this, although the type in the spec don't mention that the entries can be `null` https://github.com/ChainSafe/files-api-spec/blob/main/spec/v1/v1.spec.yml#L2701
I'll add an issue in the api repo to make sure an empty array is returned when there's no entry.

Tbaut: https://github.com/ChainSafe/files-api/issues/1383

FSM1: Perfect thanks @Tbaut

RyRy79261 (approved):

PR #1386, opened 2021-08-02
Upgarde to cypress v8.1

Tbaut:
@asnaith I removed the `--browser chrome` in ci so that it doesn't pop-up when running it locally. I don't think we saw any difference adding it back then. lmk if you think otherwise.


FSM1 (approved):

RyRy79261 (approved):

asnaith:
@Tbaut Without the chrome flag the tests will get executed in electron which most (maybe all) our users don't use. I think it's better to force the tests to run in chrome.

Tbaut:
The thing is electron is based on chromium, just like chrome and I thought if it's the default it's it's better suited to run headless in the CI. But I actually found what I wanted, my problem was that chrome was popping up, but we can run it headless, so I'll add it back with that.

edit: https://github.com/ChainSafe/ui-monorepo/pull/1402

PR #1390, opened 2021-08-02
Add validation for new shared folder

### LINKED ISSUE #1293: Use some validation for Shared folder creation
The shared folder creation has no validation. We should make sure to either use formik or to have the same validation as for bucket creation for storage-ui (trim, min 1, max 65 char, no / in the name).

https://github.com/ChainSafe/ui-monorepo/blob/dev/packages/files-ui/src/Components/Modules/FileBrowsers/CreateSharedFolderModal.tsx#L116-L120

Tbaut:
closes #1293 
- refactor to use only 1 name validator (it was redundant with only small error changes)
- remove the specific whitespace error while trim handles this nicely


RyRy79261 (approved):

tanmoyAtb (commented):
Works great! 
We just need to add the validator to create shared folder during share file!

Tbaut:
Ah thanks I thought I had that in. I'll check it out.

Tbaut:
Should be good now @tanmoyAtb

asnaith:
Looking good but I noticed these changes also stop the user from having a legitimate space in the name, for example "My Share" is no longer possible. 

🟩 User cannot create a folder with name being white space only
🟩 User cannot create a folder without entering any characters for the name
🟩 User cannot create a folder with name exceeding 65 characters 
🟩 User cannot create a folder with name containing "/"
🟥 User should be able to create folder with space in the name eg "Best Share Folder"

Tbaut:
Nice catch! will fix that.

Tbaut:
ok should be good to go now

FSM1:
This PR seems to be breaking the `Rename a file` test
```
1) File management
 |        desktop
 |          can rename a file with error handling:
 |      AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy=rename-form] span.minimal.error`, but never found it.
 |       at Context.eval (http://localhost:3000/__cypress/tests?p=cypress/tests/file-management-spec.ts:495:56)
 ```
 
 
image

File: packages/files-ui/src/Components/Modules/FileBrowsers/CopyToSharedFolderModal.tsx [hunk 1]
```diff
@@ -349,17 +373,27 @@ const CopyToSharedFolderModal = ({ close, file, filePath }: IShareFileProps) =>
     </div>
   ), [classes.loadingContainer, isDownloading, isUploading])
 
-  const Step1CreateSharedFolder = useCallback(() => (
+  const Step1CreateSharedFolder = useCallback(() =>  (
```
Tbaut:
```suggestion
  const Step1CreateSharedFolder = useCallback(() => (
```

FSM1 (approved):
great stuff :)

Tbaut:
I'll stop there, thanks a lot for the thought review, this was really needed, and I'm super glad the tests caught a bug, nice catch @FSM1 !

I think this will conflict with Tanmoy's PR, I'll merge his before

PR #1395, opened 2021-08-03
Fix test with onboard bump

Tbaut:
related to #1384


FSM1 (approved):

tanmoyAtb (approved):

packages/storage-ui/cypress/support/page-objects/authenticationPage.ts
```diff
@@ -5,6 +5,7 @@ export const authenticationPage = {
 
   // get started section elements
   web3Button: () => cy.get("[data-cy=web3]", { timeout: 120000 }),
-  metaMaskButton: () => cy.get(".bn-onboard-modal-select-wallets > :nth-child(1) > .bn-onboard-custom"),
+  showMoreButton: () => cy.get("div.svelte-q1527 > .bn-onboard-custom"),
```
asnaith, line 8: @Tbaut Just to confirm - we aren't using data-cy / data-testId identifiers here because these elements are on a third party modal?

Tbaut: yup, we have unfortunately no control over those, and they will most probably break every time we update this dep.

PR #1400, opened 2021-08-03
breadcrumbs removed from settings: storage

### LINKED ISSUE #1210: ChainSafe Storage: Remove breadcrumb icon / path from the Settings screen
A suggestion to remove the breadcumb icon / path from the settings screen in **ChainSafe Storage** for consistency. 

We don't use it on the main Bucket / CIDs pages. I spoke with with @FSM1 who said the breadcrumb makes sense in file browsers, but not for general navigation.

Screen Shot 2021-06-30 at 11 33 44 AM

Note: We do want to keep it when viewing individual buckets.

tanmoyAtb:
closes #1210


asnaith (approved):
Looks good on render :)

FSM1 (approved):

PR #1407, opened 2021-08-04
[FILES] Ensure content type is retained when sharing

### LINKED ISSUE #1399: Shared jpeg image can't be open "file format not supported" renaming to jpg works
**Describe the bug**
When sharing a `.jpEg` (with an e) I can open the file and see it in my home folder, but in the shared folder it throws that the file format isn't supported. 

**To Reproduce**
1. Upload a .JPEG to your home
2. Share using copy to shared folder
3. Visit the shared folder
4. try to open the .jpeg and see it fail
5. rename to .jpg now it works

**Expected behavior**
Open the image.

**Console error**
none

**Others** that's the one
c

FSM1:
Closes #1399


Tbaut (approved):
awesome, working perfectly!

tanmoyAtb (approved):
💯

PR #1411, opened 2021-08-04
Using default cursors

### LINKED ISSUE #1408: Change cursors for navigation
The navigation across folders uses a pointer cursor now,
This should be updated to feel more like GDrive, Dropbox and OS

tanmoyAtb:
closes #1408


asnaith:
Hey Tanmoy, this looks good in macOS Chrome & Firefox, can see that the cursor is an arrow now when hovering over a folder. It only changes to a pointer when hovering over a checkbox or kebab menu icon which makes sense 🙂

This should reduce some confusion but still might not appease the users who don't like the double click. 

A suggestion that might be out of the context of this PR - perhaps we could consider adding a pointer when hovering over the folder name and making the folder name itself a single click to enter the folder? We can keep the double click if a user is hovering on the row. This is how folder navigation works in OneDrive but the current way is more similar to Google Drive. (cc @sweetpea22)

Tbaut:
@asnaith and @sweetpea22 I agree, let's continue the conversation in #1415

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemTableItem.tsx
```diff
@@ -27,6 +27,7 @@ const useStyles = makeStyles(({ breakpoints, constants, palette }: CSFTheme) =>
   return createStyles({
     tableRow: {
       border: "2px solid transparent",
+      cursor: "default",
```
Tbaut: Rather than overriding the cursor here, I'd make sure the original behavior is not to have a `cursor: pointer`, from what I could see, it's in `/packages/common-components/src/Table/TableRow.tsx`

Tbaut (changes requested):
Working well but I think the implementation can be changed to solve the ground problem rather than overriding it.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemTableItem.tsx
tanmoyAtb: Yeah I thought about this too! will update!

tanmoyAtb: Looks like we already had a prop to handle the cursor! 🗡️

tanmoyAtb:
> Hey Tanmoy, this looks good in macOS Chrome & Firefox, can see that the cursor is an arrow now when hovering over a folder. It only changes to a pointer when hovering over a checkbox or kebab menu icon which makes sense 🙂
> 
> This should reduce some confusion but still might not appease the users who don't like the double click.
> 
> A suggestion that might be out of the context of this PR - perhaps we could consider adding a pointer when hovering over the folder name and making the folder name itself a single click to enter the folder? We can keep the double click if a user is hovering on the row. This is how folder navigation works in OneDrive but the current way is more similar to Google Drive. (cc @sweetpea22)

Yes makes sense! The UX you suggested is also used by dropbox. Will continue in #1415

Tbaut (approved):
:100:

asnaith (approved):

packages/storage-ui/src/Components/Modules/FileSystemItem/FileSystemTableItem.tsx
```diff
@@ -171,7 +171,6 @@ const FileSystemTableItem = React.forwardRef(
           droppable: isFolder && (isOverMove || isOverUpload)
         })}
         type="grid"
-        rowSelectable={true}
```
RyRy79261, line 174: Dont we still want the rows selectable just with different cursors?

Tbaut: The `selectable` prop from what I could see only changed the cursor and allows an "onClick" prop that we don't use anyway. So having it to false doesn't seem to change the behavior and actually just acts on the cursor for us, which is nice.

PR #1421, opened 2021-08-05
Use cypress session instead of locally stored files.

Tbaut:
- Use the new experimental cy.sessions https://cypress.io/blog/2021/08/03/authenticate-faster-in-tests-cy-session-command/
- removed the local/session hack
- remove a reload hack


asnaith:
I took a look at this today and at first I didn't see any issues running the test suite locally or in ci until I started running them repeatedly on mass. I ran each test over a 100 times and that then exposed a few flaws in the test suite but nothing that seemed specific with the new session changes. 

Initially there were a lot of failures that seemed to happen after the `cy.session` instruction in commands.ts. The site sometimes wouldn't load (causing an appHeaderLabel not being found error) but then the errors stopped happening and I didn't see the problem again. The last command before the `cy.visit` was to enter the account password - I'm wondering if perhaps there was a temporary issue on the API that could have caused it? That's the only thing I can think of that would correlate with the circumstantial nature. Not too sure how else the issue could have went away without any env / app or test changes 🤔. 

The only commits I added were unrelated to session stuff, just fixing a couple issues I saw when running the tests in a loop. 
Just extended some element timeouts and used `safeClick` on another dialog button that would sometimes not respond to the click (like we have seen elsewhere). 

Lastly, I also saw these two errors which I could only produce a few times and then never again, towards the end of the day I was running the test locally, hundreds of times with no issues. 

Screen Shot 2021-08-05 at 3 06 05 PM

Screen Shot 2021-08-05 at 12 45 57 PM

asnaith:
A lot of the failures in the latest CI run look genuine and related to network errors (resulting in unexpected elements being / not being on screen). 

File management -- desktop -- cannot create a folder with an invalid name (failed)

File management -- desktop -- can rename a file with error handling (failed)

File management -- desktop -- can delete and recover a folder (failed)

asnaith (approved):

FSM1 (approved):

PR #1426, opened 2021-08-06
Expose decryption key

### LINKED ISSUE #1392: Expose the decryption key in the file info (hidden by default)
The bucket decryption key is only printed in the console.
As a transparency feature, we should have it displayed in the file info, prob. hidden by default, and let users show it with a click on an eye icon, a little like it's done for the keys on Storage.

The decryption key will be the same for each file in a given bucket/shared folder but I think it's where it's needed rather than putting it in a setting tab with a table for all the shared folder :man_shrugging:

RyRy79261:
closes #1392


packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
```diff
@@ -0,0 +1,102 @@
...
+)
+
+interface IToggleHiddenText {
+  children: ReactNode
+  iconPosition?: "left" | "right"
+  hiddenLength?: number
+  className?: string
+}
+
+const loop = (count: number) => {
```
Tbaut: We can replace this by placing `new Array(hiddenLength).fill("●")` inside the `ToggleHiddenText`
Maybe inside a useMemo to prevent re-render.

packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
```diff
@@ -0,0 +1,102 @@
...
+  let text = ""
+  for (let i = 0; i < count; i++) {
+    text += "●"
+  }
+  return text
+}
+
+const ToggleHiddenText: React.FC<IToggleHiddenText> = ({
```
Tbaut: nit, redundant
```suggestion
const ToggleHiddenText = ({
```

packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
```diff
@@ -0,0 +1,102 @@
...
+  hiddenLength = 6,
+  children
+}: IToggleHiddenText) => {
+  const classes = useStyles()
+  const [hidden, setHidden] = useState(true)
+  return (
+    <span className={clsx(classes.root, className)} >
+      {
+        hidden
+          ? iconPosition === "left" ? (
```
Tbaut: This condition makes it very verbose and hard to read, what do you think about removing it?
```
        hidden
          ? (
            {iconPosition === "left" && <EyeComponent/>}
              <span>{ loop(hiddenLength) }</span>
            {iconPosition === "right" && <EyeComponent/>}
          )
```

packages/files-ui/src/Components/Modules/FileBrowsers/FileInfoModal.tsx
```diff
@@ -361,6 +363,32 @@ const FileInfoModal = ({ filePath, close }: IFileInfoModuleProps) => {
...
+                    className={classes.subSubtitle}
+                    variant="body2"
+                    component="p"
+                  >
+                    <ToggleHiddenText hiddenLength={14}>
+                      <span>
+                        {
+                          personalEncryptionKey
+                        }
+                      </span>
```
Tbaut: nit
```suggestion
                      <span>{ personalEncryptionKey }</span>
```

packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
```diff
@@ -0,0 +1,102 @@
...
+            </>
+          ) : iconPosition === "left" ? (
+            <>
+              <EyeOpenSvg
+                className={classes.icon}
+                onClick={() => setHidden(true)}
+              />
+              {
+                children
+              }
```
Tbaut:
```suggestion
              { children }
```

packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
```diff
@@ -0,0 +1,102 @@
...
+            </>
+          ) : (
+            <>
+              {
+                children
+              }
```
Tbaut:
```suggestion
              { children }
```

Tbaut (changes requested):
Nice, works well. I left some comments to get a better readability.
Also idk if you saw but @FSM1 already added an eye icon for the api keys for storage in the past, it's not used any more but the eye (at least the open one) is still there.

packages/files-ui/src/Components/Modules/FileBrowsers/FileInfoModal.tsx
```diff
@@ -361,6 +363,32 @@ const FileInfoModal = ({ filePath, close }: IFileInfoModuleProps) => {
...
+                <div className={classes.subInfoBox}>
+                  <Grid
+                    item
+                    flexDirection="row"
+                  >
+                    <Typography
+                      variant="body1"
+                      component="p"
+                    >
+                      <Trans>Decryption key:</Trans>
```
asnaith: nit (none of the other subtitles have a colon on the modal)

```suggestion
                      <Trans>Decryption key</Trans>
```

packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
RyRy79261: Ahh knew there was a smoother way will update

Regarding the eye icon, I used an existing one however it did have fixed colors so I removed the hardcoded stroke prop

packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
RyRy79261: Yeah was grating me as well, thanks!

packages/common-components/src/ToggleHiddenText/ToggleHiddenText.tsx
Tbaut: aaah you're right, that's great, I missed it

Tbaut (approved):
:+1:

FSM1 (approved):

PR #1430, opened 2021-08-10
Programmatic folder creation for tests

### LINKED ISSUE #1410: Create a command in test to programmatically create a folder
This should replace https://github.com/ChainSafe/ui-monorepo/blob/dev/packages/files-ui/cypress/support/page-objects/homePage.ts#L62-L67

Tbaut:
closes #1410


packages/files-ui/cypress/support/utils/apiTestHelper.ts
```diff
@@ -30,8 +31,39 @@ export const apiTestHelper = {
                 console.log(`Deleting bucket ${bucketType} ${JSON.stringify(toDelete)}`)
                 apiClient.removeBucketObject(buckets[0].id, { paths: toDelete }).catch()
               })
+              .catch(console.error)
           })
         })
     })
+  },
+  // create a folder with a fill path like "/new folder"
```
Tbaut:
```suggestion
  // create a folder with a full path like "/new folder"
```

asnaith:
This is going to be great, thank you for doing this 😃

This is very useful and efficient to use in the scenarios when we need a folder as a requisite for a test but are not testing the folder creation directly.

asnaith:
I was just experimenting with this new functionality and found a few issues - 

1) If you use `apiTestHelper.createFolder` more than once in a test then only the first folder is created 

2) Sometimes a folder is there when the next test starts, it seems like the clearing does not always work now

3) Sometimes a folder is not created at all

Perhaps for reliability the api test helper should confirm the intended creation / deletion was completed before exiting? 

I was using this custom test to check the new command: 

```
Cypress._.times(100, () => {
      it.only("Folder creation", () => {
        cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })

        apiTestHelper.createFolder("/test folder 1/subfolderA")
        homePage.fileItemRow().should("have.length", 1)
        apiTestHelper.createFolder("/test folder 2/subfolderB")
        homePage.fileItemRow().should("have.length", 2)
        apiTestHelper.createFolder("/test folder 3/subfolderC")
        homePage.fileItemRow().should("have.length", 3)
      })
    })
```

Tbaut:
@asnaith feel free to check my 5841a94 commit. I've tried to do the check after the folder elegantly, but it basically fails because of other things. I really think we need to get this flow together, I've added logs and when running the test we can see that they are fired in an order that doesn't work at all.

RyRy79261 (approved):

packages/files-ui/cypress/support/utils/apiTestHelper.ts
```diff
@@ -60,10 +58,20 @@ export const apiTestHelper = {
...
             console.error(e)
           }
         })
     })
+      .then(() => {
+        navigationMenu.binNavButton().click()
+        navigationMenu.homeNavButton().click()
+
+        const firstFolderName = folderPath.split("/")[1]
+        cy.get("[data-testid=table-home]", { timeout: 10000 })
```
asnaith: not really important but could also have used a pre-existing element here `homePage.fileItemName().contains(firstFolderName)`.  

(...and fileItemName works for files and folders so could probably be renamed to itemName!)

asnaith:
@Tbaut It's looking good in code but I can't get it to run successfully

Seeing this error which seems to come after trying to clear the bucket 

Screen Shot 2021-08-11 at 4 26 17 PM

also this when trying to assert on the folder name

Screen Shot 2021-08-11 at 4 36 51 PM

I can look into it more tomorrow

packages/files-ui/cypress/support/utils/apiTestHelper.ts
Tbaut: The `homePage.fileItemName()` would be an array then, would a `.contains` then work?

asnaith: Yep it can match an item in an array

Tbaut: nice, working well indeed, thanks.

Tbaut:
I managed to beat the monster. The trick is: use `new Cypress.Promise`. This makes sure that Cypress will wait for us to call the associated `resolve` before continuing. This way, we can actually do our calls the way we want. The logs that I left in the helper function show that everything is happening in the correct order, which wasn't the case before and was the origin of the race conditions we had.

asnaith (approved):

asnaith:
This is fantastic, tried it out and it's working great! 🥳

The whole test suite is going to be more efficient and reliable with this in place. Thank you 🙏

PR #1432, opened 2021-08-10
[Files] Use material menus

### LINKED ISSUE #1285: Dropdown always opens to the bottom even if I don't have the space
A video is worth 1000 words. The best would be the dropdown to open upward.

https://user-images.githubusercontent.com/33178835/125859281-d2196c71-cf85-423b-8145-a0bee7dc2f1b.mp4

Tbaut:
closes https://github.com/ChainSafe/ui-monorepo/issues/1285

As discussed in the UI meeting, here's a trial at integrating bit by bit some elements of material UI so that we don't have to re-invent the wheel for all we build. I didn't change all the dropdowns just yet because I wanted to submit this PR early on to get feedback.

Todos:
- ~dark mode~
- change other dropdown menus
  - ~shared folder~
  - ~grid items~
  - ~Mobile menu to create folders etc~ 
- ~I may have broken some tests~

https://user-images.githubusercontent.com/33178835/128902427-1edc01e4-9213-4300-be79-4a10c1bf72fc.mp4


Tbaut:
The `onBlur` on the rename input created problems again, so I decided to replace it by a transparent div component behind the edition field with a click handler to close it.

Tests are now passing and dark mode is handled.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemGridItem.tsx
```diff
@@ -114,6 +114,13 @@ const useStyles = makeStyles(({ breakpoints, constants, palette }: CSFTheme) =>
       [breakpoints.down("md")]: {
         padding: 0
       }
+    },
+    clickOutSideArea : {
```
tanmoyAtb: I see what its doing, 
but could we not use the `useOnClickOutside` ref system to catch the click outside ?

tanmoyAtb (approved):
Not sure if we had any discussion about adding material into the mix,
but moving forward we should have a discussion around how we work with the component library!

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemGridItem.tsx
tanmoyAtb: we have a `useOnClickOutside` hook in the `common/theme`

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemGridItem.tsx
```diff
@@ -114,6 +114,13 @@ const useStyles = makeStyles(({ breakpoints, constants, palette }: CSFTheme) =>
       [breakpoints.down("md")]: {
```
tanmoyAtb, line 105: The dropdown on the Grid items has not been worked on!

tanmoyAtb (changes requested):

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemGridItem.tsx
Tbaut: Thanks Tanmoy, I wanted to have feedback as early as possible and finish off the other (that's int he todo). I'll make sure to update this one.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemGridItem.tsx
Tbaut: aah thanks, I didn't know about this one, will take a look.

Tbaut:
> Not sure if we had any discussion about adding material into the mix,
but moving forward we should have a discussion around how we work with the component library!

We had a discussion indeed. At the end of the day, we think we shouldn't re-invent the wheel and spend a ton of time building our own component library when there are so many good ones out there. We've already spent a lot of time doing these, so there's a bias in changing our path now, but the maintenance also takes some time. The good part is: we can do it step by step. This dropdown issue was a good one to start with and bit by bit change most of other components as we need them by slightly customizing the material ones to our look and feel.

I added a new UI-component folder in Files, maybe I should do this in the common component package, in a separate folder :man_shrugging:

tanmoyAtb (approved):
Looks good to go :)

Tbaut:
I still have the mobile menu to do, and I'll create a new issue for the proper dropdowns (or maybe not for now :shrug` ?) that are used in the header with the username and the settings>languages

FSM1 (approved):
working really nicely

Issue #1283, opened 2021-07-15
ChainSafe Storage: Enable UI Tests


asnaith:
**Task:**

The new CIDs and Bucket Management tests are currently being skipped due to the product still awaiting other issues to be resolved. 

- Remove `Its.skip` from the new tests and ensure the new tests pass reliably
- Address any issues with `clearPins` / apiTestHelper.ts  to ensure data states can be cleaned prior to test. 

**Further Details:**

"_The new `clearPins` command is still not working but I think it's because we have other unresolved promise issues in the actual app so I was planning on looking into it more once they are resolved._"

_Originally posted by @asnaith in https://github.com/ChainSafe/ui-monorepo/pull/1236#discussion_r669161541_

PR #1433, opened 2021-08-10
Updates to storage ui test suite

asnaith:
Work towards #1283 (still a wip but submitting these changes now to keep the PRs small). 

- Update the storage ui test suite to use cypress session management like what was previously implemented for files
- Updated some of the tests to match current behaviour (they haven't been running so some things eg default navigation changed)
- Used safeClick on modal dialog confirmation buttons for increased reliability 

Notes: 

Still not ready to enable these tests because of issues with the apiTestHelper (see #1434). It is either not clearing pins (or it is but they aren't removed straight away because of the queuing system, I'm really not sure what's happening with it at the moment). This going to make it hard to build reliable ui tests for CID Pinning at least in the current structure. For example at the moment 

- I can't assert a CID was successfully pinned because there might already be one pinned at the beginning of the test so counting on the rows could be unreliable

- I can't assert that the CID was deleted with the current test because the pin deletion is usually queued for a indeterminate period of time and not instantly removed from the UI (although sometimes it is so we can't just assert on "queued" status either!).


packages/storage-ui/cypress/support/commands.ts
```diff
@@ -105,71 +65,37 @@ Cypress.Commands.add(
...
-        cy.wait(1000)
-        authenticationPage.web3SignInButton().click()
-      }
+    cy.session("web3login", () => {
+      cy.visit(url)
+      authenticationPage.web3Button().click()
+      authenticationPage.showMoreButton().click()
+      authenticationPage.detectedWallet().click()
+      authenticationPage.web3SignInButton().safeClick()
+      bucketsPage.bucketsHeaderLabel().should("be.visible")
```
asnaith, line 78: Just an FYI - The last line here is required so that we don't exit the session block early before we have fully logged in, prevents us exiting before the session data is created and saved. 

We haven't been waiting for the header label on Files in the session block because we also have to enter a password and not save the browser so I think that gives us a little bit of extra time to save the token. To avoid any potential timing issues though I’ve added a similar check for Files too

Tbaut: Thanks, the save browser modal in Files is certainly the one that helped (just a FYI because before the password is entered we haven't reconstructed the key, and the there's prob. not much of interest in the local and session storages.)

Tbaut (approved):
Nice! I also removed rimraf now that we don't use it anywhere else :)

Tbaut:
Just a note, the 1st test is always failing
>  Bucket management
 |     desktop
 |       1) can create a bucket

packages/storage-ui/cypress/support/commands.ts
asnaith: @Tbaut Ah, good to know the detail. Thanks 👍

asnaith:
> Just a note, the 1st test is always failing
> 
> > Bucket management
> > |     desktop
> > |       1) can create a bucket

Oops, that was my mistake. I didn't re-add the `it.skip` after updating the tests. We can't enable it at the moment until the apiTestHelper is clearing the storage bucket state (separate ticket). I've re-added the `skip` again

FSM1 (approved):

PR #1424, opened 2021-08-06
Storage: Download functionality on bucket files

RyRy79261:
closes #1418

PR #1437, opened 2021-08-11
[Storage] Copy cid functionality

### LINKED ISSUE #1431: Allow for copying the CID
**Describe the bug**
CID text can currently not be selected

**To Reproduce**
1. Log in to Storage
2. Navigate inside a bucket with some files uploaded to it.
3. It is not possible to manually select the CID text

**Expected behavior**
A button for the user to easily copy the CID to the clipboard for further use.

### LINKED ISSUE #1440: CID column overflowing other columns
**Describe the bug**
Due to the name column & CID column having `fr` spacing and the CID being a long, unbroken string, the sizing for CID takes more space than the name field, resulting in a significant overflow

**To Reproduce**
1. Go to the website storage
2. Set screen size to a bit smaller than 1080 (Screenshots attached)
3. Open a bucket that contains a file

**Expected behavior**
The CID field should gracefully have an overflow styling, in conjunction with #1431, the column should shrink but still allow the user to get access to the full CID path in lieu of an info screen 

**Screenshots**

 - OS:  Manjaro
 - Browser FF
 - Version 90.0.2

image (1)
image (2)
image (3)

RyRy79261:
closes #1431
closes #1440 

The issue that was fixed in a previous PR that was reverted is still present: https://github.com/ChainSafe/ui-monorepo/pull/1424#issuecomment-896099344

image


Tbaut (changes requested):
Nice, just there's no indication that clicking on the CID will copy it. Can we have a button (with the copy icon) right from the CID as suggested in the issue?

tanmoyAtb:
I feel something like what we have in the settings will look better here !

Screenshot 2021-08-17 at 1 11 12 AM

RyRy79261:
@tanmoyAtb The ticket specified a button and we dont specifically know the space requirements so cant use the center ellipsis function, previous efforts to use overflow ellipsis weren't accepted so not sure about changing from the button at this point

Tbaut:
I think it's all a matter of how useful something is for users.
- A truncated cid that you can't copy is useless.
- A full cid that you can't select (hence can't copy) is almost useless.
^ we have had both in the past and were not happy about it.

Now in my opinion:
- A full cid that you can't select but you can copy certainly works
- A truncated cid that you can copy also works IMHO.

What Tanmoy suggests is what I had in mind too when reading the ticket. A column named "CID" with just a button looks super weird, we could have had menu entry if copying the CID without seeing it was a requirement.

RyRy79261:
Ok, so we have variants of all of these so far, what should I build?

Tbaut:
I think what Tanmoy suggested [here](https://github.com/ChainSafe/ui-monorepo/pull/1437#issuecomment-899753243) is what @FSM1 talked about. The cid and a small copy button. If there's not enough space for the full cid, we can have the ellipsis overflow without issue as well since users can still copy it.

RyRy79261:
@Tbaut @tanmoyAtb 

So I've set a 200px limit for this as I couldn't find a way around the `fr` unit behaviour in the grid without needing to calculate the percentage the table takes up of the screen for `vw` based sizing. One thing that has been on my mind is where the `name` column's text is, everything else is centered so this table looks a bit weird to me 

image

packages/storage-ui/src/Components/Modules/FileSystemItem/FileSystemTableItem.tsx
```diff
@@ -103,7 +104,112 @@ const useStyles = makeStyles(({ breakpoints, constants, palette }: CSSTheme) =>
...
+    //     height: 0,
+    //     borderLeft: "5px solid transparent",
+    //     borderRight: "5px solid transparent",
+    //     borderTop: `5px solid ${ palette.additional["gray"][9]}`
+    //   },
+    //   "&.active": {
+    //     opacity: 1,
+    //     visibility: "visible"
+    //   }
+    // }
```
Tbaut: stray comment
```suggestion
```

packages/storage-ui/src/Components/Modules/FileSystemItem/FileSystemTableItem.tsx
```diff
@@ -240,7 +359,23 @@ const FileSystemTableItem = React.forwardRef(
...
               <TableCell>
-                {!isFolder && cid}
+                {!isFolder && <>
+                  <div
+                    className={classes.copyArea}
+                    onClick={onCopyCID}>
+                    <div className={clsx(classes.copiedFlag, { "active": copied })}>
+                      <span>
+                        <Trans>
+                        Copied!
```
Tbaut:
```suggestion
                          Copied!
```

tanmoyAtb:
> @Tbaut @tanmoyAtb
> 
> So I've set a 200px limit for this as I couldn't find a way around the `fr` unit behaviour in the grid without needing to calculate the percentage the table takes up of the screen for `vw` based sizing. One thing that has been on my mind is where the `name` column's text is, everything else is centered so this table looks a bit weird to me
> 
> image

I would suggest: 
Make the header `Name` left aligned as the contents are left aligned,
and make the CID column content `center` aligned like the header.

Tbaut:
Even on my small screen the CID is being cut although there's still plenty of space for it to grow. Should we have a higher limit maybe?
image

FSM1:
> @Tbaut @tanmoyAtb
> 
> So I've set a 200px limit for this as I couldn't find a way around the `fr` unit behaviour in the grid without needing to calculate the percentage the table takes up of the screen for `vw` based sizing. One thing that has been on my mind is where the `name` column's text is, everything else is centered so this table looks a bit weird to me
> 
> image

Since the content of the column is limited to 200px, would it not make sense to define it in the grid spec as a 200px column?

`export const desktopGridSettings = "50px 69px 3fr 200px 100px 45px !important"`

Tbaut:
I've played a bit with `minmax` and came up with values to achieve that, I'll commit directly, feel free to revert, but having the full cid visible when there's space for it looks nicer IMHO.

https://user-images.githubusercontent.com/33178835/130599264-4f413b7c-1233-4c87-9c31-84f826b91b92.mp4

Tbaut (approved):

Tbaut (changes requested):
It looks like something went wrong with the merge, there are many changes related to billing in this PR now :see_no_evil:

Tbaut:
I checked quickly, the unrelated code comes from c42cec5, you'll probably want to revert that one.

packages/files-ui/src/locales/de/messages.po
```diff
@@ -209,7 +209,7 @@ msgid "Enter the verification code:"
 msgstr "Geben Sie den Verifizierungscode ein:"
 
 msgid "Essentials - Free"
-msgstr "Essentials - Kostenlos"
+msgstr ""
```
Tbaut:
```suggestion
msgstr "Essentials - Kostenlos"
```

packages/files-ui/src/locales/es/messages.po
```diff
@@ -210,7 +210,7 @@ msgid "Enter the verification code:"
 msgstr ""
 
 msgid "Essentials - Free"
-msgstr "Esenciales - Gratis"
+msgstr ""
```
Tbaut:
```suggestion
msgstr "Esenciales - Gratis"
```

packages/files-ui/src/locales/fr/messages.po
```diff
@@ -210,7 +210,7 @@ msgid "Enter the verification code:"
 msgstr "Entrez le code de vérification :"
 
 msgid "Essentials - Free"
-msgstr "Essentials - Gratuit"
+msgstr ""
```
Tbaut:
```suggestion
msgstr "Essentials - Gratuit"
```

Tbaut (approved):

Tbaut:
I've reverted the problematic commit, let's merge this PR :rocket:

FSM1 (approved):
aww yuss <3

PR #1438, opened 2021-08-11
Disabled select on row names & enhance simple click experience

### LINKED ISSUE #1415: Enhance navigation UX
- Simple click on a file/folder feels slow
Our implementation of the double-click makes it so that the simple click feels slow.
This is directly related to the delay for us to detect a double-click, 250ms. https://github.com/ChainSafe/ui-monorepo/blob/dev/packages/common-theme/src/Hooks/useDoubleClick.ts#L7 Compare this to Dropbox, the click is immediate

- double clicking on a folder selects the name, maybe we should make it un-selectable?
https://user-images.githubusercontent.com/33178835/128336625-1891efac-98c4-4040-a9d7-129bbf28e0eb.mp4

RyRy79261:
closes #1415

### DIFF
--- a/packages/common-theme/src/Hooks/useDoubleClick.ts
+++ b/packages/common-theme/src/Hooks/useDoubleClick.ts
@@ -15,8 +15,6 @@ export function useDoubleClick(
       case 2:
         actionDoubleClick && actionDoubleClick(e)
         break
-      // case 3:
-      //   break
       }
     }
   }, [actionSingleClick, actionDoubleClick])
--- a/packages/common-theme/src/Hooks/useDoubleClick.ts
+++ b/packages/common-theme/src/Hooks/useDoubleClick.ts
@@ -1,41 +1,25 @@
 import * as React from "react"
-import { useCallback, useEffect, useState } from "react"
+import { useCallback } from "react"
 
 export function useDoubleClick(
   actionSingleClick: (e?: React.MouseEvent) => void,
-  actionDoubleClick: (e?: React.MouseEvent) => void,
-  delay = 250
+  actionDoubleClick: (e?: React.MouseEvent) => void
 ) {
-  const [clickCount, setClickCount] = useState(0)
-  const [event, setEvent] = useState<React.MouseEvent | undefined>()
-
-  useEffect(() => {
-    const timer = setTimeout(() => {
-      // simple click
-      if (clickCount === 1) {
-        actionSingleClick && actionSingleClick(event)
-      }
-
-      setClickCount(0)
-    }, delay)
-
-    // the duration between this click and the previous one
-    // is less than the value of delay = double-click
-    if (clickCount === 2) {
-      setClickCount(0)
-      actionDoubleClick && actionDoubleClick(event)
-    }
-
-    return () => clearTimeout(timer)
-
-    // actionDoubleClick is very unstable and has many dependencies
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [actionSingleClick, clickCount, delay])
 
   const onClick = useCallback((e?: React.MouseEvent) => {
-    setEvent(e)
-    setClickCount((prev) => prev + 1)
-  }, [])
+    if (e) {
+      switch (e.detail) {
+      case 1:
+        actionSingleClick && actionSingleClick(e)
+        break
+      case 2:
+        actionDoubleClick && actionDoubleClick(e)
+        break
+      // case 3:
+      //   break
+      }
+    }
+  }, [actionSingleClick, actionDoubleClick])
 
   return { click: onClick }
 }
--- a/packages/storage-ui/src/Components/Modules/FileSystemItem/FileSystemItem.tsx
+++ b/packages/storage-ui/src/Components/Modules/FileSystemItem/FileSystemItem.tsx
@@ -388,7 +388,7 @@ const FileSystemItem = ({
     [desktop, viewFolder, name, cid, onFilePreview, isFolder]
   )
 
-  const { click } = useDoubleClick(onSingleClick, onDoubleClick, 150)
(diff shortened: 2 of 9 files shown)


Tbaut (commented):
The problem with the selection is fixed perfectly, however with 150ms, this makes the double click way too quick, I barely manage to have a double click now, and the single click is still too slow.

I thought we could change our approach completely and use something like https://stackoverflow.com/a/66515641

We don't really care if the single click event is fired because it's only a selection, so if there's a double click, then it'll select and navigate/show preview.

What do you think?

RyRy79261:
Ahh ok, 150 was just the number that fit the people I tried it out on, nice find on that example there, I'll test it out n refactor if feasible

RyRy79261:
@Tbaut So the one issue I think we might run into is using the switch case, that it would trigger single click & double click actions

One option if this becomes an issue is to trigger the execution, alternatively being able to cancel the single click action if double is triggered

image

Tbaut:
> switch case, that it would trigger single click & double click actions

Yup, this is the behavior I meant to address with my
> We don't really care if the single click event is fired because it's only a selection, so if there's a double click, then it'll select and navigate/show preview.

So I don't think it's actually a problem, but we should check how it behaves in real life to be certain it's really not an issue..

RyRy79261:
Ahh ok rad, yeah just thinking of use beyond our row selection but granted use cases where theres a single and a double tied to the same item are usually select or execute behaviours

packages/common-theme/src/Hooks/useDoubleClick.ts
```diff
@@ -1,41 +1,25 @@
...
+    if (e) {
+      switch (e.detail) {
+      case 1:
+        actionSingleClick && actionSingleClick(e)
+        break
+      case 2:
+        actionDoubleClick && actionDoubleClick(e)
+        break
+      // case 3:
+      //   break
```
Tbaut:
```suggestion
```

packages/storage-ui/src/Components/Modules/FileSystemItem/FileSystemItem.tsx
```diff
@@ -388,7 +388,7 @@ const FileSystemItem = ({
     [desktop, viewFolder, name, cid, onFilePreview, isFolder]
   )
 
-  const { click } = useDoubleClick(onSingleClick, onDoubleClick)
+  const { click } = useDoubleClick(onSingleClick, onDoubleClick, 150)
```
Tbaut:
```suggestion
  const { click } = useDoubleClick(onSingleClick, onDoubleClick)
```

Tbaut (commented):
I think it's a great UX improvement for the simple click, and the double click just works fine from my tests :tada:

Tbaut (approved):

RyRy79261:
Rad! Thanks for the nifty approach, didn't realise events have context of other events

asnaith:
This is a nice improvement Ryan, the single click feels way better now 😃

asnaith (approved):

Tbaut:
Can you please check the tests @RyRy79261 one seems to fail for valid reasons :man_shrugging: 

 >|   1) File management
 |        desktop
 |          can delete and recover a folder:
 | 
 |       AssertionError: Timed out retrying after 20000ms: Too many elements found. Found '2', expected '1'.
 |       + expected - actual
 | 
 |       -2
 |       +1

packages/common-theme/src/Hooks/useDoubleClick.ts
```diff
@@ -1,41 +1,23 @@
...
-  const [clickCount, setClickCount] = useState(0)
-  const [event, setEvent] = useState<React.MouseEvent | undefined>()
 
-  useEffect(() => {
-    const timer = setTimeout(() => {
-      // simple click
-      if (clickCount === 1) {
-        actionSingleClick && actionSingleClick(event)
+  const onClick = useCallback((e?: React.MouseEvent) => {
+    if (e) {
```
tanmoyAtb, line 10: OMG! this is a gem! 💎 
The events had the info all along!
How come the internet's full of these hook implementations!!!!

tanmoyAtb (approved):

asnaith:
> Can you please check the tests @RyRy79261 one seems to fail for valid reasons 🤷‍♂️
> 
> > |   1) File management
> > |        desktop
> > |          can delete and recover a folder:
> > |
> > |       AssertionError: Timed out retrying after 20000ms: Too many elements found. Found '2', expected '1'.
> > |       + expected - actual
> > |
> > |       -2
> > |       +1

@Tbaut @RyRy79261 I took a quick look at this. It's a genuine failure but looks unrelated to Ryan's change. Files that shouldn't be there are present so when asserting that only the recovered folder exists it will fail because it sees other files. 

The additional files that can be seen on the screenshot were not added during the test and the test begins with `cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })` which should clear any pre-existing data. I can only assume the API call, to delete pre-existing files prior to the test beginning, did not work. 

File management -- desktop -- can addremove multiple files and upload (failed)

packages/common-theme/src/Hooks/useDoubleClick.ts
RyRy79261: @Tbaut Found this magic trick 🔥 

I can't believe I only found this out now 😂

Tbaut:
Thanks @asnaith, then let's merge and create an issue for that.

edit: done https://github.com/ChainSafe/ui-monorepo/issues/1448

packages/common-theme/src/Hooks/useDoubleClick.ts
Tbaut: Lol it's new for all of us, I didn't know either! Just I saw that Dropbox def. didn't have any timeout because their experience was great, and found that thing on wonderful SO :D

PR #1439, opened 2021-08-11
Define modals in the test suite as separate objects

### LINKED ISSUE #1436: Refactor modals being used in the test suite
At the moment we have been declaring modal elements in the page object where they are invoked from, this is inaccurate as they are not part of the page (just invoked from there) and they should be defined separately as their own object. 

Motivation:
- Test layer will be more readable
- When test errors occur the messaging will be more accurate 
- We can be less wordy on our element names because we won't be declaring multiple similar things, e.g cancel buttons, in the page object as they will be specific to the modal object
- In the future it might be possible to invoke a modal from more than one page so this prevents the need later down the line for copy and paste declaration in multiple page objects

asnaith:
closes #1436

### DIFF
--- a/packages/files-ui/cypress/support/page-objects/homePage.ts
+++ b/packages/files-ui/cypress/support/page-objects/homePage.ts
@@ -1,7 +1,7 @@
 import { basePage } from "./basePage"
 import { folderName } from "../../fixtures/filesTestData"
-
-export const click = ($el: JQuery<HTMLElement>) => $el.trigger("click")
+import { createFolderModal } from "./modals/createFolderModal"
+import { fileUploadModal } from "./modals/fileUploadModal"
 
 export const homePage = {
   ...basePage,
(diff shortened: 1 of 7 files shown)


packages/files-ui/cypress/support/page-objects/homePage.ts
```diff
@@ -1,20 +1,16 @@
 import { basePage } from "./basePage"
 import { folderName } from "../../fixtures/filesTestData"
-
-export const click = ($el: JQuery<HTMLElement>) => $el.trigger("click")
```
asnaith, line 4: Just cleaning up. This was not being used anymore since `safeClick` was added (it's defined in `commands.ts`).

Tbaut (approved):
Code looks good thanks!

RyRy79261 (approved):

PR #1444, opened 2021-08-16
Skip the entire lingui job if the author is weblate

Tbaut:
This should prevent the checkout from failing.


asnaith (approved):
Oh interesting. So was the lingui extraction putting too much strain / timing out whilst running within the docker container?

tanmoyAtb (approved):

RyRy79261 (approved):

Tbaut:
@asnaith no, it's just that the branch checkout fails when lingui ci runs because for weblate, the branch is external, and you can't actually do checkout the way the ci tries..

PR #1450, opened 2021-08-17
Refresh by navigating to the bin to fix tests

### LINKED ISSUE #1448: Fix File Management test "can delete and recover a folder" test
followup to https://github.com/ChainSafe/ui-monorepo/pull/1438#issuecomment-899827218

Quoting Andrew:
Files that shouldn't be there are present so when asserting that only the recovered folder exists it will fail because it sees other files.

The additional files that can be seen on the screenshot were not added during the test and the test begins with cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true }) which should clear any pre-existing data. I can only assume the API call, to delete pre-existing files prior to the test beginning, did not work.

| desktop
| can delete and recover a folder:
|
| AssertionError: Timed out retrying after 20000ms: Too many elements found. Found '2', expected '1'.
| + expected - actual
|
| -2
| +1
image

Tbaut:
closes https://github.com/ChainSafe/ui-monorepo/issues/1448


asnaith (approved):

asnaith:
Thank you Thibaut 🙏

FSM1 (approved):

RyRy79261 (approved):

PR #1459, opened 2021-08-19
[Files] Download multiple files in a zip

### LINKED ISSUE #1363: Download all files at once
This is a common use case for anything that isn't accessible/editable in a browser, e.g sharing a bunch of documents with a colleague, and they need to change then. Another one could be sharing pictures with friends after a trip, everyone has write access and adds their pictures, but you want to download them all at some point.

- You don't want to prompt the user to download every single file
- You wand to download every file in the background, show a download progress, then decrypt and put them in a zip that you allow users  to download (https://mega.io/ does this afaik)

This options should be available in the bulk operations controls if multiple items are selected.
We would need to handle folder downloads, but recursively checking for each file.

Tbaut:
closes #1363


asnaith:
Hey Thibaut, this is working really nicely. 

A couple of small observations, some may not be issues but I wanted to mention them and let you decide. 

a) It looks like the "download complete" modal is using the wrong colour icon (but I'm guessing this is the same one we always use and probably was pre-existing before these change). 

Screen Shot 2021-08-20 at 9 46 10 AM

b) I noticed that the option doesn't appear to download a folder as a zip (even though it contains files) if that's the only item selected on the root

Screen Shot 2021-08-20 at 9 49 26 AM

c) If you attempt to start a file upload at the same time the two toasts will overlap each other and do not stack 

Screen Shot 2021-08-20 at 9 41 27 AM

https://user-images.githubusercontent.com/3982665/130267655-49e852f6-9e13-47a5-9bc5-393bb0eb0d8f.mov

d) The option to download as a zip is missing from the kebab menu (makes sense to have it for folders containing items?). 

e) It's not possible to use this feature on a mobile device (because the checkboxes don't appear in the mobile view)

f) It's not possible to use this feature when the file browser is in grid mode (also because there are checkboxes in that view)

g) Once the selection has been made and the download has began would it make sense to disable the button until a new selection has been made? I'm guessing it would be pretty silly for someone to spam click the download button but it is possible. I was just not sure if this could be used maliciously to generate a lot of load and cause us issues? For example a 10gb collection of files spammed clicked hundreds of times. 

Screen Shot 2021-08-20 at 9 42 35 AM

Tbaut:
aah lot's of good points. Thanks for stress testing this. This feature is actually reusing the download toast, some of what you mention would also happen without this feature and require a better design, namely
- e) f) can't move/delete (and not download) multiple files in the grid mode or on mobile because there's no checkbox, this should prob be solved with a ctrl+click selection/long click on mobile. That's def for another issue.
- c) overlap with upload and download toast, this happens with a normal download as well I imagine. I'll check it out.

Preventing the click when there's a download makes sense, although I'm not sure, because we specifically don't want to block the UI, if someone wants to be malicious, they can fire programmatically 1000 calls still.

edit: what I'll do is remove the selection then, I guess that should be fine.

What I'll do
> a) It looks like the "download complete" modal is using the wrong colour

That's no me, but small enough indeed

> b) I noticed that the option doesn't appear to download a folder as a zip

Didn't think about it either, I will allow it.

> d) The option to download as a zip is missing from the kebab menu

Same, didn't think about it, totally makes sense.

asnaith:
> What I'll do
> 
> > a) It looks like the "download complete" modal is using the wrong colour
> 
> > b) I noticed that the option doesn't appear to download a folder as a zip
> 
> > d) The option to download as a zip is missing from the kebab menu
> 

Great, sounds good 👍. I've created separate issues for the other things discovered in testing but not introduced by this PR https://github.com/ChainSafe/ui-monorepo/issues/1465 and https://github.com/ChainSafe/ui-monorepo/issues/1466

De-selecting the items after initiating the download sounds like a good solution too.

Tbaut:
Added the following:
- Download complete color (for this one, but also upload and transfer)
- menu entry for folders to download as zip
- button available when selecting just one folder
- remove the selection after the user clicks on download as zip

Tbaut:
Added the following:
- Use relative link to where the user downloads from. E.g don't redo the whole structure in the archive, if you dl some files from `/fold1`, then `fold1` will not appear in the archive 
- remove the first `/` so that windows can see the files and folders. Thanks @FSM1 :)

asnaith:
The new additional changes are looking almost perfect.

> menu entry for folders to download as zip

👆 just a small issue with this - the menu option is there and works as intended if the folder contains files. However, if the folder is empty the option is still there but nothing happens then the option is clicked. Perhaps it'd be best to not show the option if the folder is empty or send feedback to the user if they are attempting to download a blank folder, what do you think?

Everything else is looking great

RyRy79261 (approved):
Well that was a beautiful experience 🔥

FSM1 (approved):

Tbaut:
> Perhaps it'd be best to not show the option if the folder is empty or send feedback to the user if they are attempting to download a blank folder, what do you think?

Good point, having nothing happening is def wrong. The problem with what you propose is that the users will first think it's a bug, because "why isn't the menu entry present for this folder, but it is for this one"? So I'll add an error toast, I think it's more explicit (it's also makes the whole logic a bit more convoluted but that's a needed evil I guess).

asnaith:
> I'll add an error toast, I think it's more explicit (it's also makes the whole logic a bit more convoluted but that's a needed evil I guess)

👍 Nice solution, works well

PR #1462, opened 2021-08-19
Implement cypress retry mechanism

### LINKED ISSUE #1460: Implement cypress retry mechanism
**Task:** 
Implement cypress's retry mechanism to automatically re-run a ui test after it fails. This will ensure less noise is caused because of a false failure (due to ci / network interruption etc) as we will only see failures that have occurred more than once. 

**Background:**
We had already discussed doing this a while back and there were legitimate concerns that retrying a UI test after a failure can hide possible flakey behaviour in the app so we decided not to proceed. 

The reality at the moment though is that the flakiness seems to be residing in the CI environment itself, the way we clear data (api test helper) and the dependence on a live API which may also have it's own issues / performance degradation rather than the app itself. 

My proposal is to implement this at least as an **_interim_** measure whilst we address other issues around the 3 aforementioned issues (CI Env, Api Test Helper, API). Bring back some usefulness to the tests, reduce the false failures and then revisit and remove at a later point if deemed necessary. 

**Further reading:** https://docs.cypress.io/guides/guides/test-retries#Introduction

asnaith:
closes #1460

### DIFF
--- a/packages/files-ui/cypress.json
+++ b/packages/files-ui/cypress.json
@@ -1,5 +1,10 @@
 {
     "integrationFolder": "cypress/tests",
     "video": false,
-    "experimentalSessionSupport": true
+    "experimentalSessionSupport": true,
+
+    "retries": {
+        "runMode": 2,
+        "openMode": 0
+    }
 }
--- a/packages/files-ui/cypress.json
+++ b/packages/files-ui/cypress.json
@@ -2,7 +2,6 @@
     "integrationFolder": "cypress/tests",
     "video": false,
     "experimentalSessionSupport": true,
-
     "retries": {
         "runMode": 2,
         "openMode": 0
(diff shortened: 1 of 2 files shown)


packages/files-ui/cypress.json
```diff
@@ -1,5 +1,10 @@
 {
     "integrationFolder": "cypress/tests",
     "video": false,
-    "experimentalSessionSupport": true
+    "experimentalSessionSupport": true,
+
+    "retries": {
+        "runMode": 2,
+        "openMode": 0
```
asnaith, line 8: Run mode is what we use in CI, open mode when we test locally. I've only added the retry for CI execution.

packages/files-ui/cypress.json
```diff
@@ -1,5 +1,10 @@
...
+    }
```
asnaith, line 8: Run mode is what we use in CI, open mode when we test locally. I've only added the retry for CI execution.

packages/files-ui/cypress.json
```diff
@@ -1,5 +1,10 @@
 {
     "integrationFolder": "cypress/tests",
     "video": false,
-    "experimentalSessionSupport": true
+    "experimentalSessionSupport": true,
+
```
Tbaut:
```suggestion
```

Tbaut (approved):

RyRy79261 (approved):

PR #1463, opened 2021-08-20
Current card: billing

### LINKED ISSUE #1369: Current card widget
This widget should either display `No card added` and a button to add a new card.

Alternatively if the user has added a card, display the current card details, and the ability to delete/update the card.

Note: This is only valid until we launch the Storage billing, where we would need to warn the user that updating a card would update all the products to use the new card.

image

tanmoyAtb:
closes #1369 

moved more of the code to the billing context !
Use this as Stripe pk
```
pk_test_1cf3P94zFb6QS9ZMWjKNwwAj00s9B2NgVN
```

Use card details from here for testing: https://gist.github.com/rymawby/9b904feec040b25a8034


packages/files-ui/src/Components/Modules/Settings/SubscriptionPlan/AddCardModal.tsx
```diff
@@ -104,22 +101,10 @@ const CreateFolderModal = ({ isModalOpen, onClose }: ICreateFolderModalProps) =>
...
-        "card[number]": cardInputs.cardNumber,
-        "card[exp_month]": cardInputs.cardExpiry.split("/")[0].trim(),
-        "card[exp_year]": cardInputs.cardExpiry.split("/")[1].trim(),
-        "card[cvc]": cardInputs.cardCvc
-      })
-    }).then((resp: AxiosResponse<{ card: {id: string}}>) => {
+    getCardTokenFromStripe(cardInputs).then((resp) => {
       // send stripe token to API
-      filesApiClient.addCard({ token: resp.data.card.id })
+      console.log(resp)
```
tanmoyAtb:
```suggestion
```

packages/files-ui/src/Contexts/BillingContext.tsx
```diff
@@ -20,51 +28,53 @@ const BillingContext = React.createContext<IBillingContext | undefined>(
...
+  const refreshDefaultCard = useCallback(() => {
+    filesApiClient.getDefaultCard().then((card) => {
+      setDefaultCard(card)
+    }).catch(console.error)
+  }, [filesApiClient])
 
-    // eslint-disable-next-line max-len
-    const dataString = `card[number]=${data.cardNumber}&card[exp_month]=${cardExpiryMonth}&card[exp_year]=${cardExpiryYear}&card[cvc]=${data.cardCvc}`
+  useEffect(() => {
+    if (isLoggedIn) {
```
Tbaut, line 45: Just out of curiosity, did you run into a behavior that required such logic?

Tbaut (commented):
Looks spotless from the code :raised_hands: 

Just from the UX perspective
- Having a "Add card" button if there's no card already would make sense. 
- The text "Credit card on file", I know it's from the design but I just don't understand it :shrug:. Does "on file" mean something I ignore?
image

packages/files-ui/src/Components/Modules/Settings/SubscriptionPlan/index.tsx
```diff
@@ -23,9 +24,10 @@ const PlanView: React.FC = () => {
   return (
     <div className={classes.root}>
       <CurrentProduct />
+      <CurrentCard />
       <div className={classes.container}>
         <Button onClick={() => setIsAddCardModalOpen(true)}>
-          <Trans>Add Card</Trans>
+          <Trans>Update Card</Trans>
```
Tbaut: This should check if there's a card or not.

packages/files-ui/src/Components/Modules/Settings/SubscriptionPlan/CurrentCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <Typography>
+           •••• •••• •••• {defaultCard.last_four_digit}
+          </Typography>
+        </div>
+        : <Typography
+          component="p"
+          variant="body1"
+          className={classes.noCard}
+        >
+          <Trans>No Card added</Trans>
```
Tbaut:
```suggestion
          <Trans>No Card</Trans>
```

packages/common-components/src/Icons/svgs/credit-card.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M21.75 3.75H2.25C1.83516 3.75 1.5 4.08516 1.5 4.5V8.25H22.5V4.5C22.5 4.08516 22.1648 3.75 21.75 3.75ZM1.5 19.5C1.5 19.9148 1.83516 20.25 2.25 20.25H21.75C22.1648 20.25 22.5 19.9148 22.5 19.5V10.3125H1.5V19.5ZM15.0703 15.1875C15.0703 15.0844 15.1547 15 15.2578 15H19.125C19.2281 15 19.3125 15.0844 19.3125 15.1875V16.875C19.3125 16.9781 19.2281 17.0625 19.125 17.0625H15.2578C15.1547 17.0625 15.0703 16.9781 15.0703 16.875V15.1875Z" fill="black"/>
```
RyRy79261: Theres hardcoded attributes here (height, width, fill, etc) best remove and check if dark mode's appearance works

RyRy79261 (approved):
Looks good, just the note about the svg, might need some styling fixes if enacted

packages/files-ui/src/Contexts/BillingContext.tsx
tanmoyAtb: Yes ! This felt weird to me too!
So thee billingContext would try to call the API before the tokens got initialized, receive a 401,
for some reason, when `fileAPIClient` updates, it sometimes makes the call properly, sometimes glitches !!!

packages/files-ui/src/Components/Modules/Settings/SubscriptionPlan/index.tsx
tanmoyAtb: Thats right! updated!

tanmoyAtb:
> Looks spotless from the code 🙌
> 
> Just from the UX perspective
> 
> * Having a "Add card" button if there's no card already would make sense.
> * The text "Credit card on file", I know it's from the design but I just don't understand it 🤷. Does "on file" mean something I ignore?
>   image

I guess that just gives the feeling thats something we have saved if we kept books of some sort!
I found it relatable. We could just have "Default Credit Card" !?

packages/common-components/src/Icons/svgs/credit-card.svg
tanmoyAtb: Great catch Ryan, pushed updates on the svg!

asnaith:
I've added the cypress retry mechanism from `dev` to this branch however the tests aren't even executing successfully at the moment in CI so that won't immediately help (affected by the same issue that's trying to be solved in https://github.com/ChainSafe/ui-monorepo/pull/1469). 

I've ran the test suite locally for this branch though and it was a 100% pass so I don't think there should be any concerns merging these changes, the ci test issue is not related to these changes.

packages/files-ui/src/Contexts/BillingContext.tsx
FSM1: This will work for now, but we should really consider only fetching this data when it is required by the component to be rendered

FSM1 (approved):
working well. left a comment about the get default card fetching, but this can be optimized later on.

PR #1464, opened 2021-08-20
[FILES] enforce maximum 32 char limit

### LINKED ISSUE #1443: Validate the amount of characters before submitting a username
The api limits the amount of char upon submission (currently 10 but should be updated), the api lookup looks good regardless of the amount of chat though, we should have a client side validation to prevent having an error after the submission.

FSM1:
Closes #1443

### DIFF
--- a/packages/files-ui/src/Components/Modules/Settings/Profile.tsx
+++ b/packages/files-ui/src/Components/Modules/Settings/Profile.tsx
@@ -286,7 +286,18 @@ const ProfileView = () => {
     const sanitizedValue = value?.toString() || ""
 
     setUsername(sanitizedValue)
-    !!sanitizedValue && debouncedOnLookupUsername(sanitizedValue)
+    if (sanitizedValue.length > 32) {
+      setUsernameData({
+        loading: false,
+        error: t`The username is too long`
+      })
+    } else {
+      setUsernameData({
+        loading: false,
+        error: ""
+      })
+      !!sanitizedValue && debouncedOnLookupUsername(sanitizedValue)
+    }
   }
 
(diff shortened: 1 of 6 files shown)


tanmoyAtb (approved):

Tbaut (approved):
looks good from the code

Render's not happy for some reason, I triggered a rebuild

FSM1:
> looks good from the code
> 
> Render's not happy for some reason, I triggered a rebuild

Looks like its passed now :)

Tbaut (commented):
Found an edge case while playing with it, I realized that when you hit 32 char, the error is displayed very briefly, but is then overridden by a valid state. How to reproduce, type 30 char, and then the last 2 quickly. If just just type the 32nd it's showing an error, but if you type quickly the 31st and 32nd, then it displays that it's fine.

edit: I think it doesn't matter much, maybe an issue with the debounce.

https://user-images.githubusercontent.com/33178835/130266732-088ce44c-3b84-4fdb-a0c4-b1a78fd64b4b.mp4

FSM1:
> Found an edge case while playing with it, I realized that when you hit 32 char, the error is displayed very briefly, but is then overridden by a valid state. How to reproduce, type 30 char, and then the last 2 quickly. If just just type the 32nd it's showing an error, but if you type quickly the 31st and 32nd, then it displays that it's fine.
> 
> edit: I think it doesn't matter much, maybe an issue with the debounce.
> 
>  debounce.mp4

ahh yeah that is definitely the debounce 🙈  Might actually try and spend a bit of time optimizing this code. Its a bit convoluted at the moment.

PR #1477, opened 2021-08-26
Sharing feature explainer

### LINKED ISSUE #1361: Add some slides to explain our sharing strategy
The way we do sharing is different than how users may be used to. For this reason, we should show a little explanation about how sharing with Files works:
- it should show on click on "shared" in the nav bar, or when sharing a file for the first time
- it should be skipable
- it should tell about creating a shared folder
- it should tell about how to create a nickname in the settings
- the more pictures/gifs the better

Tbaut:
closes https://github.com/ChainSafe/ui-monorepo/issues/1361


asnaith:
This is working really well. Found a tiny micro issue with the username text not being very legible in the last card in dark mode. 

Screen Shot 2021-08-26 at 4 46 10 PM

There's a compilation issue in the CustomModal.tsx for storage-ui that is causing the storage test failures.

asnaith:
...and I only just noticed this was a draft 🤦‍♂️

Tbaut:
Hehe yeah I'm not done yet, I checked with Cindy to make this looks good on dark mode. Also I'll need to make it appear when clicking on "share" .

Tbaut:
oops need to fix storage too.

edit: done

asnaith (approved):

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useSharingExplainerModalFlag.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useEffect, useState } from "react"
+
+export const DISMISSED_SHARING_EXPLAINER_KEY = "csf.dismissedSharingExplainer"
+
+export const useSharingExplainerModalFlag = () => {
+  const { localStorageGet, localStorageSet } = useLocalStorage()
+  const [hasSeenSharingExplainerModal, setHasSeenSharingExplainerModal] = useState(false)
+  const dismissedFlag = localStorageGet(DISMISSED_SHARING_EXPLAINER_KEY)
+
+  useEffect(() => {
```
tanmoyAtb, line 12: we can combine these `useEffects` into one !

packages/files-ui/src/Components/SharingExplainerModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { t, Trans } from "@lingui/macro"
+import React, { useCallback, useState } from "react"
+import { CSFTheme } from "../Themes/types"
+import CustomModal from "./Elements/CustomModal"
+import { SETTINGS_BASE } from "./FilesRoutes"
+import step1Image from "../Media/sharingExplainer/step1.png"
+import step2Image from "../Media/sharingExplainer/step2.png"
+import step3Image from "../Media/sharingExplainer/step3.png"
+
+export const DISMISSED_SHARING_EXPLAINER_KEY = "csf.dismissedSharingExplainer"
```
tanmoyAtb: This should be imported from the `useSharingExplainerModalFlag`

tanmoyAtb (dismissed):
This is dope!
The refactors are amazing too!

I left a couple of comments!
The modal on mobile screen needs to bee centered!

Screenshot 2021-09-01 at 6 14 55 PM

packages/files-ui/src/Components/Modules/FileBrowsers/hooks/useSharingExplainerModalFlag.tsx
Tbaut: absolutely, thanks.

packages/files-ui/src/Components/SharingExplainerModal.tsx
Tbaut: oh nice catch, refactor gone wrong.

Tbaut:
I had seen the modal being brought down on Mobile, but since I didn't introduce this, I closed my eyes and thoughts it's a feature or something. Checking it, I realized that it is indeed a feature, but only for some specific actions such as rename, moving etc. I adapted the modal I saw to not have this behavior (the Sharing modal for instance should not be brought down etc..)

FSM1 (changes requested):
Working great and looks fantastic.

Just need to remove the dependency on local storage for this, and leverage the API endpoints defined in https://github.com/ChainSafe/files-api-spec/pull/110/ to read and store this data.

I would also probably not use a separate hook for this, but rather keep this flag data on the user context that already exists.

Tbaut:
Given that the api PR is still open, https://github.com/ChainSafe/files-api/pull/1427
I thought we could go on with this and refactor later when the api is actually merged.

FSM1 (approved):

PR #1487, opened 2021-09-02
[Files] - New survey - Discord link and tiny fix

### LINKED ISSUE #1484: Show a button to support users
Our current "Send Feedback" button at the bottom left redirects to the current survey, we should make is a little more visible and actually redirect to Discord, with a "Got an issue?" CTA, so that we can talk to those users that have a problem right away.

Tbaut:
closes #1484

This PR also fixes a little glitch in the interface where the display and profile icon had the wrong color
before:
image

after:
image

todo:
The colors for the banner may change, I'll update


FSM1 (approved):

asnaith:
In comparison, the current version has a subtler aesthetic (I have a personal preference for that 😅) but it's definitely more noticeable now if the goal is to draw attention to it. 

Current: 
Screen Shot 2021-09-03 at 11 11 31 AM
Screen Shot 2021-09-03 at 11 11 53 AM

These changes: 
Screen Shot 2021-09-03 at 11 15 12 AM
Screen Shot 2021-09-03 at 11 15 24 AM

asnaith (approved):

Tbaut:
Yeah the problem is that ppl use our survey to complain, and say "make it work", but we have no idea what's the problem. To be able to troubleshoot, we would need to talk to them. I guess Cindy can come with a better design, but the idea is def. to have this visible and chat on Discord directly if possible.

PR #1489, opened 2021-09-02
Add ui tests for file preview

### LINKED ISSUE #1481: Add ui tests for preview feature
Add test coverage for the preview feature in Files UI

asnaith:
closes #1481


packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    it("can see option to download file from the preview screen", () => {
+      cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })
+      homePage.uploadFile("../fixtures/uploadedFiles/file.zip")
+      homePage.fileItemName().eq(0).invoke("text").as("fileNameA")
+      homePage.fileItemKebabButton().click()
+      homePage.previewMenuOption().click()
+      previewModal.previewKebabButton().click()
+      previewModal.downloadFileButton().should("be.visible")
+    })
```
asnaith, line 62: This test is not as useful as I'd have liked it to have been. Originally I wanted to inspect the `a` tag for the download button but it's not displayed in the DOM (I think because it's handled by an `onClick` event). 

I could do some work to actually perform the download and verify that the correct file was downloaded but it's a bit more involved and seems like that would be testing the browser capability more than our app?

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,76 @@
+import { homePage } from "../support/page-objects/homePage"
+import { previewModal } from "../support/page-objects/modals/previewModal"
+
+describe("File Preview", () => {
+
+  context("desktop", () => {
+
+    it("can preview and browse compatible file types", () => {
+      cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })
```
Tbaut: We don't need to clear the trash here I guess since it's not used.
```suggestion
      cy.web3Login({ clearCSFBucket: true })
```

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      // add files
+      homePage.uploadFile("../fixtures/uploadedFiles/logo.png")
+      homePage.uploadFile("../fixtures/uploadedFiles/text-file.txt")
+
+      // store their file names as cypress aliases for later comparison
+      homePage.fileItemName().eq(0).invoke("text").as("fileNameA")
+      homePage.fileItemName().eq(1).invoke("text").as("fileNameB")
+
+      // navigate to the preview modal for the first file
+      homePage.fileItemKebabButton().eq(0).click()
```
Tbaut: would use .first here for readability.
```suggestion
      homePage.fileItemKebabButton().first().click()
```

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      // close the preview modal to return to the home page
+      previewModal.closeButton().click()
+      previewModal.body().should("not.exist")
+      homePage.appHeaderLabel().should("exist")
+    })
+
+    it("can see option to download file from the preview screen", () => {
+      cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })
+      homePage.uploadFile("../fixtures/uploadedFiles/file.zip")
+      homePage.fileItemName().eq(0).invoke("text").as("fileNameA")
```
Tbaut: same here
```suggestion
      homePage.fileItemName().first().invoke("text").as("fileNameA")
```

packages/files-ui/cypress/tests/file-preview-spec.ts
Tbaut: I think that testing the download here doesn't matter much, what is displayed is what will be downloaded. We should make sure there's no error though and that there's actually something displayed. We can add some cy-data to the image preview.

If we wanted to be sure about what we request from the api is correct we could do a `cy.intercept` on a `POST https://stage.imploy.site/api/v1/bucket/*/download` and check that the request has a path with the right path, e.g `path	"/logo.png"`.

Tbaut (commented):
Nice that we test both the kebab and the double click. I think we should make sure there's something displayed (image/text/anything) and no error.

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      })
+
+      // close the preview modal to return to the home page
+      previewModal.closeButton().click()
+      previewModal.body().should("not.exist")
+      homePage.appHeaderLabel().should("exist")
+    })
+
+    it("can see option to download file from the preview screen", () => {
+      cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })
```
Tbaut:
```suggestion
      cy.web3Login({ clearCSFBucket: true })
```

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      homePage.uploadFile("../fixtures/uploadedFiles/file.zip")
+      homePage.fileItemName().eq(0).invoke("text").as("fileNameA")
+      homePage.fileItemKebabButton().click()
+      homePage.previewMenuOption().click()
+      previewModal.previewKebabButton().click()
+      previewModal.downloadFileButton().should("be.visible")
+    })
+
+    it("can see applicable elements for unsupported files", () => {
+      cy.web3Login({ clearCSFBucket: true, clearTrashBucket: true })
```
Tbaut:
```suggestion
      cy.web3Login({ clearCSFBucket: true })
```

packages/files-ui/cypress/tests/file-preview-spec.ts
asnaith: @Tbaut Ah yes, I added this before I'd clarified bin files were not supposed to show in search and then forgot to remove it. Thanks

RyRy79261 (approved):

packages/files-ui/cypress/tests/file-preview-spec.ts
FSM1: Agreed that there is no need to test the download further. Additionally, as far as I remember, if the preview is shown - i.e. the binary contents of the file are already in memory, no additional request to download is made.

The approach mentioned by @Tbaut could be used if the preview is not supported.

asnaith: @Tbaut @FSM1 Thanks for the suggestions 😃

- I've left this check as is so we'll only check for the presence of a download button for supported files
- I added checks to make sure supported file content is displayed (and no error message)
- Added a check for the download request of an unsupported file

Tbaut:
@asnaith for me to get pinged about "please recheck this PR", please click on the  refresh button. This will request a review again, and bring it to my attention. Otherwise I don't know when you think it's ready again.

image

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    })
+
+    it("can see unsupported file info and download option", () => {
+      cy.web3Login({ clearCSFBucket: true })
+
+      // add an unsupported file
+      homePage.uploadFile("../fixtures/uploadedFiles/file.zip")
+
+      // setup an api intercepter for download requests
+      cy.intercept("POST", "https://stage.imploy.site/api/v1/bucket/*/download").as("downloadRequest").then(() => {
```
Tbaut, line 71: That's awesome

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      homePage.uploadFile("../fixtures/uploadedFiles/logo.png")
+      homePage.uploadFile("../fixtures/uploadedFiles/text-file.txt")
+
+      // store their file names as cypress aliases for later comparison
+      homePage.fileItemName().eq(0).invoke("text").as("fileNameA")
+      homePage.fileItemName().eq(1).invoke("text").as("fileNameB")
+
+      // navigate to the preview modal for the first file
+      homePage.fileItemKebabButton().first().click()
+      homePage.previewMenuOption().eq(0).click()
```
Tbaut, line 21:
```suggestion
      homePage.previewMenuOption().first().click()
```

Tbaut (approved):
Looking great great, and you found a bug with the file path being repeated, which is also great :mushroom: 

Just made 1 suggestion

packages/files-ui/cypress/tests/file-preview-spec.ts
Tbaut: oh well the auto-merge was on so this didn't go in. Not a big deal.

PR #1492, opened 2021-09-03
[Files] - Billing history

### LINKED ISSUE #1458: Billing history
The billing history screen needs to be made under the subscriptions setting page, this is accessed by the links described in #1453 

The view receipt likely will down load a file/pdf (@sweetpea22 to confirm) 

**Additional context**
image

Tbaut:
closes #1458

use the following subscription id to see some billing history: 3344e538-e101-4008-83a2-a5477e6c73ca

e.g comment out the first `useEffect` in `BillingHistory.tsx` and set the initial state to be `const [subscriptionId, setSubscriptionId] = useState("3344e538-e101-4008-83a2-a5477e6c73ca")`


packages/files-ui/src/Components/Pages/BillingHistory.tsx
```diff
@@ -0,0 +1,102 @@
...
+        setSubscriptionId(subscription.id)
+      })
+      .catch((e) => {
+        console.error(e)
+        setIsLoading(false)
+      })
+  }, [filesApiClient])
+
+  useEffect(() => {
+
```
tanmoyAtb:
```suggestion
```

packages/files-ui/src/Components/Pages/BillingHistory.tsx
```diff
@@ -0,0 +1,102 @@
...
+    if(!subscriptionId) return
+    filesApiClient.getAllInvoices(subscriptionId)
+      .then(({ invoices }) => {
+        setInvoices(invoices)
+      })
+      .catch(console.error)
+      .finally(() => setIsLoading(false))
+
```
tanmoyAtb:
```suggestion
```

packages/files-ui/src/Components/Pages/BillingHistory.tsx
```diff
@@ -0,0 +1,102 @@
...
+              type="inherit"
+              size={32}
+            />
+          }
+          {invoices.map(({ paid_on, amount, payment_method, currency, uuid }, index) =>
+            <TableRow
+              type="grid"
+              key={index}
+            >
+              <TableCell align="left">{paid_on ? dayjs.unix(paid_on as unknown as number).format("DD MMM YYYY") : "unknown"}</TableCell>
```
tanmoyAtb: I see that the `paid_on` is Date type,
shouldn't this work ? 
`dayjs(paid_on)`

tanmoyAtb (changes requested):
Some notes on UI - 

- I would like the loading to show up in the middle.
- When there are no invoices we should show "No invoices"
- Got this warning because the loading is under `tbody`
Screenshot 2021-10-26 at 11 47 49 PM

I didn't get any invoices back with the `subscription_id` in the comment. maybe the API
does a mapping with user_id when returning invoices. !

packages/files-ui/src/Components/Pages/BillingHistory.tsx
Tbaut: nice catch, I didn't go over this code again, it used to be something else.

Tbaut:
> I didn't get any invoices back with the subscription_id in the comment. maybe the API
does a mapping with user_id when returning invoices. !

Ah that's annoying, could you check what subscription id you have and ask Arun to add some invoices to you?

The rest should be addresses.

tanmoyAtb (approved):
Looks good to go 👍 

The mobile view could get some adjustments, with the top padding and table cells.
Screenshot 2021-11-04 at 12 14 22 AM

Tbaut:
Good point @tanmoyAtb, I added some padding.
The receipt will change. It'll just be a link

FSM1 (approved):

PR #1515, opened 2021-09-13
Folder uploads 

### LINKED ISSUE #1014: Allow for upload of folders and their contents
When a user drops a folder into the file browser to upload the contents should be parsed into an array of file contents and paths, and these can all be passed to the upload handler.

FSM1:
Closes #1014

Folder uploads on DnD into file browser, as well as DnD into the Upload modal work as expected. I have been unable to force the file picker to accept folders being selected and it always opens the first selected folder, so if anyone has ideas on that, welcome to just update this PR.


asnaith:
@FSM1 Is this for both Files UI and Storage UI?

FSM1:
> @FSM1 Is this for both Files UI and Storage UI?

Only files

packages/files-ui/src/Components/Modules/FileBrowsers/CSFFileBrowser.tsx
```diff
@@ -178,23 +179,12 @@ const CSFFileBrowser: React.FC<IFileBrowserModuleProps> = () => {
...
-    } else {
-      uploadFiles(bucket, files, path)
-        .then(() => refreshContents())
-        .catch(console.error)
-    }
-  }, [addToast, uploadFiles, bucket, refreshContents])
+    const flattenedFiles = await getFilesFromDataTransferItems(fileItems)
+    const paths = [...new Set(flattenedFiles.map(f => f.filepath))]
+    paths.forEach(p => {
+      uploadFiles(bucket, flattenedFiles.filter(f => f.filepath === p), path + p)
```
tanmoyAtb: can we use the `getPathWithFile` function with instead of `path + p`?

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -137,21 +137,22 @@ const FileInput = ({
...
+  const [{ value }, meta, helpers] = useField<Array<File & {path: string}>>(name)
 
   useEffect(() => {
     onFileNumberChange && onFileNumberChange(value.length)
   }, [onFileNumberChange, value.length])
 
   const onDrop = useCallback(
-    async (acceptedFiles: File[], fileRejections: FileRejection[]) => {
+    async (acceptedFiles: any[], fileRejections: FileRejection[]) => {
       const filtered = acceptedFiles.filter((file) =>
```
tanmoyAtb, line 152: better to change this variable to `filteredFiles`

packages/files-ui/src/Components/Modules/FileBrowsers/UploadFileModal.tsx
```diff
@@ -89,15 +89,19 @@ const UploadFileModule = ({ modalOpen, close }: IUploadFileModuleProps) => {
...
+  const onSubmit = useCallback(async (values: {files: Array<File & {path: string}>}, helpers) => {
     if (!bucket) return
     helpers.setSubmitting(true)
     try {
       close()
-      await uploadFiles(bucket, values.files, currentPath)
+      const paths = [...new Set(values.files.map(f => f.path.substring(0, f.path.lastIndexOf("/"))))]
+      paths.forEach(async p => {
+        const filesToUpload = values.files.filter((f => f.path.substring(0, f.path.lastIndexOf("/")) === p))
+        await uploadFiles(bucket, filesToUpload, currentPath + p)
```
tanmoyAtb: same. if we could use the function `getPathWithFile`

tanmoyAtb (commented):
The implementation looks on the point!

Folder uploads on DnD on filebrowser doesn't seem to work!
On checking the network call, I saw that the upload path wasn't right. 

However, I'm not sure this is how we want folder uploads to work! 
We should try to re create the folders structure while uploading, unless its out of scope for this issue!

Sidenote: everywhere else I see on drive or dropbox, there are separate buttons for file and folder uploads, I think we should go for that.

asnaith:
Hey @FSM1 Just tested this out and generally working pretty well for the core functionality. 

I've noticed a few oddities but not completely sure which ones are in the scope of this ticket. I'll let you decide and I can create issues for any of the points we want to address separately out of the context of this PR. 

*Drag and Drop on File Browser and Upload Modal*

- It's possible to upload a folder that has a name over 65 characters or single space but we don't allow users to create folders like that manually 

- macOS uses proprietry ".DS_Store" files which hold attributes/meta-data about the folder containing it, I noticed that these are transferred and show up unhidden in Files (perhaps we may want to ignore and not upload them?). 

- If you drag and drop an empty folder, the empty folder is not created in Files. I'm assuming this is intentional but perhaps it would be worth informing the user that the folder they are trying to move is empty? 

- There's no notification / feedback to the user if they try and drag a drop a folder that has the same name as an existing folder in Files. It will add contained files if they are new but will not upload any files that have the same name as ones that already exist in the folder with the same name.

- There is no upload progress toast being displayed (or is that because of #1518?). I dropped a large folder containing sub-folders and large files and the smaller files were instantly visible but the larger files took longer to upload. Originally I thought something had gone wrong but after waiting a while I did see everything so I just think we need to show some notification to the end user to let them know that a folder upload is still in progress.  

*Drag and Drop on Upload Modal Only*

- If a folder contains child folder structures but no actual files it's not possible to add it to the upload modal (if you drag and drop to the file browser then all of the child folder structure is recreated).

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -137,21 +137,22 @@ const FileInput = ({
   const classes = useStyles()
   const [previews, setPreviews] = useState<any[]>([])
   const [errors, setErrors] = useState<any[]>([])
-  const [{ value }, meta, helpers] = useField(name)
+  const [{ value }, meta, helpers] = useField<Array<File & {path: string}>>(name)
```
Tbaut: I played a bit with this idea and I think it'd be cool to name this type:
```
interface FileWithPath extends File {
  path: string
}
```

And then use it here and below instead of some of the `any`.

Tbaut (commented):
I did expect as well that when I DaD a folder, I expected the structure to remain. But as Tanmoy said, this is maybe out of scope.

Tbaut:
The Storage tests are failing rightfully because storage-ui doesn't compile:
>  | Failed to compile.
 | 
 | ./src/Components/Modules/FilesList/FilesList.tsx
 | Module not found: Can't resolve '../UploadFileModal/UploadFileModal' in '/home/thib/Github/ChainSafe/ui-monorepo/packages/storage-ui/src/Components/Modules/FilesList'

packages/common-components/src/FileInput/FileInput.tsx
FSM1: Problem with this is the Dropzone.onDrop typechecking fails, unless the definition is either `File[]` or `any[]`
image

FSM1:
> Folder uploads on DnD on filebrowser doesn't seem to work!
> On checking the network call, I saw that the upload path wasn't right.

Which browser/OS were you using, and what was the structure of the folder you were trying to upload?
 
> However, I'm not sure this is how we want folder uploads to work!
> We should try to re create the folders structure while uploading, unless its out of scope for this issue!

Problem here is that the folder structure is never intentionally created (there are no mkdir calls) and the folder structure is a by-product of uploading files to the right path. Hence why empty folders are not created. Overall I would say replicating the exact folder structure is less of a priority right now and can always be added in afterwards should users express the need for this sort of functionality.

> Sidenote: everywhere else I see on drive or dropbox, there are separate buttons for file and folder uploads, I think we should go for that.

This could be a solid solution to the Native OS folder dialog issue I mentioned.

FSM1:
> * It's possible to upload a folder that has a name over 65 characters or single space but we don't allow users to create folders like that manually

Hmmm, not exactly sure what the right approach to handle this is. 

> * macOS uses proprietry ".DS_Store" files which hold attributes/meta-data about the folder containing it, I noticed that these are transferred and show up unhidden in Files (perhaps we may want to ignore and not upload them?).

This should be doable, though it becomes a slippery slope for deciding what gets ignored.

> * If you drag and drop an empty folder, the empty folder is not created in Files. I'm assuming this is intentional but perhaps it would be worth informing the user that the folder they are trying to move is empty?

The upload operation only works with files, and the file system gets updated without any calls to `mkdir` jsut by virtue of the upload paths being set.

> * There's no notification / feedback to the user if they try and drag a drop a folder that has the same name as an existing folder in Files. It will add contained files if they are new but will not upload any files that have the same name as ones that already exist in the folder with the same name.


> * There is no upload progress toast being displayed (or is that because of [Fix toasts not rendering #1518](https://github.com/ChainSafe/ui-monorepo/pull/1518)?). I dropped a large folder containing sub-folders and large files and the smaller files were instantly visible but the larger files took longer to upload. Originally I thought something had gone wrong but after waiting a while I did see everything so I just think we need to show some notification to the end user to let them know that a folder upload is still in progress.

This should be resolved in #1518
> 
> _Drag and Drop on Upload Modal Only_
> 
> * If a folder contains child folder structures but no actual files it's not possible to add it to the upload modal (if you drag and drop to the file browser then all of the child folder structure is recreated).

This should not be the case. If there are no files anywhere in the tree being uploaded, then nothing will be updated. The folder upload does not try and recreate the folder structure being dropped, it just parses out all the files (with relative paths) and uploads each of those with a valid absolute path based on the current path + relative path. This should be the same whether a folder is dropped onto the file browser or the Upload modal.

packages/common-components/src/FileInput/FileInput.tsx
Tbaut: I didn't dig too deep there, but I tested the following and it looked fine, what do you think of having the path eventually undefined: 
```
interface FileWithPath extends File {
  path?: string
}
```

I could then type the useField with `useField<FileWithPath[]>(name)`, type the `acceptedFiled` from `onDrop`, and remove the `as any[]` so that the items are correctly typed.

Tbaut:
Proposed an alternative typing to remove many `any`, see https://github.com/ChainSafe/ui-monorepo/pull/1515#discussion_r709936025

RyRy79261 (approved):

packages/common-components/src/FileInput/FileInput.tsx
FSM1: that would work, but then introduces further down-the-line changes such as requiring to check that the path is not undefined

tanmoyAtb (approved):
I tested out the folder uploads again with DnD, working great for me!
The dropzone as well.

This is amazing work! 🎉 

We should have a separate issue for native OS uploads, 
have separate buttons for file and folder uploads!

packages/common-components/src/FileInput/FileInput.tsx
FSM1: Implemented as suggested.

Tbaut (approved):
Awesome! I think having as few `any` as possible is always a better option :)

PR #1518, opened 2021-09-13
Fix toasts not rendering

Tbaut:
I broke the toasts earlier using `useRef`, a `useState` is needed in conjunction to it, to provoke a re-rendering.


tanmoyAtb (approved):
🎉 Amazing! Looks like everything's sorted!

RyRy79261 (approved):

PR #1520, opened 2021-09-13
Add new ui tests for illogical folder and file move errors

### LINKED ISSUE #1511: Add test coverage for move file / folder behaviour
As described - we need to add some coverage around moving files and folders to ensure that no regressions occur with regards to issues we have seen this past (issues like the parent folder being moved into itself etc)

asnaith:
closes #1511


packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -87,9 +87,72 @@ describe("File management", () => {
         homePage.fileItemRow().should("have.length", 2)
         homePage.fileItemName().should("contain.text", folderName)
         homePage.fileItemName().should("contain.text", $fileName)
+
+        // ensure folder already in the root cannot be moved to Home
+        homePage.fileItemName().contains(`${$fileName}`).click()
+        homePage.moveSelectedButton().click()
+        moveItemModal.folderList().contains("Home").click()
+        moveItemModal.errorLabel().should("have.text", "The files are already in this folder")
```
asnaith: I know in general we don't want to be checking string text most of the time but I feel like these strings are unlikely to change, are specific to the move being initiated and help with readability of the test so just an FYI that this was deliberate in this PR (but happy to remove if others really dislike it).

RyRy79261: Yeah this seems like a fragile approach, maybe just check for keywords or something instead of the full string? :)

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -87,9 +87,72 @@ describe("File management", () => {
         homePage.fileItemRow().should("have.length", 2)
         homePage.fileItemName().should("contain.text", folderName)
         homePage.fileItemName().should("contain.text", $fileName)
+
+        // ensure folder already in the root cannot be moved to Home
+        homePage.fileItemName().contains(`${$fileName}`).click()
```
Tbaut, line 92:
```suggestion
        homePage.fileItemName().contains($fileName).click()
```

packages/files-ui/cypress/tests/file-management-spec.ts
Tbaut: I think that checking the full sentence or a keyword ends up being the same. What I'd suggest is to have a `cy-data` to the error (which we prob. have), and check that it's visible, but not be specific about what it contains. The most important part is that there's one displayed, and that the move button is disabled (which we check). What the error contains is a detail.

Tbaut (commented):
I think we can indeed forget about the error message in particular, see my comment.

packages/files-ui/cypress/tests/file-management-spec.ts
asnaith: Ah thanks, copy and pasted this line in from another test where this needs updating too after your explanation before

asnaith: Oh actually, as discussed on slack, this one is needed here when we use `contains` (alternatives are to append `.text()` or `toString()` but neither helps it become more readable so we'll keep as is)

packages/files-ui/cypress/tests/file-management-spec.ts
asnaith: Ok. 

Yep, the error label already has a unique identifier. I've updated the spec file to just check for the presence of the element rather than its text.

Tbaut (approved):
:+1:

FSM1 (approved):
Looks great. Nice to have this covered

PR #1526, opened 2021-09-14
bulk recover operation

### LINKED ISSUE #1508: Recover multiple files from the bin
There is currently no bulk action available to recover a file from the bin, only the bulk deletion is available.

tanmoyAtb:
closes #1508 

Ok, this one was too quick !


Tbaut (approved):
This is awesome, it shows we built things in a generic way :tada:
Working well. I relaunched the test because they fail for weird reasons.

FSM1 (approved):

asnaith:
Working nicely 👍

We already have some tests around single file deletion and I think we should add a check for this kind of operation too, I've added a ticket for myself #1529.

asnaith (approved):

PR #1527, opened 2021-09-15
[Files] - Share to home or other shared folder

### LINKED ISSUE #1507: Copy a file from a shared folder to your home directory or another shared folder
When browsing a shared folder, we should have the ability to copy a file in our home directory, or any other shared folder.
I believe this should be a "share" menu entry similar to the one we currently have in the home directory.

Users could be able to select either another shared folder, or the home.

Open questions:
- should we allow to create a new shared folder at this stage, effectively re-using the modal we currently have
- calling the menu entry "share" is not immediately obvious that this would allow to copy to the home directory

What do you think?

Tbaut:
closes #1507


asnaith:
Looks good in general when I tested it out although I do agree with your comment in the original ticket about the "Share" menu entry, sharing from a share does seem a little odd. Maybe in this context, we could use "Copy To"? 

Also, I noticed a small quirk where a user that only has view access to the shared folder can still uncheck the "Keep original file" option when attempting to move a file to their home directory. If they do this then the file will still be copied to their home directory but they will also see an error toast: 

Screen Shot 2021-09-15 at 3 49 58 PM

This makes sense (to us) as they don't have the proper rights to delete the file but perhaps the ability to uncheck that "Keep original file" option should be invisible / locked for users that only have read access? This could help avoid user confusion.

Tbaut:
Very good point, Thank you so much for the deep test, I'll make sure this checkbox doesn't appear when you're not a writer.

FSM1 (approved):
Works great 🎉 

I agree that labeling this "Copy to" makes the most sense for me

packages/files-ui/src/Components/Modules/FileBrowsers/ShareModal.tsx
```diff
@@ -205,19 +197,24 @@ const CopyToSharedFolderModal = ({ close, file, filePath }: IShareFileProps) =>
   const { bucket } = useFileBrowser()
   const { profile } = useUser()
   const [nameError, setNameError] = useState("")
+  const inSharedBucket = useMemo(() => bucket?.type === "share", [bucket?.type])
+  const isReader = useMemo(() => {
+    if (!bucket) return false
 
+    return !!(bucket.readers.find(owner => owner.uuid === profile?.userId))
```
tanmoyAtb:
```suggestion
    return !!(bucket.readers.find(reader => reader.uuid === profile?.userId))
```

tanmoyAtb (approved):
Working great !
I love how these implementations are working out with good tweaks. 

There are a couple of hook dependency warnings on the console.

Tbaut:
ah I missed the warning, will check it out thanks

PR #1531, opened 2021-09-16
Don't track before opt-in

Tbaut:
Before this change, we were tracking users by default, even when they hadn't opted-in. This fixes it.


packages/files-ui/src/Contexts/PosthogContext.tsx
```diff
@@ -152,10 +161,11 @@ function usePosthogContext() {
 
 function usePageTrack() {
   const { pathname } = useLocation()
-  const { posthogInitialized } = usePosthogContext()
+  const { hasOptedIn } = usePosthogContext()
+
   useEffect(() => {
-    posthogInitialized && posthog.capture("$pageview")
-  }, [pathname, posthogInitialized])
+    hasOptedIn && posthog.capture("$pageview")
```
FSM1, line 167: I would still check that posthog is initialized here (were the ENV vars provided to this instance?) to prevent calling `.capture` on an uninitialized instance.

packages/files-ui/src/Contexts/PosthogContext.tsx
```diff
@@ -152,10 +161,11 @@ function usePosthogContext() {
...
+  }, [pathname, hasOptedIn])
```
FSM1: I would still check that posthog is initialized here (were the ENV vars provided to this instance?) to prevent calling `.capture` on an uninitialized instance, throwing a `null ref` exception

```suggestion
    posthogInitialized && hasOptedIn && posthog.capture("$pageview")
  }, [pathname, hasOptedIn, posthogInitialized ])
```

FSM1 (changes requested):

packages/files-ui/src/Contexts/PosthogContext.tsx
FSM1: I would also hide the cookie banner if `posthogInitialized` is false

Tbaut: Agreed, I had to bring it back in the context value to be able to use it.

packages/files-ui/src/Contexts/PosthogContext.tsx
```diff
@@ -64,57 +65,63 @@ const useStyles = makeStyles(
...
 
   useEffect(() => {
-    refreshPosthogState()
-  }, [refreshPosthogState])
+    if(!hasTouchedCookieBanner && localStorageGet(TOUCHED_COOKIE_BANNER_KEY) === "false"){
+      setShowBanner(true)
+    } else {
+      setShowBanner(false)
+    }
+  }, [hasTouchedCookieBanner, localStorageGet])
```
FSM1:
```suggestion
    if(posthogInitialized && !hasTouchedCookieBanner && localStorageGet(TOUCHED_COOKIE_BANNER_KEY) === "false"){
      setShowBanner(true)
    } else {
      setShowBanner(false)
    }
  }, [posthogInitialized, hasTouchedCookieBanner, localStorageGet])
```

FSM1 (changes requested):
Just the update to only show the cookie banner if posthog is initialized.

FSM1 (approved):

RyRy79261 (approved):

PR #1534, opened 2021-09-16
[Files] - Show current bucket decryption key

Tbaut:
This fixes a problem where we were showing the users current key even in the shared folders.


FSM1 (approved):

asnaith:
I wasn't too sure how to properly test this. I do see less keys in the local storage when comparing with the same account on stage though.

FSM1:
> I wasn't too sure how to properly test this. I do see less keys in the local storage when comparing with the same account on stage though.

@asnaith previously if getting the file info, even if the file was in a shared bucket, the key displayed would be for the users 'CSF` bucket, not the key that is required to decrypt this current file.

This PR fixes that to show the Shared bucket's encryption key.

To test, view the file info for a file in the CSF bucket, and compare the key displayed, with the key displayed when viewing the info of a file in a shared bucket. If the keys are different then the change worked.

asnaith:
> To test, view the file info for a file in the CSF bucket, and compare the key displayed, with the key displayed when viewing the info of a file in a shared bucket. If the keys are different then the change worked.

@FSM1 Awesome! Thank you for the explanation, I understand it properly now 👍. I can see the difference between this and what's currently in stage, looking good! 😃

asnaith (approved):

PR #1536, opened 2021-09-16
Add test coverage for storage summary adjustments

### LINKED ISSUE #1530: Add test for storage space summary updating
As described, add basic check to ensure the space summary is being changed after upload / deletion 

(we will add more checks regarding the storage summary when we have automated the sharing feature).

asnaith:
closes #1530

I have to inspect the contents of the label for this test as there's no other way to check that it was updated after upload / deletion. 

Important we have test coverage here as we've seen it not update at the right moment in the past (will be adding more scenarios once the sharing feature is automated too)


packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -350,5 +350,29 @@ describe("File management", () => {
...
+
+      // upload a file and ensure the storage space label adjusts
+      homePage.uploadFile("../fixtures/uploadedFiles/logo.png")
+      navigationMenu.spaceUsedLabel().should("not.contain.text", "0 Bytes")
+
+      // delete the file from the bin and ensure the storage space label adjusts
+      homePage.fileItemKebabButton().click()
+      homePage.deleteMenuOption().click()
+      deleteFileModal.confirmButton().safeClick()
+      homePage.deleteSuccessToast().should("not.exist")
```
Tbaut, line 369: Now that we have the ability to close the toasts, I think it'd be cool do it, this would speed up tests a bit rather than waiting 4 sec every time.

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -350,5 +350,29 @@ describe("File management", () => {
...
+      navigationMenu.binNavButton().click()
+      binPage.fileItemKebabButton().click()
+      binPage.deleteMenuOption().click()
+      deleteFileModal.confirmButton().safeClick()
+      binPage.permanentDeleteSuccessToast().should("not.exist")
```
Tbaut, line 374: same here

Tbaut (approved):
Looks good, just a comment maybe for a future issue, I think we should stop waiting for the toasts to disappear, and rather close them. This will most probably require a custom testId added to the close button in the toasts, I can help with that.

packages/files-ui/cypress/tests/file-management-spec.ts
asnaith: @Tbaut Yeah, it will definitely speed up the tests but waiting for them to disappear adds a lot of reliability as we know for certain we are not doing anything else before the upload / delete is fully complete. 

Without waiting for the toast to disappear I have seen that sometimes the UI does not update quick enough before performing the next actions / assertions and as we don't want to add hardcoded waits this was the best thing I could think of. 

I'm not sure if dismissing the toast will give us the same level of reliability but definitely something we can try (is it all good if we look into that separately out of the context of this PR?).

RyRy79261 (approved):

packages/files-ui/cypress/tests/file-management-spec.ts
Tbaut: > we know for certain we are not doing anything else before the upload / delete is fully complete.

Hmm the moment the toast for success appears, we should 100% have the UI re-render, and actually see the results of our action. Cypress has a default 4s timeout, which should really suffice IMHO

asnaith: @Tbaut Ah ok I think it's because all of the uploading status toasts have the same identifier at the moment then, maybe I need to specifically put one on the "upload complete" status and then dismiss. 

We can address this separately in this issue : https://github.com/ChainSafe/ui-monorepo/issues/1544

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -870,6 +870,7 @@ const FilesList = ({ isShared = false }: Props) => {
                       <CheckboxInput
                         value={selectedItems.length === items.length}
                         onChange={() => toggleAll()}
+                        testId="select-all"
```
asnaith, line 873: Thanks for this update to the way we add identifiers to checkbox's @Tbaut 👍

PR #1549, opened 2021-09-21
Upgrade Torus dependencies

### LINKED ISSUE #1541: DirectAuth Init fails in Firefox Private mode
**Describe the bug**
When the app is loaded in Firefox + Private the login buttons remain disabled.

**To Reproduce**
1. Open FF + Private
2. Navigate to app.files.chainsafe.io
3. Observe that the Login buttons are never enabled

**Expected behavior**
The login buttons should enable once the necessary dependencies have loaded, allowing a user to log in.

**Console error**
`Error: Please serve redirect.html present in serviceworker folder of this package on http://localhost:3000/serviceworker/redirect`

**Screenshots**
image

**Additional context**
The error returned if from the `DirectAuthSDK.init()` call failing. According to Torus, solution is to either pass in a `skipInit: true` or alternatively update the SDK.

FSM1:
Closes #1541

Left the tkey dependencies at V3, since V4 introduced some breaking changes


Tbaut (approved):

Tbaut:
> since V4 introduced some breaking changes

Breaking, as in we can't fix them ever? Would be good to keep up with the new versions if we can.

FSM1:
> > since V4 introduced some breaking changes
> 
> Breaking, as in we can't fix them ever? Would be good to keep up with the new versions if we can.

`breaking` as in SDK methods that we currently depend on are no longer available in v4, and it might require some re-architecture of the tkey context to replace the need for these.

Created #1552 to upgrade the TKey SDK

Tbaut:
ah I see, thanks for the details and the issue :)

tanmoyAtb:
I tested out this out on firefox and firefox incognito, getting errors thrown on firefox,
Running great on chrome and brave,

https://user-images.githubusercontent.com/10713353/134222380-9775bcd0-a4c3-43a7-a0b0-0c15c3892bc7.mov

Attaching the error I got here: 

Screenshot 2021-09-21 at 11 51 35 PM

FSM1:
@tanmoyAtb will check again with oauth in firefox. I only tested web3.

RyRy79261 (approved):

FSM1:
> @tanmoyAtb will check again with oauth in firefox. I only tested web3.

Tested this out and oauth is working in both FF, FF+Private

tanmoyAtb (approved):
Github is working fine in FF and FF private,
Will check on google when we get this into staging.

PR #1576, opened 2021-09-23
Hide menu when there's no menu item

### LINKED ISSUE #1560: Hide kebab menu on child folders in share for users with view-only access
**Description**
When a user with view-only access enters a shared folder containing a child folder they will see the kebab menu alongside the child folder. 

Because the view-only user does not have access to Rename / Move / Delete options no full menu is presented when they click on the kebab menu icon, just a tiny box. 

**Steps to Reproduce**
1. Log in to files with a view-only user
2. Navigate to a shared folder containing a child folder
3. Click the kebab menu for the child folder 

**Observed behavior**

https://user-images.githubusercontent.com/3982665/134430951-fb194c81-9ae5-49cd-b34e-dd375eb96a53.mov

**Expected behavior**
If there are no actions the view-only user can perform on the child folder then it would be best not to display the kebab menu. 

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Version : 93.0.4577.63

Tbaut:
closes #1560


asnaith (approved):
Looking good now 😃

FSM1 (approved):

PR #1582, opened 2021-09-24
Upgrade api client and fix issue with shared folder renaming

### LINKED ISSUE #1563: Owner cannot rename shared folder
**Describe the bug**
An owner cannot rename a shared folder. 

This looks like a regression bug as it seems like we previously fixed it in the past (see https://github.com/ChainSafe/ui-monorepo/pull/1197)

**To Reproduce**
1. Log in to files
2. Create a shared folder
3. Attempt to rename the shared folder you just created

**Expected behavior**
An owner should be able to change the name of the shared folder

**Video**

https://user-images.githubusercontent.com/3982665/134435236-651de7e4-58cd-47f7-8058-6a2dd5a0b2f2.mov

**Desktop:**
 - OS: macOS
 - Browser: Chrome (not browser specific)
 - Version : 93.0.4577.63

Tbaut:
closes #1563 

This ended up being much more involved than originally anticipated. The good point is that we end up with more information accessible in one place: the `BucketKeyPermission` type now also has all the user bucket info including names etc..


packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -170,6 +173,25 @@ const FilesProvider = ({ children }: FilesContextProps) => {
     return encryptionKey
   }, [getKeyForSharedBucket, personalEncryptionKey, userId])
 
+  const enrichUserInfo = useCallback((bucketUser: BucketUser[], lookupUser: LookupUser[]): RichUserInfo[] => {
+    console.log("bucketUser", bucketUser)
+    console.log("lookupUser", lookupUser)
```
Tbaut:
```suggestion
```

asnaith (approved):
Nice, the fix works well for the original issue. 

I also did some exploratory testing around file moving, sharing, recovering, deleting and uploading. I didn't run into any problems during that, all was functioning as intended 🙂

RyRy79261 (approved):

PR #1590, opened 2021-09-28
Posthog identify user on login

FSM1:
Links the posthog device id with a user uuid, and clears this when the user logs out

### DIFF
--- a/packages/files-ui/src/Contexts/PosthogContext.tsx
+++ b/packages/files-ui/src/Contexts/PosthogContext.tsx
@@ -5,6 +5,7 @@ import { createStyles, makeStyles } from "@chainsafe/common-theme"
 import { Trans } from "@lingui/macro"
 import { useLocalStorage } from "@chainsafe/browser-storage-hooks"
 import { CSFTheme } from "../Themes/types"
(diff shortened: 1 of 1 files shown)


packages/files-ui/src/Contexts/PosthogContext.tsx
```diff
@@ -114,6 +116,14 @@ const PosthogProvider = ({ children }: PosthogProviderProps) => {
     }
   }, [posthogInitialized, touchCookieBanner])
 
+  useEffect(() => {
+    if (profile) {
+      posthogInitialized && posthog.identify(profile.userId)
+    } else {
+      posthogInitialized && posthog.reset()
```
Tbaut, line 123: I'm not sure I understand the use of reset here. Shouldn't this be done on logout?

FSM1: Profile will be reset to undefined when the user logs out, so this achieves the exact same thing. I could just as easily make the call when the logout is clicked, but this would leak the analytics code into components.

Tbaut (approved):

tanmoyAtb (approved):

PR #1596, opened 2021-09-29
Mv and rm batch calls 

### LINKED ISSUE #1556: Update calls to `mv` and `rm` to handle multiple files at one
With the `mv` and `rm` API's now supporting an array of paths as an input we should use this to reduce the number of requests being made when moving/deleting large numbers of files

FSM1:
Closes #1556


packages/files-ui/src/Components/Modules/FileBrowsers/BinFileBrowser.tsx
```diff
@@ -52,83 +52,62 @@ const BinFileBrowser: React.FC<IFileBrowserModuleProps> = ({ controls = false }:
...
+  const deleteItems = useCallback(async (cids: string[]) => {
+    if (!bucket) return
 
-    if (!itemToDelete || !bucket) {
-      console.error("Bucket not set or no item found to delete")
-      return
-    }
+    const pathsToDelete = cids.map((cid: string) => {
+      const itemToDelete = pathContents.find((i) => i.cid === cid)
+      if (itemToDelete) {
```
Tbaut: Nit, for readability I find it easier when we earli return `if(!itemTodelete)` and then, in the normal case, return `getPathWithFile..`

packages/files-ui/src/Components/Modules/FileBrowsers/BinFileBrowser.tsx
```diff
@@ -52,83 +52,62 @@ const BinFileBrowser: React.FC<IFileBrowserModuleProps> = ({ controls = false }:
...
+  const recoverItems = useCallback(async (cids: string[], newPath: string) => {
+    if (!bucket) return
 
-      refreshContents()
-      refreshBuckets()
-      const message = `${
-        itemToDelete.isFolder ? t`Folder` : t`File`
-      } ${t`deleted successfully`}`
+    const pathsToRecover = cids.map((cid: string) => {
+      const itemToRecover = pathContents.find((i) => i.cid === cid)
```
Tbaut: This logic looks duplicated many times in the PR, can we extract it to a function?

Tbaut (changes requested):

packages/files-ui/src/Components/Modules/FileBrowsers/BinFileBrowser.tsx
tanmoyAtb: As a sidenote on this!
Always felt we should pass path instead of CID in these functions, removes an extra step

FSM1: Using paths/names feels like "looser typing", though it does enable better consistency with the API 🤔 @Tbaut @RyRy79261 thoughts on updating all of these?

Tbaut: I'm not sure what "looser typing" mean :laughing: 
If we don't need those cids.. then yeah giving the path right away would help I guess.

Tbaut (approved):
Wow nice the UX is so much better, it's snappy

RyRy79261 (approved):

tanmoyAtb (approved):
Good to go!

PR #1608, opened 2021-10-04
Fix move single file and test

### LINKED ISSUE #1605: Moving only 1 file at a time fails and renames the file
**Describe the bug**
When you move several files, all good, just moving 1 file fails though, and renames the files with the name of the directory you wanted to move it to.

**To Reproduce**
1. Select a file to move
2. Move it to a folder.
3. The file is actually not moved to the folder, but it's renamed with the name of the folder.
I move "video.mp4", to the folder "empty", and instead of being moved, the file is actually renamed "empty(5)"

https://user-images.githubusercontent.com/33178835/135658067-43b5f6b7-f39c-490c-87bd-a59fcb87c4bb.mp4

Tbaut:
closes #1605

I added a wait for the toast success to disappear in order not to click on the "home" too quickly otherwise it failed.
Also I added another `refreshbucket` after a full deletion from a shared folder so that we refresh the storage used cf https://github.com/ChainSafe/ui-monorepo/pull/1604


FSM1 (approved):
thanks for taking care of this

tanmoyAtb (approved):

asnaith:
Thanks Thibaut!

PR #1610, opened 2021-10-05
Detect and dismiss toasts in ui tests

### LINKED ISSUE #1544: Dismiss toasts instead of waiting for them to disappear
Update the tests so that we close the "upload complete" / "delete successfully" toasts instead of waiting for them to disappear, this should speed up overall test time. 

Acceptance criteria: 

- Unique identifier is added for the upload toast displaying "upload complete" (at the moment all upload-related toasts share the same identifier). 

- Add an identifier to the toast close button that appears when hovering over it

- Update all test steps where we are waiting for toasts to disappear. Replace them with an action to dismiss the toast instead

asnaith:
closes #1544

In our ui tests we have been waiting for toasts to disappear before proceeding to the next test step. This was for test stability/reliability purposes but also meant that we added a lot of unnecessary wait time. 

We will now detect and then dismiss the toast in tests instead, this should give us the same assurance that the test is in a good state before we proceed - but without the unnecessary waiting. 

I've also redefined the toasts as separate test objects because they seem similar to modals (as opposed to properties of a single page). This also improves test readibility.


asnaith:
Hey @Tbaut we spoke about this a couple of weeks back. Do you like this approach? It seems reliable so far and reduces the overall execution test time

packages/files-ui/cypress/support/page-objects/toasts/baseToast.ts
```diff
@@ -0,0 +1,3 @@
+export const baseToast = {
+  closeButton: () => cy.get("[data-testid=button-close-toast]")
```
Tbaut: I'm just worried there are a lot of small files with 1 or 2 lines. Also this one is super generic, you had the ability to check but I wonder if having it generic for all toast won't make our tests less reliable in the case where there are many toasts.

Tbaut (approved):
Looks good overall, I'm not super opiniated regarding the structure, if this works for you then I'm good with it too :)

packages/files-ui/cypress/support/page-objects/toasts/baseToast.ts
FSM1: yeah if our tests ever have 2 toasts, I am guessing it would be hard to close 1 of them and not the other with a common button `data-testid`

FSM1 (approved):

packages/files-ui/cypress/support/page-objects/toasts/baseToast.ts
asnaith: @Tbaut @FSM1 Sounds good. I've updated this so that we are now using a unique close button identifier per toast.

I've left the toasts as separate files for reasons we discussed (conforming to the same as everything else is defined in the test suite) but I've removed `baseToast` as it's no longer needed now that the close buttons are unique.

RyRy79261 (approved):

Tbaut (approved):
wunderbar :)

PR #1618, opened 2021-10-06
Add Support for heic images

### LINKED ISSUE #1558: File preview cannot render .heic images, does not show as unsupported
**Description**
The preview modal cannot display .heic images but instead of saying "File format not supported." it displays a small blue box containing a question mark. 

(.heic files are Apple's proprietary version of the HEIF format)

**Steps to Reproduce**
1) Log in to files
2) Upload a .heic image 
3) Double click the .heic file to enter the preview

**Observed behavior**

heic_files

**Expected behavior**

It would be expected that the user is shown the "File format not supported" message with option to download if the image cannot be displayed. 

**Desktop:**
 - OS: macOS
 - Browser: Chrome (not specific to this browser)
 - Version : 93.0.4577.63

FSM1:
Closes #1558 

Sample image can be obtained [here](https://filesamples.com/formats/heic)


packages/files-ui/src/Components/Modules/FileBrowsers/CSFFileBrowser.tsx
```diff
@@ -155,6 +155,7 @@ const CSFFileBrowser: React.FC<IFileBrowserModuleProps> = () => {
   const handleUploadOnDrop = useCallback(async (files: File[], fileItems: DataTransferItemList, path: string) => {
     if (!bucket) return
     const flattenedFiles = await getFilesFromDataTransferItems(fileItems)
+    console.log(flattenedFiles)
```
Tbaut:
```suggestion
```

packages/files-ui/src/Components/Modules/FilePreviewModal.tsx
```diff
@@ -196,6 +197,13 @@ const FilePreviewModal = ({ file, nextFile, previousFile, closePreview, filePath
     delta: 20
   })
 
+  const validRendererMimeType = content_type &&
```
Tbaut: I realize this is just a code move but I guess we can use `useMemo` to prevent recalculating it every render.
I'd also suggest to make sure it's a boolean (with `!!` and check for the length of the find?) and give it a more evocation name such as `isValidRendererMimeType`

packages/files-ui/src/Components/Modules/PreviewRenderers/ImagePreview.tsx
```diff
@@ -41,17 +43,28 @@ const useStyles = makeStyles(
     })
 )
 
-const ImagePreview: React.FC<IPreviewRendererProps> = ({ contents }) => {
+const ImagePreview: React.FC<IPreviewRendererProps> = ({ contents, contentType }) => {
   const [imageUrl, setImageUrl] = useState<string | undefined>()
-
   useEffect(() => {
-    setImageUrl(URL.createObjectURL(contents))
+    const handleCreateImageUrl = async () => {
```
Tbaut: The construction of an async declared, and called right away in a `useEffect` always makes me worry. Can we use a .then.catch instead?

Tbaut (commented):
I can't comment on the functionality since I don't have any heic image at hand, but left a couple comments.

RyRy79261 (commented):
Looks good but I'll review again once Thibaut's comments are resolved, guessing error handling will be added

packages/files-ui/src/Components/Modules/FilePreviewModal.tsx
FSM1: this  is actually meant to be a `string | undefined` as the value is later used as a key to look up the correct preview renderer component

Tbaut: aaah ok, didn't realize.

Tbaut:
~The test failing seems not related.~ edit: oh yeah actually it is
I went ahead and tried to upload an `heic`, now for some reason I see nothing :shrug: 
Also there's no error.
On Brave and FF. 

image

packages/files-ui/cypress/tests/file-preview-spec.ts
```diff
@@ -72,10 +72,11 @@ describe("File Preview", () => {
         homePage.fileItemName().dblclick()
         previewModal.unsupportedFileLabel().should("exist")
         previewModal.downloadUnsupportedFileButton().should("be.visible")
-        cy.get("@downloadRequest").should('be.undefined')
+        // ensure that the file download does not start until the download button is clicked
+        cy.get("@downloadRequest").should('be.null')
         previewModal.downloadUnsupportedFileButton().click()
         // ensure the download request contains the correct file
-        cy.get("@downloadRequest").its("request.body").should("contain", {
+        cy.get("@downloadRequest").should('not.be.null').its("request.body").should("contain", {
```
Tbaut: this seems to fail "null should not be null"

FSM1: Yeah there was a change in the PreviewModal which broke this test.

asnaith:
- I noticed what Thibaut mentioned but if I wait longer the image does appear (but after the progress bar has disappeared). So seems to be working but with a bit of a performance issue. 
- Also noticed if you try to move to other files before the preview for a .HEIC file has finished or back to one that wasn't then the next preview will sometimes fail to load
- The white arrow within the right-hand side next button is sometimes disappearing

These issues can be seen in the video below. I initially preview a .jpeg and then try to open the subsequent .heic files after it. 

https://user-images.githubusercontent.com/3982665/136428726-0377827f-7c22-4a4c-bd64-bcfcfa54a507.mov

tanmoyAtb:
The heic files are previewing well in my system. No issue with any of my browsers.
The arrow disappearing problem looks like an active color problem. Happens for both buttons (should be a separate issue).

Only thing, converting the heic -> png while we have the progress would give a better UX. (even if we have the progress at 100 works). Takes 5 - 10 seconds to bring up the preview on my machine!

FSM1:
> Only thing, converting the heic -> png while we have the progress would give a better UX. (even if we have the progress at 100 works). Takes 5 - 10 seconds to bring up the preview on my machine!

The loading bar is actually part of the file-preview, whereas this coversion to png only happens in this specific case. 

I will add in an indeterminate loader on the ImagePreview (if the file is HEIC).

asnaith:
I am still noticing some quirky behaviour in chrome which can be seen on this video (turn sound on to hear me talk through what I'm seeing). 

https://user-images.githubusercontent.com/3982665/136471070-9ea18468-14e6-481f-b6ad-a98d970c2610.mov

The images used in that test can be found below if you want to try them your side

[heic_images.zip](https://github.com/ChainSafe/ui-monorepo/files/7307292/heic_images.zip)

Tbaut:
I've tried to let the tab open for minutes, and still can't see the image, with the one you provided @asnaith so just forget about me, if you can see it, it's already better than what it was

FSM1:
> I am still noticing some quirky behaviour in chrome which can be seen on this video (turn sound on to hear me talk through what I'm seeing).
> 
>  preview-quirks.mov 
> The images used in that test can be found below if you want to try them your side
> 
> [heic_images.zip](https://github.com/ChainSafe/ui-monorepo/files/7307292/heic_images.zip)

I've also noticed 
- the previous image flashing - Not too sure what to do here. I've got a hunch and will try and implement it to see if it helps.
- The time it takes after the conversion is done to actually render the image - This I also have a hunch on how to fix (or at least make this less noticable)
- the spinner not spinning - 🤷🏼 Will ask @RyRy79261 why this is misbehaving. Its a simple component so should #justwork

Not sure what to do about Thibaut's Linux issue.

asnaith:
I pushed an update to the image preview test, all seems good now. 

Before the action to click the download button the test will attempt to retrieve the intercept alias (that's spying on our "POST"). It checks that the alias returns null, which ensures the download request hasn't happened before the button click.

Tbaut:
Checked again with the png->jpg but I still can't see it the image :(

FSM1:
> Checked again with the png->jpg but I still can't see it the image :(

Why does linux hate me :'(

RyRy79261:
@FSM1 I uploaded one of the sample images and it didn't render despite loading without error
image

FSM1:
> @FSM1 I uploaded one of the sample images and it didn't render despite loading without error 

Yeah this seems to be a linux issue with the heic-convert package 🤷🏼 Guess as long as it works on MacOS, iOS this is still better than nothing.

asnaith:
On macOS - I'm still seeing the issue with the first heic image being previewed where it does not obey the image constraints (but subsequent ones do). 

The file format change seems to have improved load performance and the flickering and stuck loader problems are still there but not as noticeable as before.

FSM1:
> On macOS - I'm still seeing the issue with the first heic image being previewed where it does not obey the image constraints (but subsequent ones do).

I have refactored some of the markup to resolve this issue.

asnaith:
Seeing the first image displayed normally after the markup fix :) 

As the stuck spinner issue is going to be addressed separately I think the only remaining issue is the wrong image display problem (the one shown in the video posted earlier when browsing between many .heic files).

packages/files-ui/src/Components/Modules/PreviewRenderers/ImagePreview.tsx
```diff
@@ -41,67 +44,84 @@ const useStyles = makeStyles(
...
+const ImagePreview: React.FC<IPreviewRendererProps> = ({ contents, contentType }) => {
   const [imageUrl, setImageUrl] = useState<string | undefined>()
-
+  const [loading, setLoading] = useState(false)
   useEffect(() => {
-    setImageUrl(URL.createObjectURL(contents))
-
-    return () => {
-      imageUrl && URL.revokeObjectURL(imageUrl)
+    console.log("creating new link")
```
tanmoyAtb:
```suggestion

```

packages/files-ui/src/Components/Modules/PreviewRenderers/ImagePreview.tsx
```diff
@@ -41,67 +44,84 @@ const useStyles = makeStyles(
...
+      setLoading(true)
+      contents.arrayBuffer()
+        .then(b => heicConvert({
+          buffer: Buffer.from(b),
+          format: "JPEG",
+          quality: 0.5
+        }))
+        .catch(console.error)
+        .then(c => setImageUrl(URL.createObjectURL(new Blob([c]))))
+        .then(() => console.log("created new link"))
```
tanmoyAtb: stray log !

tanmoyAtb (approved):
The image change is working fine now, my side  :)

packages/files-ui/src/Components/Modules/PreviewRenderers/ImagePreview.tsx
FSM1:
```suggestion
```

asnaith:
This is working great since the last changes 🎉

I tested it out in Chrome, Firefox and Safari on macOS and Safari on iOS. 

The display order, rendering and conversion performance for .heic all look good now. Awesome!

asnaith (approved):

PR #1620, opened 2021-10-07
Link-sharing landing page

### LINKED ISSUE #1474: Join share by opening a shared folder invite link
A user who receives this link (after login) is directed to a page that will parse this link, extract the encryption key, encrypt the key with their pubKey.

In the event of a user not being logged in the login page should notify the user that they need to log in to access the shared content.

This page will then make an API request presenting the encryptedEncryptionKey, and the JWT.

The API will validate the JWT, make sure it was signed by the Shared Folder owner, and if all this info matches up, then the user making the request is added to the share.

The app would redirect the user to /shared/{bucketId}/

Tbaut:
closes #1474


asnaith:
Hey @Tbaut, the main flow is working really well for me. I tested this in conjunction with the link creation. 

small suggestions: 

- The terms "read access", "edit access" might be more universally understood than "rights", what do you think?

- Someone receiving the link may not know what kind of rights they have been given so maybe it would be good to display their access level in the "You were added to the shared folder" modal? 

quirks: 

- If a user joins a share via a link that gives them read access and they subsequently click another link granting edit access their access right level is not updated and they remain with read access.

- When a user tries to click on a link that has been deleted the modal will stay in this state indefintiely (screenshot below) and the user is never added. Perhaps we should inform the user that the link is no longer valid in this scenario. 

Screen Shot 2021-10-08 at 8 20 53 PM

- The "Create a Shared Folder" button is no longer responding to click and showing the modal for me (firefox and chrome)

Screen Shot 2021-10-08 at 8 37 14 PM

- Edge case: If a user joins a share via a link and then subsequently clicks the link again they will see the same modal informing them that they were just added. Perhaps in this scenario, we could inform the user they already have access? 

- Edge case: Maybe we should put a cap on the amount of links that you can generate? 😅 It can become hard to navigate and it becomes unresponsive if you try to delete multiple links in quick succession. 

Screen Shot 2021-10-08 at 8 08 41 PM

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,150 @@
...
+      .catch((e:any) => {
+        console.error(error)
+        setError(e.message)
+      })
+      .finally(() => {
+        refreshBuckets()
+      })
+  }, [encryptedEncryptionKey, error, filesApiClient, jwt, refreshBuckets])
+
+  const onBrowserBucker = useCallback(() => {
```
tanmoyAtb: should this be `onBrowseBucket` ?

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,150 @@
...
+    !!newBucket && setIsLoading(false)
+  }, [newBucket])
+
+  useEffect(() => {
+    if(!jwt || !encryptedEncryptionKey) return
+
+    filesApiClient.verifyNonce({ jwt, encryption_key: encryptedEncryptionKey })
+      .catch((e:any) => {
+        console.error(error)
+        setError(e.message)
```
tanmoyAtb, line 98: not sure if we should just set the error from API, this would miss translations,
We can change this when error structure PR goes through !

tanmoyAtb (changes requested):
Working very well !

While testing I found on behaviour !
I created read and edit rights links. 

When another user opens the read rights link first  and then edit rights link, 
He does not receive edit rights, but remains as reader !

tanmoyAtb (approved):

Tbaut:
> He does not receive edit rights, but remains as reader !

Thanks @asnaith and @tanmoyAtb for the review, the edit after read not being taken into account is an api issue, I'll ask them. Similarly for the link check, I think there's an issue to implement an endpoint to check whether or not a nonce is still active, but I think it hasn't been implemented yet.

Tbaut:
>    When a user tries to click on a link that has been deleted the modal will stay in this state indefintiely (screenshot below) and the user is never added. Perhaps we should inform the user that the link is no longer valid in this scenario.

I asked the api team about it. They handle 'nonce not found' as a 404, which I think is not super elegant. I've requested an api change for this.

>    The "Create a Shared Folder" button is no longer responding to click and showing the modal for me (firefox and chrome)

Ah, I had seen this and I fixed this in the branch above, but I didn't pull it. will do.

>    Edge case: If a user joins a share via a link and then subsequently clicks the link again they will see the same modal informing them that they were just added. Perhaps in this scenario, we could inform the user they already have access?

Indeed, I'll check this out.

> Edge case: Maybe we should put a cap on the amount of links that you can generate? sweat_smile It can become hard to navigate and it becomes unresponsive if you try to delete multiple links in quick succession.

I think this should be an api limitation if any. I guess that's a too edgy case for me to handle this client side.

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
Tbaut: yes!

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
Tbaut: The thing is, we should handle common errors (when there's a revoked nonce or anything) but we can't catch and try to translate any possible error, this would never work. So We could always translate a "an error occured: [api error]" and leave the api error. This way of doing would be very useful for us to troubleshoot and support, rather than having a generic "oops something went wrong" and not give anything else.

FSM1: Agreed, and this is actually the exact approach I used with the Create Folder Modal - Handle a `409` with a custom message, and if it is some other error, display a generic message, with the original error following it

packages/files-ui/src/Utils/pathUtils.ts
```diff
@@ -103,6 +103,29 @@ export const isSubFolder = (fold1: string, fold2: string) => {
...
+  }
+
+  return decodeURIComponent(arrayOfPaths[2])
+}
+
+// return the hash from #hash
+export const getBucketDecryptionFromHash = (hash: string) => {
+
+  if(!hash.startsWith("#")){
+    console.error("Bucket encryption key extractione error, unexpected hash", hash)
```
FSM1:
```suggestion
    console.error("Bucket encryption key extraction error, unexpected hash", hash)
```

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    encryptForPublicKey(publicKey, bucketDecryptionKey)
+      .then(setEncryptedEncryptionKey)
+      .catch(console.error)
+
+  }, [bucketDecryptionKey, encryptForPublicKey, publicKey])
+
+  useEffect(() => {
+    !!newBucket && setIsLoading(false)
+  }, [newBucket])
```
FSM1: I would remove this effect and handle the `setIsLoading` in the `finally()` below. That way if an error occurs, then the loader can at least be hidden.

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,150 @@
...
+    if(!jwt || !encryptedEncryptionKey) return
+
+    filesApiClient.verifyNonce({ jwt, encryption_key: encryptedEncryptionKey })
+      .catch((e:any) => {
+        console.error(error)
+        setError(e.message)
+      })
+      .finally(() => {
+        refreshBuckets()
+      })
```
FSM1, line 102:
```suggestion
      .finally(() => {
        setIsLoading(false)
        refreshBuckets()
      })
```

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const { redirect } = useHistory()
+  const jwt = useMemo(() => getJWT(pathname), [pathname])
+  const bucketDecryptionKey = useMemo(() => getBucketDecryptionFromHash(hash), [hash])
+  const { filesApiClient } = useFilesApi()
+  const { refreshBuckets, buckets } = useFiles()
+  const { publicKey, encryptForPublicKey } = useThresholdKey()
+  const [encryptedEncryptionKey, setEncryptedEncryptionKey] = useState("")
+  const [isLoading, setIsLoading] = useState(true)
+  const [error, setError] = useState("")
+  const bucketId = useMemo(() => jwt && jwtDecode<{bucket_id?: string}>(jwt)?.bucket_id, [jwt])
```
FSM1: I would advise to wrap the jwtDecode in a `try ... catch` as I managed to crash the app when experimenting with malformed jwt's

FSM1 (changes requested):
Works really nicely. Great work Thibaut 🎉 

Left a couple of small comments on improvements that I can see.

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
Tbaut:
```suggestion
```

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
Tbaut: I'll address this as soon as the `api` is ready for it. (eventually in another PR)

Tbaut:
I addressed all the comments, added some more sugar on the cake (Showing the edit/read rights) and hopefully making it a bit more bullet proof with try/catch.

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,159 @@
...
+      .finally(() => {
+        refreshBuckets()
+      })
+  }, [encryptedEncryptionKey, error, filesApiClient, jwt, newBucket, refreshBuckets])
+
+  const onBrowseBucket = useCallback(() => {
+    newBucket && redirect(ROUTE_LINKS.SharedFolderExplorer(newBucket.id, "/"))
+  }, [newBucket, redirect])
+
+  console.log("!error && !newBucket && !permission", error, newBucket, permission)
```
Tbaut:
```suggestion
```

RyRy79261 (approved):

asnaith:
Looking even better with the latest improvements. Love the extra sugar on top informing the user of the access right level too and I verified the create folder link is working again now 👍. 

- We will have to test access right updating via links once API support is in. 

- Similarly, we will have to test clicking invalid links (created then trashed) once API support is there

- I didn't see any change to this behaviour yet:

>    Edge case: If a user joins a share via a link and then subsequently clicks the link again they will see the same modal informing them that they were just added. Perhaps in this scenario, we could inform the user they already have access?

- One minor colour change is needed for the icon on this modal 

Screen Shot 2021-10-12 at 2 12 26 PM

Tbaut:
I clicked too quickly :see_no_evil: but the good new is, it's just got merged to the other branch :)

> I didn't see any change to this behaviour yet:

I meant to comment on this. So to check this, we'd need to go through all our buckets, if we find one, we'd need to check all the reader/writers/owner and see if what we're requesting is changing. This will quite complicate the code, with some edge-cases, for a pretty minor UX improvement. I thought it wasn't worth it. 

>    One minor colour change is needed for the icon on this modal

Thanks, will do the change in `tbaut-create-link-1473`

PR #1623, opened 2021-10-08
Default card API added

### LINKED ISSUE #1616: Set the default card once the PaymentMethod has been created
Following the successful creation of a payment method, the Billing API needs to be  called to set the newly created PaymentMethod as the default billing card for the account.

The `ApiClient.updateDefaultCard()` method should be called, passing in the `{payment_method_id: 'ID'}` along with the request.

tanmoyAtb:
closes #1616


Tbaut (approved):
When updating the files client, we should do it in all sub packages, but I've a PR doing this shortly after. I didn't test the functionality for the PR, looks harmless.

FSM1 (approved):

PR #1627, opened 2021-10-13
[EPIC] Files Billing

FSM1:
Initial card billing components and flows


packages/files-ui/src/Components/Elements/NotificationsDropdown.tsx
```diff
@@ -0,0 +1,125 @@
...
+  })
+)
+
+interface INotificationsDropdownProps {
+  notifications: {
+    title: string
+    createdAt: Date
+    onClick?: () => void
+  }[]
+}
```
Tbaut: That's a nit as I review the code, we won't bother to open a PR for but we should define a `Notification` interface, and have here `Notification[]` type rather than this lengthy definition.

packages/files-ui/src/Components/Elements/NotificationsDropdown.tsx
```diff
@@ -0,0 +1,125 @@
...
+                  variant="body2"
+                  className={classes.notificationTime}
+                  component="p"
+                >
+                  {dayjs(n.createdAt).fromNow()}
+                </Typography>
+              </div>
+            ))}
+          </div>
+        </ScrollbarWrapper>
```
Tbaut: nit again, just thinking out loud. this is quite hard to read, we should define this dropdown outside of the props, and use it here.

packages/files-ui/src/Contexts/UserContext.tsx
```diff
@@ -114,6 +113,21 @@ const UserProvider = ({ children }: UserContextProps) => {
...
+    filesApiClient.getUserLocalStore()
+      .then((apiStore) => {
+        initLocalStore(apiStore)
+      })
+      .catch((e) => {
+        console.error(e)
+        initLocalStore({})
+      })
+  }, [isLoggedIn, filesApiClient, initLocalStore])
+
```
Tbaut: This is duplicated and should be removed, there's the same code, just above.
edit, there's an indentation issue, that's why it got added I think

packages/files-ui/src/Components/Elements/NotificationsDropdown.tsx
FSM1: This can be taken care of in the #1818 issue

packages/files-ui/src/Contexts/UserContext.tsx
FSM1: Removed the duplicate, great catch.

packages/files-ui/src/Components/Elements/NotificationsDropdown.tsx
FSM1: lets leave all the notifications refactoring till #1818 gets implemented

Tbaut (approved):
:rocket:

PR #1632, opened 2021-10-14
[EPIC] Link Sharing

Tbaut:


Tbaut:
The conflicts should be resolved in https://github.com/ChainSafe/ui-monorepo/pull/1653

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  }, [bucketDecryptionKey, encryptForPublicKey, publicKey])
+
+  useEffect(() => {
+    if(!jwt || !encryptedEncryptionKey || !!newBucket) return
+
+    filesApiClient.verifyNonce({ jwt, encryption_key: encryptedEncryptionKey })
+      .catch((e:any) => {
+        console.error(error)
+        setError(e.message)
```
FSM1:
```suggestion
        setError(e.error.message)
```

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -0,0 +1,159 @@
...
+      .catch(console.error)
+
+  }, [bucketDecryptionKey, encryptForPublicKey, publicKey])
+
+  useEffect(() => {
+    if(!jwt || !encryptedEncryptionKey || !!newBucket) return
+
+    filesApiClient.verifyNonce({ jwt, encryption_key: encryptedEncryptionKey })
+      .catch((e:any) => {
+        console.error(error)
```
FSM1:
```suggestion
        console.error(e)
```

FSM1 (changes requested):
Left some minor suggestions, but besides that, it looks good to go 🥇

tanmoyAtb (approved):

Tbaut:
Thanks @FSM1 I wonder if there's  better way to type these errors. I feel like we'll make the mistake with the `error.error.message` in the future.

FSM1 (approved):

FSM1:
> Thanks @FSM1 I wonder if there's better way to type these errors. I feel like we'll make the mistake with the `error.error.message` in the future.

We really just need to get used to the API error response schema and always ensure to follow it. I've tried casting the object but it just adds unnecessary boilerplate.

Tbaut:
Indeed, found nothing better than casting https://github.com/axios/axios/issues/3612#issuecomment-933263425

PR #1633, opened 2021-10-14
Merge dev to epic/link sharing

Tbaut:


RyRy79261 (approved):

FSM1 (approved):

FSM1:
Not sure if we should wait for @RyRy79261  `Current Product Widget`  to land before we merge this 🤔

Tbaut:
I don't think so, because it's only merged to link-sharing, which will then be in line with dev. It's pretty much in parallel

PR #1634, opened 2021-10-14
[Files] Fix survey banner and sharing explainer

### LINKED ISSUE #1617: Sharing tutorial shows up after a refresh
Subsequent to https://github.com/ChainSafe/ui-monorepo/pull/1591, refreshing the page seems to reset the user localstore.

How to reproduce:
click on files kebab > share and the sharing tutorial will show up.

Where the issue possibly comes from:
A POST request is fired to the `/store` api endpoint setting the `csf.dismissedSharingExplainer` to `false` again.

Tbaut:
closes #1617
- fixes the survey banner that most prob. didn't work well either
- shows the sharing explainer before bulk-sharing

To test, you can add for instance in `SharedFolderOverview` something like this to have 2 buttons to reset the store values
```js
import { DISMISSED_SHARING_EXPLAINER_KEY } from "./hooks/useSharingExplainerModalFlag"
import { DISMISSED_SURVEY_KEY } from "../../SurveyBanner"
import { useUser } from "../../../Contexts/UserContext"

const { setLocalStore } = useUser()

<button onClick={() => setLocalStore({ [DISMISSED_SHARING_EXPLAINER_KEY]: "false" }, "update")}>setExplainer to false</button>
<button onClick={() => setLocalStore({ [DISMISSED_SURVEY_KEY]: "false" }, "update")}>setSurvey to false</button>
```

### DIFF
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/CSFFileBrowser.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/CSFFileBrowser.tsx
@@ -34,6 +34,8 @@ const CSFFileBrowser: React.FC<IFileBrowserModuleProps> = () => {
   const { pathname } = useLocation()
   const currentPath = useMemo(() => extractFileBrowserPathFromURL(pathname, ROUTE_LINKS.Drive("")), [pathname])
   const bucket = useMemo(() => buckets.find(b => b.type === "csf"), [buckets])
+  const { profile, localStore } = useUser()
+  const [showSurvey, setShowSurvey] = useState(false)
 
   const refreshContents = useCallback((showLoading?: boolean) => {
     if (!bucket) return
@@ -50,10 +52,6 @@ const CSFFileBrowser: React.FC<IFileBrowserModuleProps> = () => {
       }).finally(() => showLoading && setLoadingCurrentPath(false))
   }, [bucket, filesApiClient, currentPath])
 
-  const { profile, localStore, setLocalStore } = useUser()
-
-  const showSurvey = localStore && localStore[DISMISSED_SURVEY_KEY] === "false"
-
   const olderThanOneWeek = useMemo(
     () => profile?.createdAt
       ? dayjs(Date.now()).diff(profile.createdAt, "day") > 7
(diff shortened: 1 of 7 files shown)


asnaith:
I was seeing 404's on `https://stage.imploy.site/api/v1/user/store` earlier which was making this awkward to test but it's better now. 

Nice idea for the test buttons, thanks! That made things a lot easier for validating the behavior of the sharing explainer, I can see it working as expected and respecting the key value after refreshes. 

I am not able to see the survey banner though, even when the key is `"csf.dismissedSurveyBannerV3":"false"`.

RyRy79261 (approved):

Tbaut:
@asnaith are you using an old enough account? It needs to be older than 7 days

FSM1:
> @asnaith are you using an old enough account? It needs to be older than 7 days

I've been unable to get the sharing explainer to show up, even though the local-store does not contain a key.

~~The Survey banner is working 100%~~ Just tested out by clearing the `UserStore` and when the API does not respond with keys the banner/explainer are not shown. Also tested using a fresh account, and the sharing explainer also does not show.

Manually setting the keys to `false` results in the banners being displayed.

asnaith:
> @asnaith are you using an old enough account? It needs to be older than 7 days

Ah no I wasn't, I didn't realize this was a pre-requisite.

I've re-tested on my account that was older than 7 days and now I'm seeing the same as what Mike describes. 

The banner / explainer show if the keys are returned from the API or if they are set to false via the test buttons but if the API doesn't return anything initially from `https://stage.imploy.site/api/v1/user/store` then neither of them will be displayed.

Tbaut:
Thanks for the thorough test. I had indeed forgotten to cater for the case when the account is brand new, and there's nothing stored. I now initialize the local (not remote) store to display banners if there is no value from the api returned.

asnaith:
> Thanks for the thorough test. I had indeed forgotten to cater for the case when the account is brand new, and there's nothing stored. I now initialize the local (not remote) store to display banners if there is no value from the api returned.

Hey Thibaut, I'm seeing the same behaviour as before at the moment. 

In normal scenarios it works intended depending on the relevant response from the API, shows if false and hidden if true. Relevant POST is made to update etc. 

However, for new accounts, the store endpoint returns a 404 and neither banner or explainer are displayed. It makes sense not to see the survey banner on a new account (because of the 7 day rule) but I'm presuming there's no reason they shouldn't see the sharing explainer.

Tbaut:
ah thanks, I did not test a truly new account, hence never got this 404. (This doesn't make sens to me, reached out to the api team to change this behavior). We'd need a special error message I imagine. I'll make sure to init the local store in the catch then for the time being until this gets handled specifically.

Tbaut:
Added the initiation, and tested with a fully new account, the sharing tutorial showed up, and the survey didn't as expected.

FSM1 (approved):
tested this out on both old and new accounts, and its working really well 🎉

asnaith:
Yep, working great for me now too. Nice one @Tbaut!

asnaith (approved):

PR #1642, opened 2021-10-18
Updated default card addition API call

tanmoyAtb:


FSM1 (approved):

Tbaut (approved):
Nice, was able to add a card

PR #1643, opened 2021-10-19
Hide network requests from cypress test runner

### LINKED ISSUE #1638: Filter XHR from Cypress test body
At the moment the tests steps are very hard to read in the cypress test runner because it displays every request being made by the app. 

Cypress have already stated they are working on official support but someone has a nice workaround that could be good for us in the interim (see more: https://gist.github.com/simenbrekken/3d2248f9e50c1143bf9dbe02e67f5399) 

Example of the issue is in the screenshot below. We can't see the test steps because the windows is being flooded with all the requests being displayed: 

Screen Shot 2021-10-15 at 8 05 53 PM

asnaith:
closes #1638

- Hides network requests in the window making the cypress test body readable again (interim fix until cypress has an official solution to filter the noise). 

- The tests appear to run faster now. After running the test suite multiple times and seeing some failures I needed to add a few extra checks to ensure that we're not trying to perform some actions too early.

### DIFF
--- a/packages/files-ui/cypress/support/index.ts
+++ b/packages/files-ui/cypress/support/index.ts
@@ -28,5 +28,19 @@ Cypress.on("uncaught:exception", (err) => {
   }
 })
 
+// Hide fetch/XHR requests
+// interim solution until cypress adds configuration support
+// source https://gist.github.com/simenbrekken/3d2248f9e50c1143bf9dbe02e67f5399
+const app = window.top
+
+if(app != null && !app.document.head.querySelector("[data-hide-command-log-request]")) {
+  const style = app.document.createElement("style")
+  style.innerHTML =
+    ".command-name-request, .command-name-xhr { display: none }"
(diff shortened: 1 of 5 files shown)


asnaith:
Before (left), After (right)

before_and_after

FSM1 (approved):

RyRy79261 (approved):

packages/files-ui/cypress/support/index.ts
```diff
@@ -28,5 +28,19 @@ Cypress.on("uncaught:exception", (err) => {
   }
 })
 
+// Hide fetch/XHR requests
+// interim solution until cypress adds configuration support
+// source https://gist.github.com/simenbrekken/3d2248f9e50c1143bf9dbe02e67f5399
```
Tbaut, line 33: :heart:

Tbaut (approved):
The hack is pretty ugly, but the result is beautiful

PR #1648, opened 2021-10-21
add ui test coverage for the survey banner

### LINKED ISSUE #1636: Add ui test coverage for survey banner
Intercept and stub the API request to return `"csf.dismissedSurveyBannerV3": "false"` and build tests that match the expected survey banner behaviour dependant on the key's value.

asnaith:
closes #1636


RyRy79261 (approved):

FSM1 (approved):

packages/files-ui/cypress/tests/survey-banner-spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      // intercept and stub the account creation date to be > 7 days
+      cy.intercept("GET", "https://stage.imploy.site/api/v1/user/profile", (req) => {
+        req.on("response", (res) => {
+          res.body.created_at = profileCreatedDate 
+        })
+      })
+
+      // intercept and stub the response to ensure the banner is displayed
+      cy.intercept("GET", "https://stage.imploy.site/api/v1/user/store", {
+        body: [{ "csf.dismissedSurveyBannerV3": "false" }]
```
Tbaut, line 18: This is not future-proof unfortunately (because of the version). I think we should make sure it answers nothing, an empty object.

packages/files-ui/cypress/tests/survey-banner-spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      // set up a spy for the POST response
+      cy.intercept("POST", "https://stage.imploy.site/api/v1/user/store").as("storePost").then(() => {
+
+        // dismiss the survey banner
+        homePage.closeBannerButton().click()
+        homePage.surveyBanner().should("not.exist")
+
+        // intercept POST to ensure the key was updated after the banner is dismissed
+        cy.wait("@storePost").its("request.body").should("contain", {
+          "csf.dismissedSurveyBannerV3": "true"
```
Tbaut, line 33: same here, this will break as soon as we change the version. We should find a way to assert that `csf.dismissedSurveyBannerV*` is present, and not be specific about the version.

PR #1649, opened 2021-10-21
Change plan widget

### LINKED ISSUE #1455: Select plan modal screen
The plan selection modal needs to be added

This modal should query the API for a list of Products (including a flag indicating whether the current user can migrate to the plan).

If a product is chosen by the user which they are not allowed to downgrade to, then an error message should be displayed notifying the user why it is not possible to migrate to the selected plan. The Submit button should also be disabled.

In order to translate the content of the available plans, the copy will be managed by the UI. **Teams is not included in current scope, please exclude**

The component and data fetching for the plans from the API are already implemented, but the styling and location in the modal need to be updated in line with the mockups below.

After the `Select this plan and pay` is clicked, the default card should be billed, if the user has a card on file. If no card has been added, the the `Add card` modal should be displayed.

Using the schema below as a guide, there will be either an array of plan option data set as constants, indexed by productIDs in the codebase, or bespoke components mapped to the product IDs we receive from the API, that make use of generalised layout components.

Hover/selection states should change the background & border using `animation.transform` as transition timing.

Learn more link should link out to landing page (@sweetpea22  to clarify) 

Suggestion 1, full configuration in constant array:
```ts
interface IPlan {
   title: string
   flag?: string | { color: string, text: string}
   storageOffer: string
   features: string
   priceBreakdown: string | ReactNode
}
export const Plans: {[key: string]: IPlan} = {
   "abc-123": {
       title: "Basic",
       flag?: undefined,
       storageOffer: "1tb",
       features: "Basic permissions controls",
       priceBreakdown: "Free"
   }
}
```

Suggestion 2, bespoke component from generalised layout components: 

```ts
<PlanCard>
   <TopSection>
       <div>
          { Title }
       </div>
       { flag && <Flag>{ flag }</Flag>  }
   </TopSection>
   <Content>
      <Typography> ## Set height
         { storageOffer }
      </Typography>
      <Typography> ## Set height
         { features }
      </Typography>
      <Typography> ## Set height
         { priceBreakdown }
      </Typography>
   </Content>
</PlanCard>
```

**Describe alternatives you've considered**

**Additional context**
image

RyRy79261:
closes #1455


packages/common-components/src/stories/ToggleSwitch.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+  component: ToggleSwitch
+}
+
+const sizeOptions: SizeOption[] = ["large", "medium", "small"]
+
+const actionsData = {
+  onChange: action("onChange")
+}
+
+export const ToggleSwitchDemo = (): React.ReactNode => {
```
tanmoyAtb, line 20: I tried testing out the toggle on storybooks, for some reason the toggle or size knobs aren't doing anything for me. 

The code looks right though? is it working on your side ?

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductViews/SelectPlan.tsx
```diff
@@ -0,0 +1,264 @@
...
+                    }
+                  </Typography>)
+              }
+            </div>
+          </div>)
+        }
+      </section>
+      <section className={classes.bottomSection}>
+        <a
+          href="#"
```
tanmoyAtb: Added an issue, or we may forget about adding the link.
https://github.com/ChainSafe/ui-monorepo/issues/1654

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+  className?: string
+  close: () => void
+}
+
+const ChangeProductModal = ({ active, className, close }: IChangeProductModal) => {
+  const classes = useStyles()
+
+  const { changeSubscription } = useBilling()
+
+  const [slide, setSlide] = useState<"select" | "confirm">("select")
```
tanmoyAtb, line 31: Do we still need this state ?

tanmoyAtb (dismissed):
Overall looks good,
Although the API keeps throwing 500 at me, couldn't test everything. 

The loaders looks too big for my liking. could make them smaller.
The mobile view also looks cramped. Not sure if we have designs for them.

Left a few comments.

Tbaut (commented):
A couple comments on the UI, will dig in the code later.
image
- "current" is hard to read because of the contrast
- "not sure what to pick" font should prob. be bigger
- "Free" disappears when we toggle the annual plan
- button toggle doesn't tell whether or not it's activated. Should it be green then?
- when you select another plan than yours, you can't select the free plan any more, it's disabled. I'd let users actually select it (and disable the apply button)

FSM1 (changes requested):
One flow that has not yet been catered for is if a user attempts to upgrade their subscription but does not have a card added to their account yet.

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductModal.tsx
RyRy79261: @tanmoyAtb So technically no but because I wasn't sure what the state of the other screens in the design are I've left it

CC: @FSM1

FSM1: Yeah given the future plans for this modal, might as well leave this in for now.
image

RyRy79261:
@Tbaut 

- The current tag was improv, so not sure what would be preferred here
- Adjusted the font variant
- There isn't any price data for Free's annual variant so theres nothing to interpret or display 
- Potentially, I just used the design and single colors since it works in dark mode
- Sweet removed the click blocking, though now the user wont know why they cant select a plan until they choose it

packages/common-components/src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,174 @@
...
+interface IToggleSwitch {
+  left: IToggleOption
+  right: IToggleOption
+  onChange(value: any): void
+  injectedClasses?: ToggleSwitchInjectedClasses
+  size?: "large" | "medium" | "small" | number
+  error?: string
+  disabled?: boolean
+  value?: any
+  // name?: string
```
Tbaut, line 98:
```suggestion
```

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductViews/SelectPlan.tsx
```diff
@@ -0,0 +1,268 @@
...
+            <div className={classes.panelBottom}>
+              {
+                plan.prices
+                  .filter(price => price.recurring.interval === interval)
+                  .map(price => <Typography
+                    disabled={price.is_update_allowed}
+                    variant="h5"
+                    key={`${plan.id}-${price.id}`}>
+                    {
+                      price.unit_amount === 0 ? t`Free` : <>{ price.unit_amount }&nbsp;{ price.currency }&nbsp;{ translatedPrice }</>
```
Tbaut:
```suggestion
                      price.unit_amount === 0 ? t`Free` : `${price.unit_amount} ${price.currency} ${translatedPrice}`
```

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+  active: boolean
+  className?: string
+  close: () => void
+}
+
+const ChangeProductModal = ({ active, className, close }: IChangeProductModal) => {
+  const classes = useStyles()
+
+  const { changeSubscription } = useBilling()
+
```
Tbaut:
```suggestion
```

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+      }}
+      testId="change-product"
+    >
+      {
+        slide === "select" && <SelectPlan
+          className={clsx(classes.slide, className)}
+          close={close}
+          next={(productId: string, priceId: string) => {
+            // setSlide("confirm")
+            changeSubscription(productId, priceId).then(() => close())
```
Tbaut:
```suggestion
            changeSubscription(productId, priceId)
              .then(() => close())
              .catch(console.error)
```

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductViews/SelectPlan.tsx
```diff
@@ -0,0 +1,268 @@
...
+  const classes = useStyles()
+  const { getAvailablePlans, currentSubscription } = useBilling()
+  const [selectedPlan, setSelectedPlan] = useState<string | undefined>(currentSubscription?.product.id)
+  const [plans, setPlans] = useState<Product[] | undefined>()
+  const [interval, setInterval] = useState<ProductPriceRecurringInterval>("month")
+
+  useEffect(() => {
+    if(!plans) {
+      getAvailablePlans()
+        .then((plans) => setPlans(plans))
```
Tbaut, line 124:
```suggestion
        .then((plans) => setPlans(plans))
        .catch(console.error)
```

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductViews/SelectPlan.tsx
```diff
@@ -0,0 +1,268 @@
...
+            onClick={() => setSelectedPlan(plan.id) }
+            key={`plan-${plan.id}`}>
+            <div className={classes.panelTop}>
+              <header>
+                <Typography
+                  component="p"
+                  variant="h5">
+                  {
+                    plan.name
+                  }
```
Tbaut:
```suggestion
                  variant="h5"
                >
                  {plan.name}
```

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductViews/SelectPlan.tsx
```diff
@@ -0,0 +1,268 @@
...
+          >
+            <Trans>
+              Cancel
+            </Trans>
+          </Button>
+          <Button
+            disabled={!selectedPlan || selectedPlan === currentSubscription?.product.id}
+            onClick={() => next(
+              selectedPlan as string,
+              plans?.find(plan => plan.id === selectedPlan)?.prices.find(price => price.recurring.interval === interval)?.id as string)}
```
Tbaut: This is quite involved, can we extract this to a separate `onSelecPlan` for instance?
Also the type casting will make the app crash if the `selectPlan` is undefined. we should rather add an early return like `if(!selectedPlan) return`

packages/files-ui/src/Components/Modules/Settings/Products/ChangeProductModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface IChangeProductModal {
+  active: boolean
+  className?: string
+  close: () => void
+}
+
+const ChangeProductModal = ({ active, className, close }: IChangeProductModal) => {
+  const classes = useStyles()
+
```
Tbaut:
```suggestion
```

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/CurrentProduct.tsx
```diff
@@ -62,6 +61,7 @@ const CurrentProduct = ({ className }: ICurrentProduct) => {
   const classes = useStyles()
   const { storageSummary } = useFiles()
   const { currentSubscription } = useBilling()
+  const [changeProductActive, setChangeProductActive] = useState(false)
```
Tbaut: Could we name this `isChangeProductModalVisible`?

packages/common-components/src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,174 @@
+import { createStyles, ITheme, makeStyles } from "@chainsafe/common-theme"
+import clsx from "clsx"
+import React, { ReactNode, useCallback, useMemo, useState } from "react"
+import { Typography } from ".."
```
tanmoyAtb: This removes a circular dependency warrning.
```suggestion
import { Typography } from "../Typography"
```

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/CurrentProduct.tsx
```diff
@@ -98,18 +98,18 @@ const CurrentProduct = ({ className }: ICurrentProduct) => {
...
-              fullsize
-              variant="primary"
-            >
-              <Trans>Change Plan</Trans>
-            </Button>
-          </Link>
+            <Trans>Change Plan</Trans>
+          </Button>
         </div>
+        <ChangeProductModal
```
Tbaut: For consistency with the other modals. Could we show this by default, and remove the `active` from the the props:
```js
  { isChangeProductModalVisible && (
    <ChangeProductModal close={() => setChangeProductActive(false)} />
  )}
```

packages/files-ui/src/Contexts/BillingContext.tsx
```diff
@@ -78,13 +80,40 @@ const BillingProvider = ({ children }: BillingContextProps) => {
...
+  }, [filesApiClient])
+
+  const changeSubscription = useCallback((productId: string, priceId: string) => {
+    return filesApiClient.updateSubscription(productId, {
+      price_id: priceId
+    })
+      .then(() => true)
+      .catch((error: any) => {
+        console.error(error)
+      })
```
Tbaut:
```suggestion
      .catch(console.error)
```

Tbaut (commented):
I went through the code, the "Free" flashing when we change the yearly/monthly toggle isn't a big deal I guess. Regarding the color, the "Current" plan is super hard to read in light mode with black on blue. We should have white on blue

packages/common-components/src/stories/ToggleSwitch.stories.tsx
RyRy79261: Tbh I'm not sure it worked most of the time but not after a reload 

@FSM1 ?

packages/common-components/src/ToggleSwitch/ToggleSwitch.tsx
RyRy79261: @Tbaut So this is there for a formik version that I didn't build out, not sure if we need a formik version though

Tbaut: Ah OK. It's easy enough to add I think, I guess we don't need to keep comments.

Tbaut:
as per my previous comment https://github.com/ChainSafe/ui-monorepo/pull/1649#discussion_r736732845
> type casting will make the app crash if the `selectPlan` is undefined. we should rather add an early return like `if(!selectedPlan) return`

Type casting should be last resort, I'm pretty sure we can be safer than this.

Tbaut (changes requested):
see previous comment

RyRy79261:
@Tbaut The button can only be executed if the value is not undefined, I've added a redundant check to remove the casting

packages/common-components/src/stories/ToggleSwitch.stories.tsx
FSM1: there was a small typing error with the toggle switch overrides in the theme. I think I have gotten it fixed.

Tbaut (approved):
Looking great!

FSM1 (approved):

PR #1658, opened 2021-10-26
Make intercepts in tests less version specific

### LINKED ISSUE #1655: Remove API version from request intercepts
@Tbaut made a good point that the current tests where we are using api stubbing are brittle because we include the full path of the key that is version specific, meaning whenever we update the version of the key we will break the tests.

asnaith:
closes #1655

- I went ahead and loosened the matching on the api endpoint so that the "v1" wasn't required

- ~~I looked a bit further into the code and thought it would be good to try and bring in` DISMISSED_SURVEY_KEY` from the app code and inject it into the test. I'm having issues importing at the moment though. I can't start the test runner because I'll get a webpack compilation issue in cypress because of this import. I'm still trying to work out how to resolve this. Any suggestions @Tbaut , @FSM1 ?~~


packages/files-ui/cypress/tests/survey-banner-spec.ts
```diff
@@ -1,51 +1,52 @@
...
-          res.body.created_at = profileCreatedDate 
+          res.body.created_at = profileCreatedDate
         })
       })
 
       // intercept and stub the response to ensure the banner is displayed
-      cy.intercept("GET", "https://stage.imploy.site/api/v1/user/store", {
-        body: [{ "csf.dismissedSurveyBannerV3": "false" }]
+      cy.intercept("GET", "**/user/store", {
+        body: [{ [`${DISMISSED_SURVEY_KEY}`]: "false" }]
```
asnaith: I'll experiment some more but this seems needed. I wanted to use `[DISMISSED_SURVEY_KEY]` but cypress seems to only see it as a string if written like that.

Tbaut:
I've never been able to import anything cross project unfortunately. This probably needs some webpack magic, if it's even possible.

packages/files-ui/cypress/tests/survey-banner-spec.ts
asnaith: It's not required :)

asnaith:
@Tbaut as discussed, I've just moved the key to one place where we need to update it. This was the second-best solution after not being able to resolve the webpack issue to import the key from app code.

Tbaut (approved):
:+1:

RyRy79261 (approved):

PR #1659, opened 2021-10-26
add ui test coverage for the sharing explainer

### LINKED ISSUE #1637: add ui test coverage for the sharing explainer
Intercept and stub the API request to return `"csf.dismissedSharingExplainer":"false"` and build tests that match the expected sharing explainer behaviour dependant on the key's value.

asnaith:
closes #1637


packages/files-ui/cypress/tests/sharing-explainer-spec.ts
```diff
@@ -0,0 +1,55 @@
...
+        sharingExplainerModal.body().should("not.exist")
+
+        // intercept POST to ensure the key was updated after the explainer is dismissed
+        cy.wait("@storePost").its("request.body").should("contain", {
+          "csf.dismissedSharingExplainer": "true"
+        })
+      })
+    })
+
+    it("User should not see sharing explainer if previously dismissed", () => {
```
asnaith, line 36: This 2nd test is failing although it could be for legitimate reasons. The explainer is showing despite the key being set to true. I've also experimented with hardcoded waits and pauses to ensure it's not a timing issue. 

@Tbaut is there another data point that I need to stub in order for the explainer not to show?

packages/files-ui/cypress/tests/sharing-explainer-spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      cy.intercept("GET", "**/user/store", {
+        body: [{ "csf.dismissedSharingExplainer": "true" }]
```
Tbaut: it's a small thing that made a diff.
```suggestion
        body: { "csf.dismissedSharingExplainer": "true" }
```

packages/files-ui/cypress/tests/sharing-explainer-spec.ts
Tbaut: see my suggestion, the expected answer should be an object. It was stubbed as an array.

asnaith: Ah oops! Thank you so much for spotting that!

RyRy79261 (approved):

FSM1 (approved):

Tbaut (approved):

PR #1692, opened 2021-11-02
Move Material-ui menu to common-components

Tbaut:


Tbaut:
This has been open for way too long. It can be re-opened if we really want to make this happen.

PR #1698, opened 2021-11-03
Release - 2021-11-03

FSM1:


Tbaut (approved):

tanmoyAtb (approved):

asnaith:
It seems like we will be introducing #1699 if we proceed with this release. I can only reproduce this issue on stage, doesn't seem to happen on current production.

asnaith:
Let's bring #1709 into this once it has been merged and then we are good to go for this release :)

asnaith (approved):

PR #1703, opened 2021-11-04
Improve reliability of file moving test 

### LINKED ISSUE #1700: Improve stability of file moving test
I'm noticing some circumstantial failures when moving files but it seems to be due to flakiness in the test. Needs investigating further but some stability improvements are probably required.

asnaith:
closes #1700

This test is performing two file moves but was only dismissing the "move success" toast once (the last time).

This meant the first toast would sometimes be disappearing from the ui around the same time the second toast was expected to appear confusing the test runner and resulting in a "detached from DOM" error. 

It just needed some extra steps to detect and dismiss the first "move success" toast.


asnaith:
...actually, it was more than that 😅

I found waiting for the bucket refresh made this reliable (ran it 100 times locally without fail).

RyRy79261 (approved):

Tbaut (approved):
That's elegant!

PR #1713, opened 2021-11-06
CID cutoff fix

### LINKED ISSUE #1679: File Info Modal: CID text is cut off (Mobile)
On mobile, the CID text is cut off at the edge instead of continuing on the next line.

**Suggested Solution:** 
Allow the text to continue onto the next line instead.

tanmoyAtb:
closes #1679


RyRy79261 (approved):

Tbaut (approved):
Can we have this ellipsis be dynamic (it's a fixed 300) and only show when needed. Right now it's sometimes truncating it although it's not needed

https://user-images.githubusercontent.com/33178835/140725168-5cfd2e46-c2c4-44db-b94d-fc8f39fa3151.mp4

.

Tbaut (changes requested):
Oops, that was a change request

tanmoyAtb:
> Can we have this ellipsis be dynamic (it's a fixed 300) and only show when needed. Right now it's sometimes truncating it although it's not needed
> 
>  Peek.2021-11-08.11-21.mp4 
> .

I actually did try that approach, but the text was just not detecting an overflow !!, I'll give it another go, if it doesn't work, maybe you can take a look.

tanmoyAtb:
> > Can we have this ellipsis be dynamic (it's a fixed 300) and only show when needed. Right now it's sometimes truncating it although it's not needed
> > Peek.2021-11-08.11-21.mp4
> > .
> 
> I actually did try that approach, but the text was just not detecting an overflow !!, I'll give it another go, if it doesn't work, maybe you can take a look.

Alright! got going, the ellipsis is now dynamic :)

asnaith:
I'm not 100% sure if this change is to make the string shrink to fit all screen sizes but I am seeing truncation of the CID on my iPhone 12 Mini.

Tbaut (approved):
Thanks Tanmoy, I guess the trick was the max-width of the parent?

if @asnaith perceived it as a bug, then maybe we should actually have some kind of copy button right next to it (also to the decryption key while we're at it). We can add that to another issue.

FSM1 (approved):

tanmoyAtb:
> Thanks Tanmoy, I guess the trick was the max-width of the parent?
> 
> if @asnaith perceived it as a bug, then maybe we should actually have some kind of copy button right next to it (also to the decryption key while we're at it). We can add that to another issue.

Yup that max-width was the trick.

There already is a `COPY CID` button in the bottom, we can remove that and add one right next to it in another issue!

Tbaut:
yup absolutely

PR #1715, opened 2021-11-08
Add second section for modals and move away from absolute positionning

### LINKED ISSUE #1664: Modals cut off at the top (mobile)
Modal #1: When opening the **Info** menu item in _mobile landscape orientation_, the modal is cut off at the top of the screen with no option to scroll up to see the whole thing. 

The swipe-up gesture moves the background content instead (see gap above **Close** button).

https://images.zenhubusercontent.com/285884697/39c73530-bf2e-4701-9f8d-cbf6f6beea3d/info_modal_cut_off.mp4

Modal #2: Same issue but with the **Share** menu item.

Tbaut:
closes #1664 
I tested some modal such as the sharing explainer.

 todo,test the following: 
- [x] Shared transfer
- [x] Sharing explainer
- [x] Team modal (fake door)
- [x] Folder creation
- [x]  Create or edit shared folder
- [x] File Info
- [x] Move File
- [x] Report File
- [x] Share Modal
- [x] Upload File
- [x] Mobile folder/file edit
- [x] forget browser

To test the 2nd section you can add the prop `subModal={<div>Some section here</div>}` to any `CustomModal` element, I did it in the following video for the Shared modal creation.

https://user-images.githubusercontent.com/33178835/140749074-808a1d3f-8377-4eb4-a63c-8905f5e1975a.mp4


packages/files-ui/src/Components/Modules/FileBrowsers/CreateOrEditSharedFolderModal.tsx
```diff
@@ -201,6 +204,7 @@ const CreateOrEditSharedFolderModal = ({ mode, isModalOpen, onClose, bucketToEdi
       closePosition="none"
       maxWidth="sm"
       testId="create-or-edit-shared-folder"
+      subModal={<div>Second section!</div>}
```
Tbaut:
```suggestion
```

packages/common-components/src/Modal/Modal.tsx
```diff
@@ -190,7 +214,14 @@ const Modal = ({
             <CloseSvg />
           </div>
         )}
-        {children}
+        <ModalBase>
```
RyRy79261: Couldn't we in theory allow an arbitrary amount of inner modals, if its generic, then you just assign the class to the modal base when you pass it in perhaps?

For example, we could have the modal pop up something akin to a metrics dashboard, by dumping a couple different metric modals in as a `ReactNode<ModalBase>[]` (No idea if we can do that typing, just for example) 

Then tweak the wrapper to lay them out differently

packages/files-ui/src/Components/Modules/FileBrowsers/ReportFileModal.tsx
```diff
@@ -26,7 +26,9 @@ const useStyles = makeStyles(
     return createStyles({
       modalRoot: {
         zIndex: zIndex?.blocker,
-        [breakpoints.down("md")]: {}
+        [breakpoints.down("md")]: {
+          paddingBottom: Number(constants?.mobileButtonHeight) + constants.generalUnit
```
RyRy79261, line 30: This has come up a couple times, could we maybe have it as a modal mixin? Not sure if we've defined any but I usually just define them in the `theme.ts` file and import

Tbaut: The thing is that it's not a general thing, and it's depending on the cancel button height. If there's a button, then we should leave some space for it + a little margin. I don't think we can add a mixing that is taking the cancel button height can we?

packages/common-components/src/Modal/Modal.tsx
Tbaut: Tbh I don't want to make this thing even more complex than it already is, if we don't have the need for it now. If the need ever arise, we can def make it an array.

FSM1 (changes requested):
The Delete confirmation needs  padding and radius:
image
Previously: 
image

Tbaut:
Good catch, the `Dialog` was also using `Modal`. It should be fixed now. I checked a couple other modals, and they looked fine, but a proper go around is prob needed again.

asnaith:
I inspected all the modals on the list on the desktop without issue but noticed multiple problems on mobile (iOS)

**New issues:** 
 
- For modals that have a cancel button outside of the modal, when the keyboard is invoked it covers the cancel button (on production the cancel button is adjusted so that it is displayed above the keyboard). 

- "Create or edit shared folder" modal (mobile): After invoking a “Folder name is already in use” error the input bar needs to be tapped a few times before it responds and can be edited again.

- "Create or edit shared folder" modal (mobile): I managed to cause an issue where the cursor was appearing on the wrong part of the screen and not where I was typing. I couldn't nail down reproduction steps but caught it on video. 

https://user-images.githubusercontent.com/3982665/141024860-38bbdb16-09bc-4bc3-87bb-4bcb3db88028.MP4

- "Report File" modal (mobile): The decryption key is being cut off



- Rename file modal (mobile): The modal is fixed in the middle but the background is scrollable (this happens on all but it is more noticeable on smaller modals like this). 

https://user-images.githubusercontent.com/3982665/141025095-8c8f1672-743c-4ea8-a33a-c2600138713b.MP4

- "Manage Access" modal (mobile): Only 2 / 3 characters of the share link url are shown the rest of the string is truncated due to lack of room (side note - looks like link sharing is not available in prod for mobile view so I guess we all missed that!). 



- "Forget browser" modal (mobile): The cancel button is within the modal but the top corners are rounded and the bottom is straight. 

**Issues that also occur in prod**

Some issues are already present in prod so they are probably out of the scope here but I wanted to mention them first and then we can decide whether to create separate issues:

- Some modals have the cancel / close button inside the actual modal. Others it is a secondary bar fixed to the bottom of the screen. Lots of inconsistencies. 

- Modals that have a cancel button outside the modal have rounded corners on the top of the modal but straight edges on the bottom. This is the same in prod but now the distance between modal and cancel bar is larger the straight edges look more like a bug than intentional design. 

- Rotating the phone to landscape and back to portrait whilst a modal is in view causes the zoom level to get messed up and requires the user to pinch to zoom back out. 

- "Share" modal (mobile): The placeholder text is very close to the edges of the input box, some padding would make this look nicer. 

- Tapping a dropdown or text input on a modal adjusts the zoom level of the screen and requires the user to manually zoom out to return to normal.

packages/files-ui/src/Components/Modules/FileBrowsers/ReportFileModal.tsx
RyRy79261: I'm not sure where the cancel button height is, but this exact line shows up 6 times in this PR

Tbaut: Do you know if there's a way to add a constant that is function of another constant?
Like: `constant.paddingModal : Number(constants?.mobileButtonHeight) + constants.generalUnit`?

Or if there's another way to do it? This padding Modal is a function of a mixin already, I don't think you can use a mixin for a mixin.

Tbaut:
Thanks for the thorough review. Many issues here are actually also happening in prod. I'll try to fix some, but I don't want to keep this PR open forever and try to fix them all here. The potential to add some more breaking changes is high and I value small and to the point PRs rather than big catch-all.

> For modals that have a cancel button outside of the modal, when the keyboard is invoked it covers the cancel button (on production the cancel button is adjusted so that it is displayed above the keyboard).

But you can scroll to it right? I wonder if showing a cancel button all the time above the keyboard is the way to go. it takes quite some display real estate.

> "Create or edit shared folder" modal (mobile): After invoking a “Folder name is already in use” error the input bar needs to be tapped a few times before it responds and can be edited again.

Can you please confirm this doesn't happen in prob?

> Rename file modal (mobile): The modal is fixed in the middle but the background is scrollable

That's also in prod. Applying a `overflow: hidden` to the body should fix it, I'll consider it (it's not that trivial and could break yet other things).
edit: added it, I think it's doing a pretty good job.

> "Report File" modal (mobile): The decryption key is being cut off

This must happen in prod too. We have an issue related that could handle this as well https://github.com/ChainSafe/ui-monorepo/issues/1720 and https://github.com/ChainSafe/ui-monorepo/pull/1713

> Issues that also occur in prod

Thanks for finding all those. Would you mind creating issues for these. This PR breaks many things already for me to take care of everything here that's already broken :)

asnaith:
> > For modals that have a cancel button outside of the modal, when the keyboard is invoked it covers the cancel button (on production the cancel button is adjusted so that it is displayed above the keyboard).
> 
> But you can scroll to it right? I wonder if showing a cancel button all the time above the keyboard is the way to go. it takes quite some display real estate.
> 

Yeah, you can still scroll to it. I agree that we are already limited on available space and showing it limits it even more. I just wanted to mention it as it differed from prod behaviour. 

> > "Create or edit shared folder" modal (mobile): After invoking a “Folder name is already in use” error the input bar needs to be tapped a few times before it responds and can be edited again.
> 
> Can you please confirm this doesn't happen in prob?
> 

Yep, I wasn't able to reproduce this on production but I also can't reproduce now either. Sorry for the red herring, might have been an issue between localhost and my phone 

> > Rename file modal (mobile): The modal is fixed in the middle but the background is scrollable
> 
> That's also in prod. Applying a `overflow: hidden` to the body should fix it, I'll consider it (it's not that trivial and could break yet other things). edit: added it, I think it's doing a pretty good job.
> 

Nice, I'll check it out.

> > "Report File" modal (mobile): The decryption key is being cut off
> 
> This must happen in prod too. We have an issue related that could handle this as well #1720 and #1713
> 

Ah yes you're right it's in prod. The issues you mentioned are for the "File Info" modal though (but similar fix probably). I created a separate issue for the "Report File" modal #1726.

> > Issues that also occur in prod
> 
> Thanks for finding all those. Would you mind creating issues for these. This PR breaks many things already for me to take care of everything here that's already broken :)

Yep pre-existing ones are now separate issues #1727, #1728, #1729

Apologies for the myriad of things discussed at once in this PR. I just had to read it all back for my own sake/sanity 😅.  

**TL;DR** I think everything discussed here that is fixable we will address in another issue now (due to them being pre-existing). I can't recreate the weirdness with the mispositioned cursor or frozen input so they could have been edge cases that exist in production too. I'll continue to see if I can recreate those issues but they shouldn't hold up this change imo.

asnaith (approved):

FSM1 (approved):

Issue #1499, opened 2021-09-07
Single file download request has file name repeated in the path


asnaith:
**Description**
If I intercept download requests for a supported file (https://stage.imploy.site/api/v1/bucket/*/download) I see that the file name is repeated twice in the path which seems odd.

Example: 

```       
        "body": {
            "path": "/logo.png/logo.png"
        }
```

**Expected behavior**
I would expect the path to just be "/logo.png". 

**Notes**
The feature is working as intended even with the file name repeated in the path but I wanted to point this out in case it's unintentional. 

- This happens when downloading supported files from the Preview kebab menu download option
- This doesn't happen when a download request is made for unsupported files
- This doesnt happen for downloads from the File Browser kebab menu

FSM1:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Files-5eb5ea92d698464ed9543f62/issues/ChainSafe/ui-monorepo/1499?planning-poker) with ZenHub @RyRy79261 @tanmoyAtb @Tbaut

Tbaut:
I cannot reproduce this. Neither on FF nor on Brave.
Note that the `download` api call is only fired when we:
- click download from the file browser (before viewing the file)
- double-click on the file or click on the preview menu entry. Downloading from the menu in the preview doesn't fire any other call.

> This happens when downloading supported files from the Preview kebab menu download option

So this to me isn't possible. The files was already downloaded at this point, and there's no more download api call fired when you click on Download.
Also, just to see what would happen, I manually fired a request to the api with a double path, and this result in a "500 internal server error".

Let me know @asnaith if you can pin-point how this happens to you, otherwise we can close.

asnaith:
@Tbaut I've just spent 30 mins trying to reproduce it but cannot anymore whereas it was pretty consistent before. I guess we can close this

asnaith:
![File Preview -- desktop -- can see unsupported file info and download option (failed)](https://user-images.githubusercontent.com/3982665/140187338-b5fe79a6-ef60-4e90-95bb-8820bc9fb613.png)

I'm seeing instances of this happening again in CI, see the screenshot. 

This doesn't really make sense as it went away for a while and then started happening plus when I test it manually in the browser with the dev tools it looked as expected `{"path":"/file.zip"}`. 

Any ideas?

Tbaut:
I was able to reproduce when manually downloading a zip file. Will dig deeper.

[...]

PR #1717, opened 2021-11-08
fix cancel download, content not clearing, search preview content type

### LINKED ISSUE #1707: error.error is undefined when canceling a download
While debugging another issue, I came across this error:
> Uncaught (in promise) TypeError: error.error is undefined
> message: Object { dispatchConfig: null, _targetInst: null, _dispatchListeners: null, … }
in FilesContext.tsx:665:54

How to reproduce:
- try to download a single file
- cancel the download

FSM1:
In attempting to cause #1499 to occur, I fixed a couple of other small issues.
- [x] Ensure that search result content type is attempted to be guessed if not explicitly set
- [x] File download cancellation error handling (closes #1707)
- [x] Ensure that File Preview content is cleared when the effect is run


RyRy79261 (approved):

asnaith:
I wasn't too sure what you meant by the content type being guessed if not explicitly set. I tried removing some of the file extensions (eg remove .jpg) to see if it would still show an image icon and it did but it also currently does that in prod some I'm assuming it's not in relation to that? 

Also, the part about the preview window being cleared after the effect is run - which effect is that and how do I invoke that?

FSM1:
> I wasn't too sure what you meant by the content type being guessed if not explicitly set. I tried removing some of the file extensions (eg remove .jpg) to see if it would still show an image icon and it did but it also currently does that in prod some I'm assuming it's not in relation to that?

The content-type `guessing` only happens if the file's content-type field gets mangled (for instance all HEIC images that are uploaded from a windows machine have their content-type set to the generic `application/octet-stream` instead of `image/heic` which leads to the preview failing to display. In these cases we need to try and guess the content-type based on the file's extension. This had been wired up to happen for `bucket/ls` responses, but was not implemented for search results, leading to previews of certain files failing when trying to preview these from the search result list.

> Also, the part about the preview window being cleared after the effect is run - which effect is that and how do I invoke that?
This error (and the fix) can be tested in the following manner:
On staging:
1. Preview a file with a valid preview renderer
2. Preview next/previous file until one with no preview renderer is selected.
3. Attempt to download the binary of the file with no preview
4. No download is initiated, and the binary saved to disk is actually the binary of the last previewed file

After this fix, each time the user switches to a new file to be previewed, the file contents are cleared and only populated if a valid preview renderer is available. When the user downloads the file with no preview, the download request is correctly initiated and the correct binary is saved to disk.

packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -662,7 +662,11 @@ const FilesProvider = ({ children }: FilesContextProps) => {
       return Promise.resolve()
     } catch (error: any) {
       console.error(error)
-      let errorMessage = `${t`An error occurred: `} ${typeof(error) === "string" ? error : error.error.message ? error.error.message : ""}`
+      let errorMessage = `${t`An error occurred: `} ${typeof(error) === "string"
+        ? error
+        : error?.error?.message
+          ? error.error.message
+          : ""}`
```
Tbaut:
```suggestion
        : error?.error?.message || ""}`
```

packages/files-ui/src/Components/Modules/FilePreviewModal.tsx
```diff
@@ -215,8 +215,8 @@ const FilePreviewModal = ({ file, nextFile, previousFile, closePreview, filePath
       bucketId = bucket.id
     }
 
+    setFileContent(undefined)
```
Tbaut, line 218: I'm not sure if this is related, but when I browse a folder content from the file preview, using the arrows, I keep getting an error "an error occurred while generating the preview" if I don't wait for the initial file to load.

Here an example. I browse slowly (e.g I wait for the file to be previewed), all image show. Then I start clicking faster, so the file preview don't have time to load and show the page, then the error appears.

https://user-images.githubusercontent.com/33178835/141099449-05bc6cef-3ff4-418f-9d12-58af7f94df82.mp4

FSM1: Unrelated to this, and should be dealt with in a separate issue - #1725

Tbaut (approved):
Thanks, then we're good to go!

asnaith:
@FSM1 Thank you for the explanation of the change. I wasn't aware of the "guessing" type or the other issue with wrong file being downloaded so that was very insightful :)

FSM1:
> @FSM1 Thank you for the explanation of the change. I wasn't aware of the "guessing" type or the other issue with wrong file being downloaded so that was very insightful :)

Yeah the content-type guessing was a temporary work around from the early days of the project, when the API was not storing this data. Once this got resolved and the content-type was correctly being returned, I was going to remove the guessing code, but eventually decided to leave it in for weird cases like the above referenced one (HEIC images from windows).

PR #1719, opened 2021-11-09
Adding dynamic positioning to menu drop down

RyRy79261:
closes #1684

PR #1722, opened 2021-11-09
Link sharing redesign

### LINKED ISSUE #1630: Make link creation design nicer
I came up with this design, it's not ideal I guess. Let's brainstorm and find its flaws and come up with something better

Some [comments](https://github.com/ChainSafe/ui-monorepo/pull/1619#pullrequestreview-775057022) from @tanmoyAtb 
> This one is about preference, I would like to have the Create new Link button on the right side of permissions dropdown, feels more natural to me. (I haven't seen the designs though).

I guess my buttons are just too big?

https://user-images.githubusercontent.com/33178835/136352400-edbd8912-82f6-4d7d-b0fc-05c5b04e5558.mp4

Tbaut:
closes #1630 

https://user-images.githubusercontent.com/33178835/140957586-050f41bb-8875-4be0-8581-d320c405c433.mp4

### DIFF
--- a/packages/files-ui/src/Components/Elements/MnemonicForm.tsx
+++ b/packages/files-ui/src/Components/Elements/MnemonicForm.tsx
@@ -136,8 +136,10 @@ const MnemonicForm = ({ buttonLabel, onComplete }: Props) => {
               component="p"
               className={classes.loader}
             >
-              <Loading type="inherit"
-                size={16} />
+              <Loading
+                type="inherit"
+                size={16}
+              />
               <Trans>
                 Generating…
               </Trans>
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/CreateOrEditSharedFolderModal.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/CreateOrEditSharedFolderModal.tsx
@@ -13,7 +13,6 @@ import { useLookupSharedFolderUser } from "./hooks/useLookupUser"
 import { nameValidator } from "../../../Utils/validationSchema"
 import { getUserDisplayName } from "../../../Utils/getUserDisplayName"
 import LinkList from "./LinkSharing/LinkList"
-import clsx from "clsx"
 
 const useStyles = makeStyles(
   ({ breakpoints, constants, typography, zIndex, palette }: CSFTheme) => {
@@ -109,7 +108,8 @@ const useStyles = makeStyles(
         color: palette.error.main
       },
       sharingLink: {
-        padding: constants.generalUnit * 1.25
+        padding: constants.generalUnit * 2,
+        margin: 0
       }
     })
   }
@@ -204,6 +204,12 @@ const CreateOrEditSharedFolderModal = ({ mode, isModalOpen, onClose, bucketToEdi
       closePosition="none"
       maxWidth="sm"
       testId="create-or-edit-shared-folder"
+      subModal={mode === "edit" && !!bucketToEdit && (
+        <LinkList
+          bucketEncryptionKey={bucketToEdit.encryptionKey}
+          bucketId={bucketToEdit.id}
+        />
+      )}
     >
       <div className={classes.root}>
         <div className={classes.iconBacking}>
@@ -293,17 +299,6 @@ const CreateOrEditSharedFolderModal = ({ mode, isModalOpen, onClose, bucketToEdi
             data-cy="tag-edit-permission-user"
           />
         </div>
-        {mode === "edit" && !!bucketToEdit && (
-          <div className={clsx(classes.modalFlexItem, classes.sharingLink)}>
-            <Typography className={classes.inputLabel}>
-              <Trans>Sharing link</Trans>
-            </Typography>
-            <LinkList
-              bucketEncryptionKey={bucketToEdit.encryptionKey}
-              bucketId={bucketToEdit.id}
-            />
-          </div>
-        )}
         <Grid
           item
           flexDirection="row"
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/LinkSharing/LinkList.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/LinkSharing/LinkList.tsx
@@ -2,7 +2,7 @@ import { Button, Loading, MenuDropdown, Typography } from "@chainsafe/common-com
 import { createStyles, makeStyles } from "@chainsafe/common-theme"
 import { NonceResponse, NonceResponsePermission } from "@chainsafe/files-api-client"
 import { t, Trans } from "@lingui/macro"
-import React, { useCallback, useEffect, useMemo, useState } from "react"
+import React, { ReactNode, useCallback, useEffect, useMemo, useState } from "react"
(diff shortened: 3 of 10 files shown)


asnaith:
Hey, I noticed a few quirks with this one. 

- On Desktop, the drop-down options get cut off and cause a scroll bar to appear on the modal 

https://user-images.githubusercontent.com/3982665/141052058-12b4dcdf-964e-479a-addc-5d8edec6f590.mov

 - After creating a view link the view option still appears in the dropdown even though you cannot create more. The options only update if you choose the edit option, after edit is chosen the view option is not selectable. 

https://user-images.githubusercontent.com/3982665/141052188-3a956e9b-a865-42a9-be34-5b45a575e9ad.mov

- After copying the link to the clipboard the "Copied" element appears in the middle of the link. Not sure if this is intentional or if it should be displayed over the copy icon? There also doesn't seem to be any copy confirmation displayed on mobile (on a real device, it displays it on a desktop browser in mobile view). 



- The majority of link options aren't visible in the default view that you see in iOS, users may not realize there are hidden options to scroll to and in general, the experience could be more refined. It took me several attempts to create the link, the first few times when trying to click the "Create link" button the whole modal would close. See the video below. I think it took me 4 or 5 attempts.
 
https://user-images.githubusercontent.com/3982665/141054638-9bb96309-b7bf-4006-bc2d-b39c19ff9258.MP4

- I experienced the UI freezing a few times whilst using the feature on mobile and I saw this error below



- This one is more subjective but I think it's possible the end-user might not realize immediately that the "view" and "edit" elements next to the link correspond to user access rights, I think people may try to click them as buttons.

RyRy79261:
Would it make better use of the space to have the links module on the right instead of bellow?

Tbaut:
@RyRy79261 
> Would it make better use of the space to have the links module on the right instead of bellow?

We kind of tried to stay close to what GG drive does :man_shrugging:
Maybe the first screen is very big (needlessly) to leave some space for many names. I guess we could reduce that as well. In any case the side by side wouldn't work for mobile.

@asnaith 
> On Desktop, the drop-down options get cut off and cause a scroll bar to appear on the modal

That's a dropdown issue addressed here: https://github.com/ChainSafe/ui-monorepo/pull/1719

> After creating a view link the view option still appears in the dropdown.

I noticed this as well (it's visible in my video). I'll see if I can force the dropdown to re-render. 

> [...] the "Copied" element appears in the middle of the link. Not sure if this is intentional 

It is, there's a little arrow too, to convey the idea that it's pointed at a certain link.

> There also doesn't seem to be any copy confirmation displayed on mobile (on a real device, it displays it on a desktop browser in mobile view).

Can confirm on android. Not only is there no confirmation, the link is also not copied. Pretty sure this is also the reason for the UI freeze. Note that I copied this code from another place, so it means we have the problem elsewhere as well.

> I think people may try to click them as buttons.

Any suggestion? The design is showing a dropdown although it can't be changed, so I came up with this label. Maybe @serenaho has a good suggestion.

> The majority of link options aren't visible in the default view that you see in iOS. [...] I think it took me 4 or 5 attempts.

Yup this sounds like a non ideal scenario. The fact that things are dynamic, and that the modal gets resized makes it annoying on mobile. Pretty sure things wouldn't look as slick if we reserved the space though. Maybe I can reserve the space only on mobile, will take a look.

serenaho:
> I think people may try to click them as buttons.

@Tbaut @asnaith I think a quick fix for this may be to:
1. Remove the dark grey background so it cannot be mistaken for a button
2. Make the text regular-weight (right-aligned) so it's consistent with passive copy 

Here's what I was thinking:
image

A few notes:
1. The copy icon looks like it needs to be scaled a little smaller and change color to gray-8 (#595959)
2. It would help to leave some space between the URL preview copy and the "view-only" / "can edit" text

Question: 
Will users be able to physically copy the URL address without using the "copy" icon button? Wondering because otherwise it may be redundant to show such a long preview (something we can fix in later updates).

asnaith:
@Tbaut Ah ok, great I didn't realize the drop-down fix Ryan had added in #1719 was universal.

So just to summarize for anyone taking the remainder of this on: 

- Re-render the Drop-down after creating a link (so if the user already has a link then that type is no longer displayed as an option)
- Investigate the copy button not working / freeze issue on mobile 
- Serena's suggestion for the view-only / can edit labels
- See if it's feasible to reserve space for the layout on mobile

Tbaut:
100% thanks for the recap. I've been thinking about how to reserve space with a min-height. I really don't know how to do this without it being super ugly.. @serenaho @sweetpea22 any idea very welcome

Tbaut:
Thanks Serena for the suggestions, will add them

> Will users be able to physically copy the URL address without using the "copy" icon button? Wondering because otherwise it may be redundant to show such a long preview

Nope, it's way too long. They'll never be able to copy it manually. On the design you give labels, but tbh they also don't really mean anything. I could make the preview smaller sure, or put "Edit link" , "View-only link" wdyt?

serenaho:
> Nope, it's way too long. They'll never be able to copy it manually. On the design you give labels, but tbh they also don't really mean anything. I could make the preview smaller sure, or put "Edit link" , "View-only link" wdyt?

@Tbaut Are you concerned that the labels don't communicate the viewing / editing limitations for those links? 

In terms of other options I've developed a few designs that remove or shorten the URL preview, but it's hard to visualize that you're copying a link when you don't see it.

Below is another option, but it slightly contradicts the point about reserving space (and requires redesigning previous screens):

image

*Possible solutions:
One method that I observed Google drive implementing is shrinking the modal above when generating sharing links below.  I'll draft up some more options and get feedback on the Design meeting on Monday!

Tbaut:
I hope to have addressed all of the above beside the jumping issue, which is inherent to our design, kind of.
Note that:
- any copy button in Files doesn't work now on mobile for me, I've implemented something that does work for this button specifically, but is not ideal. I created an issue to properly research this, and change it across the board. https://github.com/ChainSafe/ui-monorepo/issues/1733 
- The font-weight `regular` didn't yield any difference for the label, pretty sure the font doesn't support it, or maybe it's only on my machine?

Here's how it looks
image

image

Tbaut:
@serenaho 
> Are you concerned that the labels don't communicate the viewing / editing limitations for those links

Sorry this is a misunderstanding, I'm totally fine with your first suggestion. Just I thought that it was better to at least show a bit of the links, Rather than "Link 1"

tanmoyAtb (approved):
Code and implementation both look good to me!
The jumping on mobile and modal designs need some work with time.

serenaho:
> Sorry this is a misunderstanding, I'm totally fine with your first suggestion. Just I thought that it was better to at least show a bit of the links, Rather than "Link 1"

Ahh thanks for the clarification. Got it-- sorry I didn't bother changing the copy for Link 1 and Link 2 to actual URL preview, my apologies for the confusion!

asnaith (approved):
Great updates @Tbaut! 😎
Re-rendering for the access rights options, link copy button, and link labels are all perfeito

FSM1 (approved):
working fantastically here 🎉

PR #1731, opened 2021-11-11
Fixed placeholder spacing

### LINKED ISSUE #1728: "Create Shared Folder" placeholder text has no padding (mobile)
**Description**
The "Create Shared Folder" placeholder text is very close to the edge of the input element it is within and would be neater with some extra padding. 

**Screenshots**

IMG_7EAFF2C796FB-1

RyRy79261:
closes #1728

### DIFF
--- a/packages/common-theme/src/Provider/ThemeSwitcherContext.tsx
+++ b/packages/common-theme/src/Provider/ThemeSwitcherContext.tsx
@@ -10,6 +10,8 @@ import { useLocalStorage } from "@chainsafe/browser-storage-hooks"
 
 type ThemeSwitcherContext = {
   desktop: boolean
+  tablet: boolean
+  mobile: boolean
   themeKey: string
   availableThemes: string[]
   setTheme(themeName: string): void
@@ -26,6 +28,8 @@ type ThemeSwitcherProps = {
 const ThemeSwitcher = ({ children, themes, storageKey = "cs.themeKey" }: ThemeSwitcherProps) => {
   const breakpoints = createBreakpoints({})
   const desktop = useMediaQuery(breakpoints.up("md"))
+  const tablet = useMediaQuery(breakpoints.up("sm"))
+  const mobile = useMediaQuery(breakpoints.up("xs"))
   const { canUseLocalStorage, localStorageGet, localStorageSet } = useLocalStorage()
 
   // TODO: check min 1 theme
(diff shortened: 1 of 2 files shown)


FSM1 (approved):

packages/common-theme/src/Provider/ThemeSwitcherContext.tsx
```diff
@@ -26,6 +28,8 @@ type ThemeSwitcherProps = {
 const ThemeSwitcher = ({ children, themes, storageKey = "cs.themeKey" }: ThemeSwitcherProps) => {
   const breakpoints = createBreakpoints({})
   const desktop = useMediaQuery(breakpoints.up("md"))
+  const tablet = useMediaQuery(breakpoints.up("sm"))
```
tanmoyAtb, line 31: question: with this implementation show the `tablet` variable as `true` in desktop screens ?
will this be better:
`const tablet = useMediaQuery(breakpoints.up("sm")) && useMediaQuery(breakpoints.down("md"))`

tanmoyAtb (commented):
Leaving a comment about the `tablet` and `mobile` variables

packages/common-theme/src/Provider/ThemeSwitcherContext.tsx
RyRy79261: Yeah so because desktop is an upward query, technically then we'd need to check if its between `xl` as well, though I'm not sure how many things we'd need to change to maintain that consistency

FSM1: This is basically the equivalent of the `between` operator in MUI

asnaith (approved):

packages/common-theme/src/Provider/ThemeSwitcherContext.tsx
tanmoyAtb: yes, I would say keep the desktop as is, above `md`, 
keep the tablet and mobile between screen sizes !

RyRy79261: So personally, I would also prefer only the two concepts to query against, but since this is in the theme kit, I would want it to be a bit more consistent with what one would expect from the kit, so just the Desktop & Tablet, I'd have it provided within the consuming codebase

tanmoyAtb (approved):

Issue #1729, opened 2021-11-11
Inconsistent style on modals (mobile) 


asnaith:
**Describe the bug**
I noticed that we have a variety of different modal styles in the mobile view. Sometimes we use rounded corners, sometimes flat edges. Sometimes a separate section for the cancel button and sometimes it's within the modal itself

**Screenshots**

Examples

<img width="1946" alt="Screen Shot 2021-11-10 at 4 12 56 PM" src="https://user-images.githubusercontent.com/3982665/141214710-9692d5ee-b758-40c6-a0f4-e6e74d27fb3a.png">

**Smartphone (please complete the following information):**
 - Device: iPhone 12 Mini
 - OS: iOS 15
 - Browser Safari

asnaith:
@serenaho I logged this as I thought we might want to aim for some consistency across all modals but I also may be unaware of other plans in design etc. Do you have any advice on what would be best here?

serenaho:
In terms of plans, Cindy and I will be auditing all Files components (very soon, ~next week!) in order to build a consistent design system. We will definitely be aiming for consistency across Files, especially with buttons and CTAs like this one. Thank you for bringing it up, I'll put this on my list!

RyRy79261:
@serenaho Just need some clarity, what would the preferred appearance be here? :)

serenaho:
> <img alt="Screen Shot 2021-11-10 at 4 12 56 PM" width="1946" src="https://user-images.githubusercontent.com/3982665/141214710-9692d5ee-b758-40c6-a0f4-e6e74d27fb3a.png">
> 



@RyRy79261 Let's use the same Cancel button in the left + middle screens to apply to the right modal for Create Folder.

To keep things consistent, I'd also remove the rounded edges from the middle + right modals. Thank you!

PR #1736, opened 2021-11-12
Share files from browser support

### LINKED ISSUE #1631: Allow link creation when sharing files
One should be able to create a sharing link to share a shared folder.

The current design in Figma [here](https://www.figma.com/file/wRMGqc11iQLVpDVmqXVXWK/ChainSafe-Files?node-id=4397%3A90289).

The first step would be to either select an existing folder, or to create a new shared folder. The sharing method would be the second step only in the case that we create a shared folder. 

So here are the flows:
- step1: Select either 1) an existing shared folder 2) create a new shared folder and give it a name
  if 1) then a click on "Share" would start the DL/encryption/UL
  if 2) then go to next step
- step2: We launch the DL/encryption/UL in the background and allow users to manage access with name/sharing link.

image

tanmoyAtb:
closes #1631


asnaith:
Looking good for the most part on desktop and mobile. 

I found a small issue related to the refresh logic though and I think we need to trigger a refresh after a shared folder is deleted to ensure this correct modal is always displayed.

As demonstrated in the video below - after deleting all the shared folders the wrong modal was presented when trying to share again and it didn't update until the browser was refreshed. 

https://user-images.githubusercontent.com/3982665/142082112-b7f63f51-6c97-4575-bb74-b551c1683982.mov

FSM1:
> I found a small issue related to the refresh logic though and I think we need to trigger a refresh after a shared folder is deleted to ensure this correct modal is always displayed.

This is caused by the bucket only being  marked as `status: deleting` after the request. We filter for these on the buckets overview, but not so in the `ShareModal.tsx` where `bucketOptions` is calculated. Just need to filter out any buckets with a `status: deleting` property.

packages/files-ui/src/Components/Modules/FileBrowsers/ShareModal.tsx
```diff
@@ -121,82 +91,70 @@ const useStyles = makeStyles(
...
   const [destinationBucket, setDestinationBucket] = useState<BucketKeyPermission | undefined>()
+  const [isFolderCreationLoading, setIsFolderCreationLoading] = useState(false)
   const { buckets, transferFileBetweenBuckets } = useFiles()
   const { bucket, currentPath } = useFileBrowser()
   const { profile } = useUser()
   const [nameError, setNameError] = useState("")
   const inSharedBucket = useMemo(() => bucket?.type === "share", [bucket])
+
+  const classes = useStyles({
+    width: bucketToUpload ? 600 : 500
```
FSM1, line 155: Any particular reason for this difference in width?

FSM1 (changes requested):
image

Don't know if we want to deal with this in this ticket, but the modal covers the `cancel` button at the bottom of the screen now. We should also maybe consider renaming the button to `Close` as nothing is really being cancelled by closing the modal.

tanmoyAtb:
> Looking good for the most part on desktop and mobile.
> 
> I found a small issue related to the refresh logic though and I think we need to trigger a refresh after a shared folder is deleted to ensure this correct modal is always displayed.
> 
> As demonstrated in the video below - after deleting all the shared folders the wrong modal was presented when trying to share again and it didn't update until the browser was refreshed.
> 
>  Screen.Recording.2021-11-16.at.3.17.26.PM.mov

@FSM1 thanks for the heads up, that status was it.
@asnaith should be good now. your observation is amazing !

packages/files-ui/src/Components/Modules/FileBrowsers/ShareModal.tsx
tanmoyAtb: Yeah so, The designs have this screen around 500px, however the manage shared folder modal (from shares overview) had this screen a bit larger. I wanted to keep the consistency.

tanmoyAtb:
> image
> 
> Don't know if we want to deal with this in this ticket, but the modal covers the `cancel` button at the bottom of the screen now. We should also maybe consider renaming the button to `Close` as nothing is really being cancelled by closing the modal.

`Close` sounds more appropriate. 
As for the `close` button, I think we cover that in the modal redesign issue. #1729

asnaith (approved):
Nice fix @tanmoyAtb, looking good now after shared folder deletion, the right modal is shown.

FSM1 (approved):

PR #1737, opened 2021-11-12
User lookup consent 

### LINKED ISSUE #1601: Allow users to control whether they can be looked up by wallet address
The account details screen should have a checkbox to allow a user to enable the `lookup by username/wallet address` feature for their account.

FSM1:
closes #1601 

~~There is currently a known issue at the API, which does not allow for turning the flag to `false` but this is being investigated by the API team.~~

API Issue has now been resolved

### DIFF
--- a/packages/common-theme/src/Hooks/useLongPress.tsx
+++ b/packages/common-theme/src/Hooks/useLongPress.tsx
@@ -86,8 +86,7 @@ export const useLongPress = (
           }
         }
       }
-
-    }, [onLongPress, delay]
+    }, [clear]
   )
 
   return {
--- a/packages/files-ui/package.json
+++ b/packages/files-ui/package.json
@@ -6,7 +6,7 @@
     "@babel/core": "^7.12.10",
     "@babel/runtime": "^7.0.0",
     "@chainsafe/browser-storage-hooks": "^1.0.1",
-    "@chainsafe/files-api-client": "^1.18.19",
+    "@chainsafe/files-api-client": "^1.18.20",
     "@chainsafe/web3-context": "1.1.4",
     "@lingui/core": "^3.7.2",
     "@lingui/react": "^3.7.2",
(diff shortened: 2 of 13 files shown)


RyRy79261 (approved):

asnaith:
This is working well, we just need to ensure we remember to test the ability to uncheck to remove yourself from lookup once the API issue is resolved. 

To test the scenario when it's not enabled I used a brand new account. Whilst doing this I also noticed that until lookup via share key does not work until consent is checked. I'm not sure if this is intended? Only wallet address and username are mentioned on the label. 

I'll add the below scenarios to our documentation. 

- User can be looked up via wallet address when consent is enabled (no username)
- User can be looked up via wallet address when consent is enabled (with username)
- User can be looked up via username when consent is enabled
- User cannot be looked up via username when consent is not enabled
- User cannot be looked up via wallet address when consent is not enabled

...and I'll add a scenario for the expected share key too once expected behaviour is confirmed :)

asnaith (approved):

FSM1:
>This is working well, we just need to ensure we remember to test the ability to uncheck to remove yourself from lookup once the API issue is resolved.

Tested this now and all is resolved.

> Whilst doing this I also noticed that until lookup via share key does not work until consent is checked. Only wallet address and username are mentioned on the label.

@ohmpatel1997 @dhyaniarun1993 Is this the correct behavior of the user lookup endpoint? If so, we should update the label of the checkbox.

ohmpatel1997:
>Is this the correct behavior of the user lookup endpoint? If so, we should update the label of the checkbox.

I think thats expected, since the flag should work like a global permission to make themselves searchable. do we want to change this?

FSM1:
> > Is this the correct behavior of the user lookup endpoint? If so, we should update the label of the checkbox.
> 
> I think that's expected, since the flag should work like a global permission to make themselves searchable. do we want to change this?

No need, I will just update the label to more accurately reflect what the checkbox does. Just wanted to make sure that this was intended behavior. There were a lot of discussions around whether lookup by the User Pub Key were to be allowed even if the lookup flag was set to false.

ohmpatel1997:
@kalambet  @dhyaniarun1993  yours will be the final call, what do you think?

dhyaniarun1993:
> @kalambet @dhyaniarun1993 yours will be the final call, what do you think?

Yep. I agree there should be only 1 flag that decide if user is searchable or not

PR #1759, opened 2021-11-18
Revert Clipboard changes

### LINKED ISSUE #1733: Copying anything to the clipboard doesn't work on mobile
the `navigator.clipboard.writeText` is not supported by the latest chrome mobile on the latest android, and I think any safari mobile either.
An alternative (as implemented for testing in https://github.com/ChainSafe/ui-monorepo/pull/1722 is to use `document.execCommand("copy")` but it's deprecated! Still this is working cross-platform.

In any case, we should write a hook for this and reuse it everywhere we need to copy something.

FSM1:
Since none of us (Ryan, Andrew, myself) are able to replicate the issue reported, will revert the changes made previously. If the error comes back the original issue can be reopened.

This reverts commit 424f4278a558c64d6d9317693aa40a4ee477dd03.

Closes #1733


tanmoyAtb (approved):
Tested on my phone and laptop, works fine!

asnaith:
👍 Share key copy, share link copy etc still working on Safari + iOS after these changes

asnaith (approved):

PR #1766, opened 2021-11-19
Report modal styles and cut off fix

### LINKED ISSUE #1726: Decryption key on "Report File" modal is cut off
**Description**

Report A File

**Notes**

Seems like the required changes could be very similar to the improvements made in #1720 and #1713 for the "File Info" modal

tanmoyAtb:
closes #1726


asnaith:
I'm seeing it truncating instead of adding a scroll now and the new copy button works well on mobile. 👍

The additional copy icon button at the end of the string seems a little redundant with a copy info button already in place. Probably only one would suffice? 

Also noticed we have two different styles being used on the "copied" text. I think the one with black background is the standard? 

IMG_3131

RyRy79261 (approved):

tanmoyAtb:
@asnaith Alright, so I've removed the extra copy icon button. I felt that was redundant too.

asnaith (approved):

FSM1 (approved):

Issue #1758, opened 2021-11-17
File Browser Mobile - Bulk actions toolbar optimization


FSM1:
**Is your feature request related to a problem? Please describe.**
The spacing for bulk action buttons currently takes up a lot of screen real estate when inactive, while also being insufficient in width when multiple items have been selected and there are 4 menu options available

**Describe the solution you'd like**
It should be possible to better utilize the screen real estate to display more content, while still allowing the user easy access to the functionality currently accessible via the buttons.

The Table header (which controls sorting when no items is selected) should be used as a label for the number of items selected, and menu options for the menu should be derived from the selected items, as they are for the bulk action buttons presently.

Vertical padding between the page header / survey banner and table header should be decreased as appropriate, to maximize efficient space usage.

![image](https://user-images.githubusercontent.com/12774278/143310665-c51d64e9-d9d7-4b48-a8ec-9b441ee10635.png)
![image](https://user-images.githubusercontent.com/12774278/143310691-14b77a0d-adfb-486b-b6ed-6614a3cd216d.png)
![image](https://user-images.githubusercontent.com/12774278/143310713-d1fd464b-5c8a-4136-800d-d1fed05ec059.png)

FSM1:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Files-5eb5ea92d698464ed9543f62/issues/ChainSafe/ui-monorepo/1758?planning-poker) with ZenHub @RyRy79261 @tanmoyAtb

PR #1786, opened 2021-11-29
add sorting header for mobile file browser

### LINKED ISSUE #1672: No name header for table(s) (mobile)
Mobile version of Files does not have name header for File Browser tables (or ability to sort content using alphabetical order like the desktop version).

Suggested Solutions:
Add name header for consistency and implement alphabetical sort like below
image

When no items are selected (in the file browser), the 3-dot menu should allow the user to switch between various sorting options:
image

The selected sorting option and direction should be displayed in the header of the first column.

FSM1:
closes #1672 

---

Submission checklist: 

> Remove anything below that is not applicable 

- [x] Functionality
- - [x] Feature works as intended after change

- [x] Layout
- - [x] Change looks good in the mobile web ui

- [x] Theme
- - [x] Components / elements inspected in light mode 
- - [x] Components / elements inspected in dark mode


packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -21,7 +21,8 @@ import {
   GridIcon,
   TableIcon,
   UploadSvg,
-  PlusCircleSvg
+  PlusCircleSvg,
+  MoreIcon,
```
Tbaut: Lin is complaining
```suggestion
  MoreIcon
```

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -928,6 +937,58 @@ const FilesList = ({ isShared = false }: Props) => {
...
+                    <TableHeadCell>
+                      {/* Checkbox */}
+                    </TableHeadCell>
+                    <TableHeadCell
+                      sortButtons
+                      align='left'
+                      onSortChange={toggleSortDirection}
+                      sortDirection={direction}
+                    >
+                      {column === "name" ? "Name" : column === "date_uploaded" ? "Date uploaded" : "Size"}
```
Tbaut: This is not translated

Tbaut (commented):
I wonder how usual it is to have such a "more" menu on a header, maybe we should have radio buttons in the menu to convey the idea of selecting what to show?

Also I'm not sure if that's how you intended to do the feature, but having the sort by size or date without seeing the size/date was really weird to me and looked like a bug, because we keep seeing the names, but the headers say "Size" or "Date uploaded".


https://user-images.githubusercontent.com/33178835/144042120-b2d324c4-0e82-4940-90ab-37c7713cabab.mp4

FSM1:
> I wonder how usual it is to have such a "more" menu on a header, maybe we should have radio buttons in the menu to convey the idea of selecting what to show?

Not 100% sure what you mean by radio buttons here?

> Also I'm not sure if that's how you intended to do the feature, but having the sort by size or date without seeing the size/date was really weird to me and looked like a bug, because we keep seeing the names, but the headers say "Size" or "Date uploaded".
> 
>  sorting.mp4

This header area is more of a contextual area than a column description. The changes should also be viewed in the context of the designs for https://github.com/chainsafe/ui-monorepo/issues/1758

image

Tbaut:
Understood, what I mean is that having something like this looks like a bug to me, if I were a normal user, bc the header doesn't match with what I see:

image


We can forget about the radio button thing.

FSM1:
> Understood, what I mean is that having something like this looks like a bug to me, if I were a normal user, bc the header doesn't match with what I see:
> 
> image
> 
> We can forget about the radio button thing.

I disagree that this would be perceived as a bug, since a user would manually have to change the sorting, and would by then have associated the label's value with their selected sorting parameter.

Would adding an icon or something like `Sorted by: {column}` as a label solve the lack of clarity?
image
image

Another alternative is to have the param selected dropdown menu appear by clicking on the column, rather than the 3-dot menu next to it?

@sweetpea22 @serenaho any input?

Tbaut:
Ah yes, the icon change would be great (I like the first one more). I still see the header not actually showing what's sorted as an issue. I'd keep the header as "Name" then, let's see what the other think.

asnaith:
A unique sorting icon sounds like a good idea to me

serenaho:
Hi all, highly appreciate the feedback for this mobile header.

@Tbaut I agree that there should be some sort of visual context for what date or size is being sorted.

**Issue #1:** The issue with adding an icon is that it would make the UI less consistent since the desktop version uses small arrows to sort name, date, and size like so:

image


**Issue #2:** Adding an icon on the left (A) would make it look like it's functionally similar to the icons below but instead is a button (while the others do nothing). An icon on the right (B) does not allow you to pick what you would like to sort by (Name, Size, Date).

image


**Proposed Solution:** I would propose to add right-aligned grey text beside the menu, if this is possible:

image

Please let me know your thoughts! @asnaith @FSM1

Tbaut:
Thanks @serenaho. I think with your idea and 2 columns, we could keep the arrow to sort by what we want, this is something widely used. And we could have a menu button to select what additional column to show. We would show the "Name" header all the time, above the names, and then either the date or the size. Every header would have the arrows to sort. Let me know if that's not clear I can do an ugly mock. 

Repeating what I think is weird:
- Having a "Size"  or "Date" header with Names below. This is confusing. 
- having the same 3 dots buttons on headers and file row, although the menu it totally different.

FSM1:
> **Issue #2:** Adding an icon on the left (A) would make it look like it's functionally similar to the icons below but instead is a button (while the others do nothing). An icon on the right (B) does not allow you to pick what you would like to sort by (Name, Size, Date).
> 
> image

I prefer `B` from these 2. The messaging is very clear, with the description displayed. This solved the issue pointed out by @Tbaut 
> * having the same 3 dots buttons on headers and file row, although the menu it totally different.

> 
> **Proposed Solution:** I would propose to add right-aligned grey text beside the menu, if this is possible:
> 
> image

This increases the technical complexity immensely since the secondary value selector (property name) would now need to be passed in though the stack of FileSystemItem > FileSystemTableItem components. This also greatly increases the information density of the page unnecessarily.

See examples from Google Drive with Sorting by Name:
image

Google Drive with Sorting by Size:
image

asnaith:
> Please let me know your thoughts! @asnaith @FSM1

@serenaho I think Mike, Thibaut and yourself have better insights and opinions than me with regards to this but I do like version **B** and think a different icon to access the sort types (rather than the kebab menu) would be good. 

For sorting/filtering I have often seen "doner menu" icons (sometimes called "strawberry menu") used on mobile apps and the one used on the mocks is almost like one but left aligned rather than centred. Happy to defer to whatever you deem best as this is your area of expertise 🙂

serenaho:
@FSM1 Thanks for attaching some ideas for reference. I will further experiment with option B and let you know what might work.

In the meantime, is there any way that we can display information such as date uploaded and file size without increasing technical complexity? Another way that google drive displays this info is as a subheader (below).

Screenshot_20211202-162033

Tbaut:
I think adding a sub-line is elegant, but is prob same amount of technical annoyance. I think we could leave the "Name" header at all time, and if we don't have the second column because it's problematic, we could have the filter menu be more clear, something like (this is the radio button I was mentioning originally, showing which is selected basically):
image

I made this mock using the menu we use btw: https://mui.com/components/menus/#dense-menu

FSM1:
> I think adding a sub-line is elegant, but is prob same amount of technical annoyance. I think we could leave the "Name" header at all time, and if we don't have the second column because it's problematic, we could have the filter menu be more clear, something like (this is the radio button I was mentioning originally, showing which is selected basically): image
> 
Agreed that the Android app version (with the secondary info) is a nice touch, but feel that we can always add this in later. This will require hoisting the sort param into a context and then reading it from there in all the various child components.

Will also update the menu as per @Tbaut example.

serenaho:
Hmm, @Tbaut I'm wondering where this menu would appear? Since we are replacing the three-dot menu in the header with an icon.

FSM1:
> Hmm, @Tbaut I'm wondering where this menu would appear? Since we are replacing the three-dot menu in the header with an icon.

The menu would appear when the sort button icon is clicked (if no items are selected).

If there are selected items, then the Sort menu would be replaced by the 3-dot menu, with bulk file actions.

FSM1:
One thing missing from the current designs is an indication of the sort direction (ascending/descending). Not too sure where we would fit this in @serenaho @Tbaut

serenaho:
In that case, I would say that this is a menu with a different set of actions and should have a more consistent approach with the rest of the UI and how we use functional icons (similar to the "+" icon for uploading on mobile).

https://user-images.githubusercontent.com/91699405/144651528-bbbe08e5-2a84-4836-b3b9-4db5960792b2.mov

serenaho:
> One thing missing from the current designs is an indication of the sort direction (ascending/descending). Not too sure where we would fit this in @serenaho @Tbaut

The arrow next to the "Name" header should indicate sorting direction and be interacted with like so:

https://user-images.githubusercontent.com/91699405/144652242-ae9ca782-13c6-4328-be5e-b6e2c191c034.mov

Tbaut:
@serenaho 
> The arrow next to the "Name" header should indicate sorting direction

The problem here, is that if you only have the name displayed, and the arrow for the sort direction, but want to sort by size (but without seeing the sizes) this is not going to be comprehensible. I feel like this is not something we can solve without showing both the name and the other info we want to sort by.

FSM1:
> @serenaho
> 
> > The arrow next to the "Name" header should indicate sorting direction
> 
> The problem here, is that if you only have the name displayed, and the arrow for the sort direction, but want to sort by size (but without seeing the sizes) this is not going to be comprehensible. I feel like this is not something we can solve without showing both the name and the other info we want to sort by.

By default, the sorting arrow renders on the far right of the cell, and I feel that this actually makes things even clearer than the designs

https://user-images.githubusercontent.com/12774278/144837613-2b1e58f3-17c2-4192-a363-7377f31cba5b.mp4

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -945,41 +951,52 @@ const FilesList = ({ isShared = false }: Props) => {
...
+                      {t`Name`}
                     </TableHeadCell>
                     <TableHeadCell align='right'>
                       <Menu
                         testId='fileDropdown'
-                        icon={<MoreIcon className={classes.dropdownIcon} />}
+                        icon={<SortIcon className={classes.dropdownIcon} />}
                         options={[{
+                          contents: (
+                            <ListItemText inset>
```
Tbaut, line 963: Do we really need to use the material imports here if we want to get rid of the dependency in Files ultimately?

packages/files-ui/src/UI-components/Menu.tsx
```diff
@@ -68,20 +72,24 @@ export default function Menu({ icon, options, style, testId }: Props) {
...
-            }}
-            focusVisibleClassName={clsx(style?.focusVisible)}
-            className={classes.options}
-          >
-            {option.contents}
-          </MenuItem>
-        ))}
+        <List>
+          {options.map((option, index) => (
+            <ListItem
```
Tbaut: Not sure what this adds, but the example [from mui](https://mui.com/components/menus/#dense-menu) does not use `ListItem`, but rather `MenuItem`

Tbaut (commented):
I believe I'll have to let go my first grief about the fact that sorting by something you can't see doesn't make sense to me :(

My take is that ppl will be confused but not say anything because it's a small annoyance and only on mobile. Still I'm not happy with that and hope that a user will raise the issue eventually.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
FSM1: It saves us from having to reimplement the `inset` prop. This would just be an update of where the component is imported from, once we actually move forward with the upgrade.

packages/files-ui/src/UI-components/Menu.tsx
FSM1: The example you referenced is for MUI v5. See here for documentation: https://v4.mui.com/components/menus/

List and MenuList are nearly identical, with MenuList just providing some accessibility benefits.

FSM1: Either case, I took a look at the markup generated, and Menu already handles this internally, so I have removed the `<List>` wrapper.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
Tbaut: Hehe yeah that'll probably be on me, hence my complaints :sweat_smile:

Tbaut (approved):

tanmoyAtb (approved):
Looks very functional !

PR #1792, opened 2021-11-30
Use a new random user for every test run

### LINKED ISSUE #1696: Use a unique user account per automated test session
In order to be able to run several test suites at once (either in a parallel test session or multiple PRs being validated at the same time), we need to use a unique user account (eth address) per test. This is to prevent the chance of conflict when multiple test suites are running at the same time. 

Currently, we are using the same account for all tests so it's theoretically possible one test suite could be deleting data for the account at the same time another test suite is trying to inspect the data which could lead to a false failure.

Tbaut:
closes #1696


Tbaut:
ah the survey banner fails, that makes sense, will force the test user for this one :)

RyRy79261 (approved):

FSM1 (approved):

PR #1798, opened 2021-12-02
Edit PR template

Tbaut:
Some small modifications/suggestions.
- I think the checkbox should be unchecked by default.
- I don't see the value in having a checkbox for a title, changed them for a title
- Placed the comment as comment, so that it doesn't appear in the PR
- I removed the first part because I think that it's self explanatory. This checkbox will not prevent me from making mistakes and pushing a commit that's breaking things. No-one wants to have a broken PR :man_shrugging:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,14 +6,10 @@ Submission checklist:
 
 <!-- Remove anything below that is not applicable -->   
 
-#### Functionality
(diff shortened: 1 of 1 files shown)


FSM1 (approved):

tanmoyAtb (approved):

asnaith:
Makes sense, thanks for the adjustments

PR #1803, opened 2021-12-03
Add ui test coverage for link sharing

### LINKED ISSUE #1781: Add ui test coverage for joining shares via links
As described, add ui test coverage for the joining of shares via links. 

Includes: 

- add links for all types fixtures
- write tests for asserting the joining shares via both types of links
- write test to ensure access rights change if already a view-only participant and click on a can-edit link
- write to test to ensure user can leave a folder once they have joined

asnaith:
closes #1781


packages/files-ui/cypress/fixtures/linkData.ts
```diff
@@ -0,0 +1,6 @@
...
+export const deletedFolderLink = "http://localhost:3000/link-sharing/edit/eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0eXBlIjoibGlua19zaGFyaW5nIiwicGVybWlzc2lvbiI6IndyaXRlIiwiaWF0IjoxNjM4NDAwMjg1LCJidWNrZXRfaWQiOiIyMDc5ZWQ0NC1hMmUzLTRkN2ItYWRmZC01OTUwNmVhYjk4MTciLCJub25jZV9pZCI6ImQ0NzVmODhmLWNhOGQtNDFjNC05ODU0LTk4MmQyZjAyZjllMCJ9.bT46XMPExoG66qpnG7wrwEzsbxIcwqbTyg23uBr806XamKveEtueNw74cMoTvrD0r1MiYxNnLodWjYYsjEFL1A#XwZs4tdObMmMIglBtg2o%2Bbn8xvQ14xe0cQ2HriyqTvk%3D"
+export const malformedTokenShareLink = "http://localhost:3000/link-sharing/edit/eyJhbGciOiJFUzI1NiIsInR5cCI7IkpXVCJ9.eyJ0eXBlIjoibGlua19zaGFyaW5nIiwicGVybWlzc2lvbiI6IndyaXRlIiwiaWF0IjoxNjM3ODcwNTgxLCJidWNrZXRfaWQiOiJjNWUwNGU3My01ZGM1LTQ2MDgtOTg2Mi1hNTNmM2NhM2JjODkiLCJub25jZV9pZCI6ImVlOTRiYTQwLTg5OGUtNDBkMy1iMmVhLTgwYWY1NGE2MzBiMSJ9.sM759Ys6e9RaFSq2qLRiczU4yqcqipAO1ey4frxYWRFJwCL1YF_Vct_iAlEXsPSJ4RdiKcIyIEDALQsUOQLtyQ#SXqdF0tablsTcN%2FaOvAFY9hptaIc4oIH935OBOp9Cg0%3D"
```
asnaith: I had to disable the lint rule to accomodate these long strings. Please let me know if there's a better way :)

Tbaut: there's no better way IMHO :+1:

packages/files-ui/cypress/support/page-objects/authenticationPage.ts
```diff
@@ -3,7 +3,12 @@ import { basePage } from "./basePage"
 export const authenticationPage = {
   ...basePage,
 
+  goToLoginButton: () => cy.get("[data-cy=button-go-to-login]"),
+  invalidLinkMessage: () => cy.get("[data-cy=label-invalid-link]", { timeout: 10000 }),
+  ErrorIcon: () => cy.get("[data-cy=icon-link-error", { timeout: 10000 }),
```
Tbaut:
```suggestion
  errorIcon: () => cy.get("[data-cy=icon-link-error", { timeout: 10000 }),
```

packages/files-ui/cypress/tests/link-sharing-spec.ts
```diff
@@ -73,5 +77,213 @@ describe("Link Sharing", () => {
...
+        .should("be.visible")
+        .click()
+      folderContentsPage.previewMenuOption().should("be.visible")
+      folderContentsPage.downloadMenuOption().should("be.visible")
+      folderContentsPage.infoMenuOption().should("be.visible")
+      folderContentsPage.reportMenuOption().should("be.visible")
+      folderContentsPage.copyToMenuOption().should("be.visible")
+      folderContentsPage.renameMenuOption().should("be.visible")
+      folderContentsPage.moveMenuOption().should("be.visible")
+      folderContentsPage.deleteMenuOption().should("be.visible")
```
Tbaut: These along with the view-only options are repeated many times in this file. Could we define them once and then call the set of tests from here?

Also for these ones in particular, I feel that they are not needed. We can call the edit, then the view only, and verify that we have the edit permissions. No need to test here that the view-only link actually gives the view-only permissions, this is tested before. wdyt?

packages/files-ui/src/Components/Modules/LoginModule/InitialScreen.tsx
```diff
@@ -576,16 +580,20 @@ const InitialScreen = ({ className }: IInitialScreen) => {
           <ExclamationCircleIcon
             className={classes.exclamationIcon}
             size={48}
+            data-cy="icon-link-error"
           />
           <Typography variant='h2'>
-            <Trans>This link is not valid any more.</Trans>
+            <span data-cy="label-invalid-link">
```
Tbaut: we can add the `data-cy` on the `Typography` element afair, I think it'd be better than having a span specifically for the test.

packages/files-ui/cypress/fixtures/linkData.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable max-len */
+export const viewOnlyShareLink = "http://localhost:3000/link-sharing/read/eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0eXBlIjoibGlua19zaGFyaW5nIiwicGVybWlzc2lvbiI6InJlYWQiLCJpYXQiOjE2Mzc4NzA1NzEsImJ1Y2tldF9pZCI6ImM1ZTA0ZTczLTVkYzUtNDYwOC05ODYyLWE1M2YzY2EzYmM4OSIsIm5vbmNlX2lkIjoiMzZmYjkxNzYtZDA4NS00Nzc5LTg0NTUtNTM0YjcxMzQzZjdkIn0.QE0zwQrY1kFAc_U-FsrJTcWLPMicK5Nq91RdD0szFDmg63R5YRTyBq25xJcOWlIqkAfm_WuwctzWSW5VrgYrJw#SXqdF0tablsTcN%2FaOvAFY9hptaIc4oIH935OBOp9Cg0%3D"
```
Tbaut: To be a tiny bit safer, we could use the `localHost` from the `fixtures/loginData` and have:
```suggestion
export const viewOnlyShareLink = `${localHost}/link-sharing/read/eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0eXBlIjoibGlua19zaGFyaW5nIiwicGVybWlzc2lvbiI6InJlYWQiLCJpYXQiOjE2Mzc4NzA1NzEsImJ1Y2tldF9pZCI6ImM1ZTA0ZTczLTVkYzUtNDYwOC05ODYyLWE1M2YzY2EzYmM4OSIsIm5vbmNlX2lkIjoiMzZmYjkxNzYtZDA4NS00Nzc5LTg0NTUtNTM0YjcxMzQzZjdkIn0.QE0zwQrY1kFAc_U-FsrJTcWLPMicK5Nq91RdD0szFDmg63R5YRTyBq25xJcOWlIqkAfm_WuwctzWSW5VrgYrJw#SXqdF0tablsTcN%2FaOvAFY9hptaIc4oIH935OBOp9Cg0%3D`
```

In case we change this in the future..

Tbaut (commented):
This looks awesome :boom: 
Just left a couple comments, mostly nits

packages/files-ui/cypress/support/page-objects/linkSharingConfirmation.ts
```diff
@@ -0,0 +1,8 @@
+export const linkSharingConfirmation = {
+  browseButton: () => cy.get("[data-cy=button-browse-share-folder]", { timeout: 10000 }),
+  linkErrorMessage: () => cy.get("[data-cy=label-other-error-message]", { timeout: 10000 }),
+  invalidLinkMessage: () => cy.get("[data-cy=label-invalid-link]", { timeout: 10000 }),
+  editAccessConfirmationLabel: () => cy.get("[data-cy=label-share-confirmation-write-access", { timeout: 20000 }),
+  viewAccessConfirmationLabel: () => cy.get("[data-cy=label-share-confirmation-read-access", { timeout: 20000 }),
+  ErrorIcon: () => cy.get("[data-cy=icon-link-error", { timeout: 10000 })
```
Tbaut:
```suggestion
  errorIcon: () => cy.get("[data-cy=icon-link-error", { timeout: 10000 })
```

packages/files-ui/cypress/tests/link-sharing-spec.ts
```diff
@@ -73,5 +77,213 @@ describe("Link Sharing", () => {
...
+
+      // ensure the shared folder is now shown on the main share page
+      navigationMenu.sharedNavButton().click()
+      sharedPage.sharedFolderItemRow().should("have.length", 1)
+    })
+
+    it("can see an invalid link message when it has been deleted", () => {
+      cy.web3Login()
+      cy.visit(deletedShareLink)
+      linkSharingConfirmation.ErrorIcon().should("be.visible")
```
Tbaut:
```suggestion
      linkSharingConfirmation.errorIcon().should("be.visible")
```

packages/files-ui/cypress/tests/link-sharing-spec.ts
```diff
@@ -73,5 +77,213 @@ describe("Link Sharing", () => {
...
+      linkSharingConfirmation.invalidLinkMessage().should("be.visible")
+    })
+
+    it("can see an error message for a link with a malformed token", () => {
+      cy.web3Login()
+      cy.visit(malformedTokenShareLink)
+      linkSharingConfirmation.ErrorIcon().should("be.visible")
```
Tbaut:
```suggestion
      linkSharingConfirmation.errorIcon().should("be.visible")
```

packages/files-ui/cypress/tests/link-sharing-spec.ts
```diff
@@ -73,5 +77,213 @@ describe("Link Sharing", () => {
...
+      linkSharingConfirmation.linkErrorMessage().should("be.visible")
+    })
+
+    it("can see an error message for a link to a deleted shared folder", () => {
+      cy.web3Login()
+      cy.visit(deletedFolderLink)
+      linkSharingConfirmation.ErrorIcon().should("be.visible")
```
Tbaut:
```suggestion
      linkSharingConfirmation.errorIcon().should("be.visible")
```

packages/files-ui/cypress/tests/link-sharing-spec.ts
```diff
@@ -73,5 +77,213 @@ describe("Link Sharing", () => {
...
+    })
+
+    it("can see a prompt to sign in when trying to visit a share link", () => {
+      cy.visit(viewOnlyShareLink)
+      authenticationPage.signInToAccessShareLabel().should("be.visible")
+    })
+
+    it("can see an invalid link message when signed out", () => {
+      cy.visit(deletedShareLink)
+      authenticationPage.ErrorIcon().should("be.visible")
```
Tbaut:
```suggestion
      authenticationPage.errorIcon().should("be.visible")
```

packages/files-ui/cypress/fixtures/linkData.ts
asnaith: Good idea, thank you

packages/files-ui/cypress/support/page-objects/linkSharingConfirmation.ts
asnaith: Thanks for catching this mistake 👍

packages/files-ui/src/Components/Modules/LoginModule/InitialScreen.tsx
asnaith: 👍 Sure, I've updated this now.

packages/files-ui/cypress/tests/link-sharing-spec.ts
asnaith: > These along with the view-only options are repeated many times in this file. Could we define them once and then call the set of tests from here?

👍 Good call, tbh...I don't know why I didn't do that originally. I've pushed a change to consolidate it and I think makes it a lot more readable. Let me know what you think. 

> Also for these ones in particular, I feel that they are not needed. We can call the edit, then the view only, and verify that we have the edit permissions. No need to test here that the view-only link actually gives the view-only permissions, this is tested before. wdyt?

Completely agree here too, another good call out. No point in repeating already tested behavior. We already have that initial permission check covered explicitly in another test 👍

packages/files-ui/src/Components/Modules/LinkSharingModule.tsx
```diff
@@ -164,14 +167,19 @@ const LinkSharingModule = () => {
...
               >
                 { isValidNonce === false
-                  ? <Trans>This link is not valid any more.</Trans>
-                  : error
+                  ? <span data-cy="label-invalid-link">
+                    <Trans>This link is not valid any more.</Trans>
+                  </span>
+                  : <span data-cy="label-other-error-message">
+                    {error}
+                  </span>
```
asnaith, line 182: @Tbaut, @FSM1 for conditional identifiers like this one if we want to remove the span here too I'll either have to put the identifier on the `Trans` prop (does that seem strange?) or repeat the `isValidNonce === false` within the `Typography` which seems a bit messier. 

What would be best here?

Tbaut: Agreed that in this case we can keep the `span`, to keep it cleaner.

packages/files-ui/cypress/support/page-objects/sharedFolderContentsPage.ts
```diff
@@ -0,0 +1,50 @@
...
+  // helpers and convenience functions
+  hasViewOnlyPermissionOptions(elementVisible: boolean) {
+    let assertion: string
+
+    if (elementVisible == true) {
+      assertion = "be.visible"
+    }
+    else {
+      assertion = "not.exist"
+    }
```
Tbaut: I'd go with a ternary here
```suggestion
    const assertion = elementVisible
      ? "be.visible"
      : "not.exist"
```

edit: we can actually get rid of this all together if we ask for "be.visible" or "not.exist" arguments right away, see other comment.

packages/files-ui/cypress/support/page-objects/sharedFolderContentsPage.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  hasCanEditPermissionOptions(elementVisible: boolean) {
+    let assertion: string
+
+    if (elementVisible == true) {
+      assertion = "be.visible"
+    }
+    else {
+      assertion = "not.exist"
+    }
```
Tbaut:
```suggestion
    const assertion = elementVisible
      ? "be.visible"
      : "not.exist"
```

edit: we can actually get rid of this all together if we ask for "be.visible" or "not.exist" arguments right away, see other comment.

packages/files-ui/cypress/support/page-objects/sharedFolderContentsPage.ts
```diff
@@ -0,0 +1,50 @@
...
+  downloadMenuOption: () => cy.get("[data-cy=menu-download]"),
+  infoMenuOption: () => cy.get("[data-cy=menu-info]"),
+  reportMenuOption: () => cy.get("[data-cy=menu-report]"),
+  copyToMenuOption: () => cy.get("[data-cy=menu-share]"),
+  renameMenuOption: () => cy.get("[data-cy=menu-rename]"),
+  moveMenuOption: () => cy.get("[data-cy=menu-move]"),
+  deleteMenuOption: () => cy.get("[data-cy=menu-delete]"),
+
+  // helpers and convenience functions
+  hasViewOnlyPermissionOptions(elementVisible: boolean) {
```
Tbaut: A suggestion to increase the readability of these functions, we could name it and type the argument this way
`canEditPermissionOptionsShould(elementVisible: "be.visble" | "not.exist")` so that when we call it, we know right away what this argument is for `canEditPermissionOptionsShould("be.visible")`. This then lets us get rid of the `elementVisible` boolean check all together.

packages/files-ui/cypress/fixtures/linkData.ts
Tbaut: I think you forgot to implement this suggestion :)

Tbaut (commented):
A couple more comments from the last commits

packages/files-ui/cypress/fixtures/linkData.ts
asnaith: Oops! Thanks

asnaith: Updated now :)

packages/files-ui/cypress/support/page-objects/sharedFolderContentsPage.ts
asnaith: Another good suggestion, implemented :)

Tbaut (approved):
:tada:

PR #1788, opened 2021-11-30
Billing credit card flow

tanmoyAtb:
closes #1757 

API updates are now in.

Submission checklist: 

- [x] Select plan page
- [x] plan details page
- - [x] monthly
- - [x] yearly

- [x] Card or wallet page
- [x] Add card functionality
- [x] payment page
- [x] Plan change success page

- [x] Layout
- - [x] Change looks good in the desktop web ui
- - [x] Change looks good in the mobile web ui

- [x] Theme
- - [x] Components / elements inspected in light mode 
- - [x] Components / elements inspected in dark mode

PR #1804, opened 2021-12-05
Downgrade path

### LINKED ISSUE #1783: Downgrade Plan
Overall flow as described in [Figma](https://www.figma.com/file/wRMGqc11iQLVpDVmqXVXWK/ChainSafe-Files?node-id=5129%3A85273):

If user is over the limit of the desired plan (`is_update_allowed` is false) then offer users to remove files until they fit in to the target plan.
image

If the user is able to downgrade to the desired plan:
image

Tbaut:
closes #1783

@serenaho here is the current flow where there's a slide after the new plan selection in case the user downgrade and where we show them what they loose, and have to explain them all these things with warning etc: https://github.com/ChainSafe/ui-monorepo/issues/1783#issuecomment-986893369

BTW, the flow for downgrade to Free plan is super simple since we don't need any payment selection or anything.
Select Free plan > Are you sure (with a list of what they loose) > Done

image
### This next one is where we should add the warning etc..
image
image
image
image
image

---
Submission checklist: 

   

#### Layout

#### Theme


serenaho:
Thanks for the updates @Tbaut! I'm not sure if the screens you shared are finalized yet but some feedback I have is:

1. For the first modal it appears quite text-dense from the image, but if we **stick to lighter colors for copy that isn't as important** then it would help combat that (sample below). In addition, the **bottom-left link has been updated** because I've confirmed that we don't actually have a page that goes into detail about our plans (most likely a V2 thing).

image

2. We should avoid using the term "Downgrade" in both the title and button and update the black button to be "Cancel" instead of "Back". There is a mini typo that should be "lose" instead of "loose".

3. The top navigation for each modal will ideally be minimized as soon as there are >2 screens they can go back to. To make things consistent, I used the same design that Cindy had for future Teams designs. Users should be able to click on the three-dot icon to go back to any stage they've already visited in this flow (backward or forward, assuming information that they've already input can be saved no matter where they go). Order of items in the menu is from most recent (top)--> least recent (bottom).

Screen Shot 2021-12-08 at 11 17 56 AM

4. For this Confirm Plan Change screen, if storage is the only differentiating feature then there is no need to add "basic/plus/pro" plan underneath "5 GiB of storage." 

image

Everything else is looking good so far, I will keep you posted on any updates!

Tbaut:
Thanks for the comments @serenaho 
Regarding 1 and 3, this PR mostly takes what had been done before, that indeed didn't stick very thoroughly to your design (see my comments here: https://github.com/ChainSafe/ui-monorepo/pull/1788#pullrequestreview-822757988)

I personally think we should try to implement what you designed, because I think it makes a difference, however I'll create a dedicated issue for this. The previous issue where this got implemented was huge, and this one is intended at bringing together the pieces for downgrade. I'll make the 2. and 4. changes in this PR

edit: created https://github.com/ChainSafe/ui-monorepo/issues/1825 to follow up

serenaho:
> I'll create a dedicated issue for this. The previous issue where this got implemented was huge, and this one is intended at bringing together the pieces for downgrade. I'll make the 2. and 4. changes in this PR

Sounds great, thanks @Tbaut !

Tbaut:
Addressed the comments 2 and 4.

We now have a downgrade confirmation looking like this
image

The details looks like this (remove the redundant plan name)
image

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/ChangePlan/ChangePlanModal.tsx
```diff
@@ -33,29 +34,35 @@ type ChangeModalSlides = "select" |
...
   const classes = useStyles()
-  const { getAvailablePlans, changeSubscription } = useBilling()
+  const { getAvailablePlans, changeSubscription, currentSubscription } = useBilling()
   const [selectedPlan, setSelectedPlan] = useState<Product | undefined>()
   const [selectedPrice, setSelectedPrice] = useState<ProductPrice | undefined>()
   const [slide, setSlide] = useState<ChangeModalSlides>("select")
   const [plans, setPlans] = useState<Product[] | undefined>()
   const [isLoadingChangeSubscription, setIsLoadingChangeSubscription] = useState(false)
   const [isSubscriptionError, setIsSubscriptionError] = useState(false)
+  const didSelecFreePlan = useMemo(() => !!selectedPlan && getPrice(selectedPlan, "month") === 0, [selectedPlan])
```
tanmoyAtb:
```suggestion
  const didSelectFreePlan = useMemo(() => !!selectedPlan && getPrice(selectedPlan, "month") === 0, [selectedPlan])
```

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/ChangePlan/ChangePlanModal.tsx
```diff
@@ -82,65 +89,85 @@ const ChangeProductModal = ({ onClose }: IChangeProductModal) => {
...
-        plan={selectedPlan}
-        planPrice={selectedPrice}
-      />
-      }
+      )}
+      { slide === "downgradeDetails" && selectedPlan && (
+        <DowngradeDetails
+          goBack={() => {setSlide("select")}}
+          goToPlanDetails={() => setSlide("planDetails")}
+          shouldCancelPlan={didSelecFreePlan}
```
tanmoyAtb:
```suggestion
          shouldCancelPlan={didSelectFreePlan}
```

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/ChangePlan/ConfirmPlan.tsx
```diff
@@ -139,7 +145,11 @@ const ConfirmPlan = ({
         component="h4"
         className={classes.heading}
       >
-        <Trans>Confirm plan change</Trans>
+        {
+          isDowngrade
+            ? <Trans>Confirm plan change</Trans>
+            : <Trans>Confirm plan downgrade</Trans>
+        }
```
tanmoyAtb:
```suggestion
        {
          isDowngrade
            ? <Trans>Confirm plan downgrade</Trans>
            : <Trans>Confirm plan change</Trans>
        }
```

tanmoyAtb (approved):
Looks neat and ready to go 👍

PR #1824, opened 2021-12-10
Small notification refactor

Tbaut:
---
Submission checklist: 

   

#### Layout
- [ ] Change looks good in the desktop web ui
- [ ] Change looks good in the mobile web ui

#### Theme
- [ ] Components / elements inspected in light mode 
- [ ] Components / elements inspected in dark mode


packages/files-ui/src/Components/Elements/Notifications/NotificationsDropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+    },
+    icon: {
+      transition: "none"
+    }
+  })
+)
+
+export interface Notification {
+  title: string
+  createdAt: Date
```
FSM1: Rather use a number (unix timestamp) here, as it is easier to sort and work with.
```suggestion
  createdAt: number
```

FSM1 (approved):

PR #1827, opened 2021-12-14
standardize element identifier pattern in ui tests

### LINKED ISSUE #1795: Page object identifier standardization
Just a minor house-keeping task. 

We have some inconsistencies with identifiers, nothing problematic but best to follow a standard format.

Will update all identifiers with element type at the beginning of the string.

asnaith:
closes #1795

---

Some housekeeping for the element identifiers. I have been prefixing them with the element type (but not initially) and we had a bit of a mixture in use, just a small update to make them all follow the same convention.


Tbaut (approved):
Everything's passing and no logic change :100:

PR #1832, opened 2021-12-17
Add link sharing to the sharing explainer

### LINKED ISSUE #1629: Update the sharing tutorial to talk about links
Right now, the sharing tutorial only talk about adding ppl to a shared folder. We could think about adding a mention of link sharing? wdyt @sweetpea22 @serenaho ?

to be precise, I'm talking about this "tutorial"
image

Tbaut:
closes #1629

I added a test to check that the cross allows to dismiss the modal.
TODO:
- [x] Update the image to have a high res of it.

---
Submission checklist: 

   

#### Layout

#### Theme


packages/files-ui/cypress/tests/sharing-explainer-spec.ts
```diff
@@ -32,6 +33,30 @@ describe("Sharing Explainer", () => {
...
+
+      cy.web3Login()
+      navigationMenu.sharedNavButton().click()
+      sharingExplainerModal.body().should("be.visible")
+
+      // set up a spy for the POST response
+      cy.intercept("POST", "**/user/store").as("storePost").then(() => {
+
+        // dismiss the sharing explainer
+        sharingExplainerModal.closeButton().click()
```
asnaith, line 50:
```suggestion
        sharingExplainerModal.closeButton().safeClick()
```

I tend to use `safeClick` for modal closures to remove any chance of "detached from DOM" issues or clicking before an event listener has been registered etc.

asnaith:
Nice addition to the explainer, thanks for adding the new test too 🤩

I left one tiny suggestion and can approve once we get the high res asset :)

Tbaut:
Updated the PR with the right image, it's now ready to go

packages/files-ui/cypress/tests/sharing-explainer-spec.ts
Tbaut: The safeclick is really meant for buttons, as we're checking for the `enabled` state, which is not available on this cross image. I went back to the click. We can figure it out if we ever have an issue. Safeclick remains kind of a hack that I'd rather use only when needed.

asnaith: Ah ok 👍, just by looking at the test I thought it was a button. I guess to the user it is, at least functions as one, but if the element type is not actually `button` then yeah no need to use `safeClick`

asnaith (approved):

PR #1834, opened 2021-12-20
Profile and username add tests

### LINKED ISSUE #1830: Add ui test coverage for username creation / name editing
As described. Add new test coverage in the settings spec for the following scenarios:

- can add a username
- can add / edit a first and last name

tanmoyAtb:
closes #1830


tanmoyAtb:
Everything seems to be working well, But `cy.web3Login()` seems to be logging into the same account everytime its called. This is why if we run the tests second time in a session, add username fails as this user already has a username.

packages/files-ui/cypress/tests/settings-spec.ts
```diff
@@ -1,21 +1,58 @@
...
+      settingsPage.firstNameInput().type(newFirstName)
+      settingsPage.lastNameInput().type(`${newLastName}{enter}`)
+
+      profileUpdateSuccessToast.body().should("be.visible")
+      settingsPage.firstNameInput().should("have.value", newFirstName)
+      settingsPage.lastNameInput().should("have.value", newLastName)
+    })
+
+    // random username generation
+    const newUserName = Buffer.from(Math.random().toString()).toString("base64").substr(10, 12)
```
Tbaut: I wouldn't bother with a random number and just add the date here with `new Date.now()` or something similar. With seconds precision, the odds for collision is pretty low.

Tbaut (commented):
If you use the Cypress interface, relaunching with the refresh button will not refresh the session. I guess the auto-reload function when the code changes does the same, the user will be logged in, and the sessions re-used. I launched it locally to check and can attest there's a new user created, with a new username, and all was green when launched for the first time.

If you want to test this locally, you need to close the browser (chrome) and relaunch the test manually.

packages/files-ui/cypress/tests/settings-spec.ts
tanmoyAtb: Sounds like the saner approach !

packages/files-ui/cypress/tests/settings-spec.ts
```diff
@@ -1,21 +1,62 @@
...
+
+    // username from date
+    const newUserName = Date.now().toString()
+
+    it("can add a username", () => {
+      settingsPage.signOutDropdown().should("be.visible")
+      settingsPage.addUsernameButton().should("be.visible")
+      settingsPage.addUsernameButton().click()
+      settingsPage.usernameInput().should("be.visible")
+      settingsPage.usernameInput().type(`${newUserName}`)
```
Tbaut:
```suggestion
      settingsPage.usernameInput().type(newUserName)
```

Tbaut (approved):
Looks good, just a small suggestion.

asnaith:
>>Everything seems to be working well, But cy.web3Login() seems to be logging into the same account everytime its called. This is why if we run the tests second time in a session, add username fails as this user already has a username.

Ah yes, I can see how this would cause an issue when repeating the test within the same session but this is just something we do in development whilst creating the test. I wrap the `it` in a loop to run the test many times to check its reliability before checking in, I guess this is what you were trying too?

When it's running as part of the suite that gets executed against PRs we will be fine 😊

asnaith:
This is great Tanmoy, an excellent first foray into UI test writing! Thanks for this 🙏

asnaith (approved):

PR #1837, opened 2022-01-03
TestId naming consistency

Tbaut:
I noticed that we are using `testid` and `testId`. This PR makes it consistent to `testId` everywhere.


tanmoyAtb (approved):
💯

asnaith (approved):

PR #1839, opened 2022-01-03
Notifications context

### LINKED ISSUE #1818: Notifications context
Since notifications will in future be created by various components/contexts, it is necessary to have a notification context wrapper, that will expose a hook to add and remove notification items, as well as expose the notification list to be rendered.

```
type Notification  = {
    title: string
    createdAt: number //timestamp
    onClick?: () => void
}
```

Add Notification
```
const addNotification = (notification: Notification) => string
```

Remove Notification
```
const removeNotification = (id: string) => void
```

tanmoyAtb:
closes #1818

---
Submission checklist: 

#### Layout

#### Theme


packages/files-ui/src/Components/Layouts/AppHeader.tsx
```diff
@@ -204,7 +206,7 @@ const AppHeader = ({ navOpen, setNavOpen }: IAppHeader) => {
               </section>
               <section>
                 <NotificationsDropdown
-                  notifications={[]}
+                  notifications={notifications}
```
Tbaut: This doesn't need to be passed as a prop, we could use the `notifications` from the context directly in the `NotificationDropdown` component and not have any prop on this component.

packages/files-ui/src/Contexts/NotificationsContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+const NotificationsContext = React.createContext<INotificationsContext | undefined>(
+  undefined
+)
+
+const NotificationsProvider = ({ children }: NotificationsContextProps) => {
+  const [notifications, setNotifications] = useState<Notification[]>([])
+
+  const addNotification = (notification: Omit<Notification, "id">) => {
```
Tbaut: we should  use `useCallback` here to prevent useless re-rendering.

packages/files-ui/src/Contexts/NotificationsContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+    >
+      {children}
+    </NotificationsContext.Provider>
+  )
+}
+
+const useNotifications = () => {
+  const context = React.useContext(NotificationsContext)
+  if (context === undefined) {
+    throw new Error("useNotifications must be used within a BillingProvider")
```
Tbaut:
```suggestion
    throw new Error("useNotifications must be used within a NotificationsProvider")
```

packages/files-ui/src/Contexts/NotificationsContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+const NotificationsContext = React.createContext<INotificationsContext | undefined>(
+  undefined
+)
+
+const NotificationsProvider = ({ children }: NotificationsContextProps) => {
+  const [notifications, setNotifications] = useState<Notification[]>([])
+
+  const addNotification = (notification: Omit<Notification, "id">) => {
+    setNotifications([...notifications, {
+      id: (notifications.length + 1).toString(),
```
Tbaut: I think we shouldn't have the index as the id, for several reasons:
- this info is accessible easily, it's the index, no need to have an id for this
- we could have similar ids if we don't take care of this. E.g I just created a "create" and a "delete" button, with the logic that I delete the first notification in the array. If I create 2 notifications, remove 1, and create another, I end up with 2 notifications with the id `2`.

I'd go ahead and leave the `addNotification` accept a `Notification` as param, without the `Omit`, and let the logic about the `id` be handled by the notification creator. This has also the positive effect that the creator knows the notification id beforehand, and can have logic such as `onClick` removing the right notification, or marking it as read etc.

lmk if I'm not clear.

Tbaut (changes requested):

packages/files-ui/src/Components/Layouts/AppHeader.tsx
tanmoyAtb: Pulling from contexts from inside the components in `elements` folder feels odd to me. Keeping these as separate makes for better reusability IMO.

packages/files-ui/src/Contexts/NotificationsContext.tsx
tanmoyAtb: Ahh, thats a great example. I should have thought of this!!

So, About passing the `id`, I think it's better to create the `id` in the context. Making the notification creator worry about the `id` adds extra concern there. We can return the `id` from `addNotification` method. this is sth we're doing in `ToastsContext`.

To create the `id`s, we can either use `uuid` as we're doing with `toasts`, or 
we can increment the previous id.

What do you think ?

Tbaut: Agreed, then we can return the id upon creation, for things like "mark as read" or "delete" we'll need to keep this id in the state somewhere.

tanmoyAtb: Updated to work with uuid and return the id.

packages/files-ui/src/Components/Layouts/AppHeader.tsx
Tbaut: I won't fight it, but I'm not sure I agree with that. Contexts are specifically used to avoid props drilling. We have components in the `Elements` directory that pull some context state as well such as `InvoiceLines`, `MnemonicForm`. I get your point in that it eases re-usability, but I don't see the Notification dropdown being reused, it's very specific.

Tbaut (approved):

packages/files-ui/src/Components/Layouts/AppHeader.tsx
tanmoyAtb: Cool, I realise we're already using from the contexts in elements. Made the updates 👍

PR #1841, opened 2022-01-05
Fix React warning with data attributes

Tbaut:
Something I missed, thanks @tanmoyAtb: React doesn't like having camelCase for data attributes somehow.

So we'll have to live with `data-testid` all lower case, still more consistent than before.

image

### DIFF
--- a/packages/common-components/src/Button/Button.tsx
+++ b/packages/common-components/src/Button/Button.tsx
@@ -356,7 +356,7 @@ const Button: React.FC<IButtonProps> = ({
       )}
       disabled={disabled || loading}
       {...rest}
-      data-testId={`button-${testId}`}
+      data-testid={`button-${testId}`}
     >
       {loading && (
         <>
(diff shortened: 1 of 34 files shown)


packages/common-components/src/Button/Button.tsx
```diff
@@ -356,7 +356,7 @@ const Button: React.FC<IButtonProps> = ({
       )}
       disabled={disabled || loading}
       {...rest}
-      data-testId={`button-${testId}`}
+      data-testid={`button-${testId}`}
```
asnaith, line 359: Should all the `${testId}` references be updated too?

Tbaut: no I think it's ok because that's a prop in a component that doesn't end up as an html attribute. The problem from the warning is that the prop was directly translated as html, and html attributes should not have camelCase

asnaith: Ah ok! Thanks for the explanation :)

asnaith (approved):

tanmoyAtb (approved):

PR #1843, opened 2022-01-07
add ui tests for add / update / remove card

### LINKED ISSUE #1842: add new ui tests for adding a card to profile
As described, at new ui tests in cypress for adding a card to the user profile.

asnaith:
closes #1842 

- Adds new ui tests for adding, update and removing a card on the user account.


Tbaut:
All right, after looking for things on the stripe doc, I ended up finding a pretty elegant way to actually select the input fields inside an iframe, and type in it. Thanks to [this](https://medium.com/@michabahr/testing-stripe-elements-with-cypress-5a2fc17ab27b) and [this](https://medium.com/@xavitb3/thank-you-very-much-for-the-code-ff3543b49c59) articles.


https://user-images.githubusercontent.com/33178835/148565825-9c41f979-489a-4275-9d2a-3c41c1fd657e.mp4

asnaith:
>>All right, after looking for things on the stripe doc, I ended up finding a pretty elegant way to actually select the input fields inside an iframe, and type in it. Thanks to this and this articles.

Thank you so much @Tbaut, this was a great help 🙏.

I didn't have a stripe key in my .env file either which was also causing me a problem!

asnaith:
I've marked this as `it.skip` for now until we have the ability to remove cards programmatically (#1847) 

update: resolved!

packages/files-ui/cypress/support/commands.ts
```diff
@@ -143,6 +150,31 @@ Cypress.Commands.add("safeClick", { prevSubject: "element" }, ($element?: JQuery
...
+      })
+    }
+  })
+})
+
+Cypress.Commands.add("getInDocument", { prevSubject: "document" }, (document: any, selector: keyof HTMLElementTagNameMap) =>
+  Cypress.$(selector, document))
+
+Cypress.Commands.add("getWithinIframe", (targetElement: any, selector: string) =>
+  cy.get(selector || "iframe")
```
Tbaut: edit: I wrote this comment before I found the other blog post, those wait should be avoidable I believe: 

We should add the wait here, rather than hardcoding it anywhere else.
I believe doing something like this should work
```suggestion
  cy.get(selector || "iframe", { timeout: 10000 })
```

Tbaut (commented):
I found another blog post, from cypress themselves, that goes into details, and feels a little less hacky. You can take a look, but I can help for the implementation if you want.

here: https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/

packages/files-ui/cypress/support/commands.ts
asnaith: Thank you. I added it but when running repeatedly still got some issues when I removed the 1 second waits. 

I _think_ because there is a slight delay from when it becomes visible to when it becomes interactive (these iframe elements specifically).

packages/files-ui/cypress/tests/subscription-plan-spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      // add a card
+      settingsPage.addCardButton()
+        .should("be.visible")
+        .click()
+      addOrUpdateCardModal.body().should("be.visible")
+      addOrUpdateCardModal.addCardHeader().should("be.visible")
+      // wait is needed to allow time for the stripe components to be editable
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(1000)
```
Tbaut: Just to be clear, I would like to get to something robust enough to get rid of those `wait`, or at least try hard to avoid them before letting them in.

asnaith: Ah sorry, I noticed my comment from yesterday disappeared (it looked like it posted twice so I deleted one of them but removed both!). 

I tried with the iframe timeout in place and with the waits removed and I was still having issues. I think it's because there is a momentary lag between when the elements become visible and when they become interactive (stripe elements specifically). 

I'll keep experimenting.

Tbaut: yeah I'm not totally surprised, that's why I suggested the other blog post from cypress, where each step is explained. The issue is that our `Cypress.$(getter, document)` doesn't have a timeout, this is the part that will check for the intput field inside the iframe. I checked and couldn't figure a way to make this wait, so maybe we should check in details the other solution and see if it's more elegant. In the end, both end up doing the same.

asnaith: I don't think the issue was the timeout, as you'd added one and it would return a successful `"be.visible"` check but sometimes then fail when trying to type. 

It seems to me that the stripe elements return true for visibility before it is actually possible to type in them (but I'm talking in terms of < 1 second, that's how I'd avoided it with the 1 second wait).

In the dev tools, I watched all the requests that came in after opening the add card modal and by adding a spy on all the stripe requests (the `r.stripe` calls which seem to be responsible for initializing/activating the elements) and waiting for them to complete I was able to get reliable input without the hardcoded wait.  

That was a lot of seconds spent in dev to avoid the hardcoded 1 second wait 😅. I do think this is better though as if there's ever a failure in the future because of a failed request at least it will be more obvious to diagnose.

asnaith: Oh, @Tbaut I was talking about getting the stripe key in .env file in render before...and I just realized that doesn't really matter as we're not running these tests on render instances 🤦‍♂️. 

We need to get the stripe key in the .env file in the docker container running the tests. I know that can be done via `test-files.yml` but I don't have repo permissions to add GitHub secrets so can't do that at the moment. Would you be able to assist or give me the appropriate access? 

After that, I think we will be good to go on this.

packages/files-ui/cypress/support/page-objects/settingsPage.ts
```diff
@@ -2,7 +2,9 @@ import { basePage } from "./basePage"
 
 export const settingsPage = {
   ...basePage,
-  profileTabButton: () => cy.get("[data-testid=tab-profile]"),
+
+  // profile tab
+  profileTabButton: () => cy.get("[data-testId=tab-profile]"),
```
Tbaut:
```suggestion
  profileTabButton: () => cy.get("[data-testid=tab-profile]"),
```

packages/files-ui/cypress/support/page-objects/settingsPage.ts
```diff
@@ -12,7 +14,30 @@ export const settingsPage = {
   usernameErrorLabel: () => cy.get("[data-cy=input-profile-username] span.error"),
   setUsernameButton: () => cy.get("[data-cy=button-set-username]"),
   usernamePresentInput: () => cy.get("[data-cy=input-profile-username-present]"),
-  securityTabButton: () => cy.get("[data-testid=tab-security]"),
+
+  // security tab
+  securityTabButton: () => cy.get("[data-testId=tab-security]"),
```
Tbaut:
```suggestion
  securityTabButton: () => cy.get("[data-testid=tab-security]"),
```

packages/files-ui/cypress/support/page-objects/settingsPage.ts
```diff
@@ -12,7 +14,30 @@ export const settingsPage = {
...
   securityTabHeader: () => cy.get("[data-cy=label-security-header]"),
-  subscriptionTabButton: () => cy.get("[data-testid=tab-subscription]")
+
+  // subscription tab
+  subscriptionTabButton: () => cy.get("[data-testId=tab-subscription]"),
+  addCardButton: () => cy.get("[data-testid=button-add-a-card]"),
+  updateCardButton: () => cy.get("[data-testId=button-update-a-card]"),
```
Tbaut:
```suggestion
  subscriptionTabButton: () => cy.get("[data-testid=tab-subscription]"),
  addCardButton: () => cy.get("[data-testid=button-add-a-card]"),
  updateCardButton: () => cy.get("[data-testid=button-update-a-card]"),
```

Tbaut (approved):
This look robust and is now passing with the stripe key! Just found a couple nits, but it's good to go.

packages/files-ui/cypress/tests/subscription-plan-spec.ts
Tbaut: aah not sure what I was thinking of. Just added the stripe keys, although I don't think it's a secret, since it's available in the calls if you inspect them, I added it as such.

asnaith: @Tbaut Ah ok, thank you for doing that 🙏. Perhaps still best to keep it a little bit obscured even if the key is obtainable through looking at the calls.

packages/files-ui/cypress/support/page-objects/settingsPage.ts
asnaith: Thanks for catching those!

packages/files-ui/cypress/fixtures/cardData.ts
```diff
@@ -0,0 +1,6 @@
+export const visaNumber = "4242424242424242"
+export const visaCvc = "123"
+export const visaExpiry = "12/30"
+export const mastercardNumber = "5555555555554444"
```
FSM1, line 4: Don't know whether it makes much sense to try and test with a card that requires 3DS authentication - it is outside the scope of our system, but at the same time is part of the core business flow

asnaith: @FSM1 yeah it's a bit of a conundrum. My own inclination would be to not do it as it is out of our control and if it breaks it would be Stripe that need to fix it (and I'm sure they'd know when it breaks as probably the same used for all their clients). 

It seems like we aren't wrapping/styling it so currently there's no way of assigning any id's to the elements within it (like we did for the inputs in `AddCard.tsx`). Cypress cannot see it in the selector playground so I think we would need to do some extra work to wrap/expose it.

Maybe we should discuss it briefly in our planning meeting and if we decide to do it then we could create a new issue out of the context of this PR, does that sound ok?

FSM1: sounds good :)

FSM1 (approved):

PR #1836, opened 2021-12-25
DnD with breadcrumbs

tanmoyAtb:
closes #1772 

---
Submission checklist: 

#### Layout
- [x] Change looks good in the desktop web ui
- [x] Change looks good in the mobile web ui

#### Theme
- [x] Components / elements inspected in light mode 
- [x] Components / elements inspected in dark mode

PR #1849, opened 2022-01-11
Folders drag and drop

### LINKED ISSUE #959: Select and drag works for files, not for folders
Not sure if this is supposed to be that way, or if it's a bug (could be linked to https://github.com/ChainSafe/files-ui/pull/896 ?)
In any case as a user I was surprised that one type works and not the other.

ps: I will test in prod as soon as I can to tell if it's a regression or not

tanmoyAtb:
closes #959

### DIFF
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemItem.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemItem.tsx
@@ -383,7 +383,6 @@ const FileSystemItem = ({
 
   const [, dragMoveRef, preview] = useDrag({
     type: DragTypes.MOVABLE_FILE,
-    canDrag: !isFolder,
     item: () => {
       if (selectedCids.includes(file.cid)) {
         return { ids: selectedCids }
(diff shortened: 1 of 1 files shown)


asnaith:
Nice. This is working really well for the most part. 

Preventing a folder from being moved into itself is working but I did notice if a folder is already in the root directory you can still move it to home (same for DND files too but I didn't notice this before). 

https://user-images.githubusercontent.com/3982665/149205053-2c4ecbb4-a6e7-49cd-bba6-19ceac33389d.mov

FSM1 (approved):
working great :)

FSM1:
> Nice. This is working really well for the most part.
> 
> Preventing a folder from being moved into itself is working but I did notice if a folder is already in the root directory you can still move it to home (same for DND files too but I didn't notice this before).
> 
>  folder.moving.in.root.mov

@asnaith the bug you described actually relates to the work completed for #1836. I will create a separate issue to have this resolved.

asnaith (approved):
>>@asnaith the bug you described actually relates to the work completed for #1836. I will create a separate issue to have this resolved.

@FSM1 Ok, sounds good 👍

PR #1850, opened 2022-01-12
Add programmatic card deletion for tests

### LINKED ISSUE #1847: Extend apiTestHelper to programmatically remove cards
*Summary*
At the moment we have no ability to remove the card associated to a user profile prior to beginning a test. This means we cannot reliably run our "add / update / remove card" tests.

Adding this functionality will give us the ability to always start the test in a clean state no matter what happened in previous test executions.

Acceptance Critieria: 

- Add new functionality to apiTestHelper.ts / commands.ts that gives us the ability to remove credit cards on the profile programmatically
- Update `cy.web3Login()` calls in `subscription-plan-spec.ts` to include new parameter for removing any pre-existing card. 
- Remove the `skip` on the `it` for the "can add, update and remove a credit card for billing" test.

Tbaut:
closes #1847

### DIFF
--- a/packages/files-ui/cypress/support/commands.ts
+++ b/packages/files-ui/cypress/support/commands.ts
@@ -47,6 +47,7 @@ export interface Web3LoginOptions {
   clearTrashBucket?: boolean
   deleteShareBucket?: boolean
   withNewUser?: boolean
+  deleteCreditCard? : boolean
 }
 
 Cypress.Commands.add(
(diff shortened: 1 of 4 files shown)


asnaith (approved):

asnaith:
Fantastic! Thank you for this @Tbaut 🙏

Tbaut:
The test failed interestingly:
> AssertionError: Timed out retrying after 4000ms: Expected to find element: `#iframe-card-number iframe`, but never found it.

I relaunched it, I don't think it's linked to this PR though

asnaith:
> The test failed interestingly:
> 
> > AssertionError: Timed out retrying after 4000ms: Expected to find element: `#iframe-card-number iframe`, but never found it.
> 
> I relaunched it, I don't think it's linked to this PR though

@Tbaut I think this was the failure I was seeing when running locally when I didn't have the stripe key in my .env file. 

Perhaps we need to add the key to the .env that render is using?

Tbaut:
render has the stripe key in the env though, I just verified it. I'll merge this one since it's not related, and we can keep checking on the other PR.

PR #1861, opened 2022-01-18
Billing notifications

### LINKED ISSUE #1819: Billing notifications
The following conditions should trigger a notification
- [x] There is an unpaid invoice (whether failed CC payment, or an upcoming Crypto payment). The crypto invoices will be created 1 week before expiry.
- [x] The account is in restricted mode
- [x] There is an active subscription, and the default card expires before the next billing date

FSM1:
closes #1819
---
Submission checklist: 

#### Layout
- [ ] ~~Change looks good in the mobile web ui~~ - Currently N/A as the mobile notification area still needs to be designed and implemented.

#### Theme


tanmoyAtb (approved):
Looks 🚀  to go !

asnaith:
Hey Mike, should we be able to verify this by modifying the responses on `https://stage.imploy.site/api/v1/billing/subscriptions`? Meaning, if we managed to intercept and stub out the expiry date to be within 7 days should we see one of these notifications? 

That was the most logical way I could think of testing. However, I got the traffic routing through Charles Proxy and set a breakpoint on `billing/subscription`. I manually edited the expiration date to be next week but still couldn't see the notification in Files.

FSM1:
>Which of the notifications were you trying to test? If it's the first one, then the best way to test this would be to intercept >the invoices response and add in an invoice with a status of `open` in which case a notification should be displayed.

Yeah! Thank you for the tips, it's looking good

Screen Shot 2022-01-20 at 1 36 57 PM

>The account restricted one is a bit trickier as it is driven by the contents of the JWT. I will need to take a deeper look at >the code to see if this one is easily testable.

Ok 👍

>The card expiry notification is the easiest to test. Just need to add a card with an expiry date in the current month and >choose a monthly plan, and the notification should pop up. Updating the card to one expiring after the next payment date >should make the notification disappear.

Awesome!
Screen Shot 2022-01-20 at 1 47 21 PM

packages/files-ui/src/Components/Elements/InvoiceLines.tsx
```diff
@@ -69,6 +71,18 @@ const InvoiceLines = ({ lineNumber }: IInvoiceProps) => {
       : invoices
   }, [invoices, lineNumber])
 
+  const downloadInvoice = useCallback(async (invoiceId: string) => {
```
Tbaut: Should we have this logic in the BillingContext just for the sake of having any billing related logic there?

packages/files-ui/src/Components/Elements/InvoiceLines.tsx
```diff
@@ -108,11 +122,18 @@ const InvoiceLines = ({ lineNumber }: IInvoiceProps) => {
...
               >
-                {dayjs(paid_on).format("MMM D, YYYY")}
-              </Typography>
-              <Typography>
-                {uuid}
+                {dayjs.unix(period_start).format("MMM D, YYYY")}
               </Typography>
+              {(status === "paid") && (
+                <Button onClick={() => downloadInvoice(uuid)}>
+                  Download
```
Tbaut:
```suggestion
                  <Trans>Download</Trans>
```

packages/files-ui/src/Components/Elements/InvoiceLines.tsx
```diff
@@ -108,11 +122,18 @@ const InvoiceLines = ({ lineNumber }: IInvoiceProps) => {
...
+                </Button>
+              )}
+              {(status === "open") && (
+                <Button onClick={() => console.log("Not implemented")}>
+                  Pay invoice
```
Tbaut:
```suggestion
                  <Trans>Pay invoice</Trans>
```

packages/files-ui/src/Components/Elements/InvoiceLines.tsx
```diff
@@ -108,11 +122,18 @@ const InvoiceLines = ({ lineNumber }: IInvoiceProps) => {
...
-                {uuid}
+                {dayjs.unix(period_start).format("MMM D, YYYY")}
               </Typography>
+              {(status === "paid") && (
+                <Button onClick={() => downloadInvoice(uuid)}>
+                  Download
+                </Button>
+              )}
+              {(status === "open") && (
+                <Button onClick={() => console.log("Not implemented")}>
```
Tbaut, line 119: Should we have a comment there with an issue number maybe?

packages/files-ui/src/Components/Elements/InvoiceLines.tsx
FSM1: yeah that is not a bad idea at all. Let me move it across quickly

packages/files-ui/src/Contexts/BillingContext.tsx
```diff
@@ -49,32 +53,86 @@ const BillingContext = React.createContext<IBillingContext | undefined>(
...
 const BillingProvider = ({ children }: BillingContextProps) => {
-  const { filesApiClient, isLoggedIn } = useFilesApi()
+  const { filesApiClient, isLoggedIn, accountRestricted } = useFilesApi()
+  const { redirect } = useHistory()
+  const { addNotification, removeNotification } = useNotifications()
   const { refreshBuckets } = useFiles()
   const [currentSubscription, setCurrentSubscription] = useState<CurrentSubscription | undefined>()
   const [defaultCard, setDefaultCard] = useState<Card | undefined>(undefined)
   const [invoices, setInvoices] = useState<InvoiceResponse[] | undefined>()
+  const [restrictedNotification, setRestrictedNotification] = useState<string | undefined>()
```
Tbaut, line 64: I wonder if we could find a better place for this code since it's not directly related to Billing. Something a little more down the tree that would use the BillingContext and the NotificationContext.

packages/files-ui/src/Components/Elements/InvoiceLines.tsx
FSM1: already working on this...

packages/files-ui/src/Contexts/BillingContext.tsx
FSM1: The account restricted flag is very much related to billing - the only time an account goes into restricted mode is when bills are not paid. The only reason the flag is calculated at the FilesApiContext is due to the source being the access token, which I don't really want to expose anywhere further down the tree

Tbaut: I meant more all the notifications here in the BillingContext. But yeah you're right, it's def related to billing, just I couldn't convince myself that the context was the right place. But I can't come up with a better place anyway.

Tbaut:
I didn't play with it yet, but the code looks good.

asnaith (approved):

PR #1867, opened 2022-01-19
ENS lookup

### LINKED ISSUE #1848: Allow for user lookup by ENS Name
When a ENS name (eg `name.eth`) is entered into the tag input, it should be resolved to an ETH address which would be submitted to the API to look up the target user.

image

**Additional context**
ENS lookup should be performed using `ethers.provider.resolveName(ensName)`

tanmoyAtb:
closes #1848


asnaith:
I was unable to verify this as any ENS names I tried did not resolve to any wallet addresses. 

Does it only work for ENS name that is registered to a wallet that a files user has signed in with and has approved lookup for? If so perhaps we also need to update this copy too: 

Screen Shot 2022-01-19 at 7 25 24 PM

tanmoyAtb:
> I was unable to verify this as any ENS names I tried did not resolve to any wallet addresses.
> 
> Does it only work for ENS name that is registered to a wallet that a files user has signed in with and has approved lookup for? If so perhaps we also need to update this copy too:
> 
> Screen Shot 2022-01-19 at 7 25 24 PM

I couldn't verify that completely as well. We will need an ens, which uses an address that is registered on files and has loookup enabled. Thanks for bringing this up @asnaith

FSM1:
> I was unable to verify this as any ENS names I tried did not resolve to any wallet addresses.
> 
> Does it only work for ENS name that is registered to a wallet that a files user has signed in with and has approved lookup for? If so perhaps we also need to update this copy too:
> 
> Screen Shot 2022-01-19 at 7 25 24 PM

We can now test this on staging with Ryan's account.

Agree that the label for the checkbox, as well as the labels in the share dialog should be update to inform the user that they can use an ENS name.

I still have some digging to do into whether the ENS regex used is sufficient. its a bit hard to find a definitive list of allowed/disallowed characters in an ENS/DNS name, since it seems like most of unicode is supported, including emojis: 
`🔥.wallet.eth` is a valid ENS name 🤯

tanmoyAtb:
> 🔥.wallet.eth

Added the placeholders.
I updated thee regex and tested these ENS/DNS (the emoji support is just mind blowing :O)

🔥.wallet.eth
🔥😃.eth
stackoverflow.com
stackoverflow.co.uk
vitalik.eth

FSM1 (approved):

asnaith (approved):
Working nicely. I just added a tiny commit to update the placeholder on the profile tab too

PR #1868, opened 2022-01-19
add ui test for checking invalid card details

### LINKED ISSUE #1863: Add new ui tests for invalid card data
Add new test to ensure that a user cannot add a card to their profile if 
- the card number is invalid
- the expiration date is invalid
- the cvc date is invalid

asnaith:
closes #1863


Tbaut (approved):
We're testing stripe's components here rather than ours :)
But it can't hurt anyway.

PR #1876, opened 2022-01-24
[Fix] Remove the artificial 65 char limit and fix caption on black theme

### LINKED ISSUE #1874: Change # chars for uploaded files to unlimited
_[Adapted from Slack thread]_

Uploading a file with a super long name is fine until a user tries to rename it (even to a shorter name). It seems there’s a max # of chars we can input for names, but after shortening the title, the system still doesn't take the shorter renamed title (even after pressing ‘enter’).

Solution: Update # chars for uploaded files (and all relevant things) to be unlimited

@Tbaut @FSM1

Tbaut:
closes #1874 

after some tests, it seems that the limit in the name isn't enforced by the api side, for folders or shared folders (buckets) so I removed them all. While testing in dark mode I realize than we had a certainly long standing issue where the error message weren't readable, black on black. That's also fixed now

Before:
image

After
image

---
Submission checklist: 

   

#### Layout

#### Theme

### DIFF
--- a/packages/files-ui/cypress/tests/file-management-spec.ts
+++ b/packages/files-ui/cypress/tests/file-management-spec.ts
@@ -375,12 +375,11 @@ describe("File management", () => {
       createFolderModal.errorLabel().should("be.visible")
       createFolderModal.body().should("contain.text", "Please enter a name")
 
-      // ensure a folder can't be created if name exceeds 65 characters
+      // ensure a folder can be created even with a large amount of
       createFolderModal.folderNameInput().type("{selectall}{del}")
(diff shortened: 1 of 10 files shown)


packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -196,7 +196,7 @@ describe("File management", () => {
     })
 
     it("can rename a file with error handling", () => {
-      const newName = "awesome new name"
+      const newName = "awesome new name that is pretty long and it shouldn't matter that much anyway"
```
asnaith, line 199: 👍

asnaith (approved):

FSM1 (approved):

PR #1877, opened 2022-01-24
Programmacically get a user back to the free plan

### LINKED ISSUE #1865: Extend apiTestHelper to programmatically set users subscription plan
**_Summary_**

At the moment we have no ability to set a user's subscription plan level at the beginning of a test. 

Whilst the free plan is default and we use a unique user per test session we still need to put this in place because if a test fails after a user has been changed to a different plan then non of the automatic retries will work either. 

In addition to this, it would be great if we could also programmatically set a user to a standard or premium plan too. If this is possible it would help make future downgrade path tests more efficient. 

**_Acceptance Critieria:_**

- Add new functionality to apiTestHelper.ts / commands.ts that gives us the ability to set the subscription plan programmatically

Tbaut:
closes #1865

---
Submission checklist: 

   

#### Layout

#### Theme

### DIFF
--- a/packages/files-ui/cypress/support/commands.ts
+++ b/packages/files-ui/cypress/support/commands.ts
@@ -48,6 +48,7 @@ export interface Web3LoginOptions {
   deleteShareBucket?: boolean
   withNewUser?: boolean
   deleteCreditCard? : boolean
+  resetToFreePlan?: boolean
 }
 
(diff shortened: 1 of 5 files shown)


asnaith (approved):
Looks great, thanks so much for this. Will help a lot 👍

FSM1 (approved):

PR #1879, opened 2022-01-25
Fix Posthog Identify integration

FSM1:
This PR just updates the posthog js client used, as well as adds some more event logging - The underlying issue of identified users appearing as unidentified is still present and is likely caused by a bug in the posthog server, which should be resolved after this is updated.


asnaith (approved):

Tbaut (approved):

Tbaut:
Failing tests :scream_cat: re-ran them

PR #1882, opened 2022-01-26
add ui test for upgrading a plan via credit card payment

### LINKED ISSUE #1880: add new ui tests for subscription upgrade with credit card
Add new ui tests that cover: 

- choosing a plan
- adding a credit card (via select payment method, not pre-existing)
- confirming the upgrade 
- downloading the invoice.

asnaith:
closes #1880


packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/ChangePlan/ChangePlanModal.tsx
```diff
@@ -102,13 +102,7 @@ const ChangeProductModal = ({ onClose }: IChangeProductModal) => {
...
       }}
       onClose={onClose}
-      testId={
-        (slide === "select")
-          ? "select-plan"
-          : slide === "downgradeDetails"
-            ? "downgrade-details"
-            : "plan-details"
-      }
+      testId={`${ slide }`}
```
asnaith: Please let me know if anyone can think of a better thing to do here. 

I decided to do it like this as I wanted each modal in the flow to have a unique identifier and I had to add more (for select payment, confirm plan and plan success). 

The only other alternative seemed to be to have an even uglier if statement, as I seem to have discovered there's no straightforward way to use a switch in a react component (is that true?).

packages/files-ui/cypress/support/commands.ts
```diff
@@ -181,6 +180,26 @@ Cypress.Commands.add("getWithinIframe", (targetElement: any, selector: string) =
...
+})
+
+Cypress.Commands.add("awaitDefaultCardRequest", () => {
+  cy.intercept("GET", "**/billing/cards/default").as("defaultCard")
+
+  cy.wait("@defaultCard").its("response.body").should("contain", {
+    type: "credit"
+  })
+})
+
```
asnaith, line 202: Just an fyi - I decided to move these out of page objects and into cypress commands.ts for two reasons 

- the functions are used in more than one area of the app (but are not suited to being on the basePage as they're not applicable everywhere)
- they relate to waiting on API responses so they're not actually part of the page itself.

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/ChangePlan/ChangePlanModal.tsx
Tbaut: Looks good to me. We can only have ternary statements in react indeed, and having a dedicated function for a test name wouldn't really make more sense I guess. Just a small syntax simplification suggestion, since `slide` is already a string.
```suggestion
      testId={slide}
```

Tbaut (approved):
Awesome, that's a pretty huge one.

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/ChangePlan/ChangePlanModal.tsx
asnaith: @Tbaut Ah yes, thank you 👍

tanmoyAtb (approved):
The tests looks very well scoped 👍

PR #1889, opened 2022-01-28
CIDs paging [storage]

### LINKED ISSUE #1325: Add paging to CIDs page
Add pagination to the CIDs page,
The lists API already has support for limits, need to check if there's support for page number.

We need an API endpoint that will return the total number of PinStatus objects that the user has created, and this would drive the logic of whether there are additional pages or not.

tanmoyAtb:
closes #1325

Some functionality to look into
Move into next pages,
Move into previous pages,
Search Pins,
Add pin,
Unpin

---
Submission checklist: 

   

#### Layout

#### Theme


FSM1 (changes requested):
The code looks really great. The only comment I have is when switching pages, both the buttons become disabled (which is expected), and a spinner appears in both of the page buttons  (this is less expected). 

Not sure what the solution is here - either show a spinner only on the button that was clicked, or do not have a spinner on the button during the request at all.

https://user-images.githubusercontent.com/12774278/163420485-9a84fe67-bc7c-416b-8c36-90663175e032.mp4

tanmoyAtb:
> The code looks really great. The only comment I have is when switching pages, both the buttons become disabled (which is expected), and a spinner appears in both of the page buttons (this is less expected).
> 
> Not sure what the solution is here - either show a spinner only on the button that was clicked, or do not have a spinner on the button during the request at all.
> 
>  CIDs.-.Chainsafe.Storage.-.Brave.2022-04-14.17-11-51.mp4

All right, we're handling to show spinner only on the next or previous button now. 👍

asnaith:
Hey @tanmoyAtb, I found an issue that I've demonstrated in the video below. I had two pages of cids but after sorting one of the columns the paging system broke. 

https://user-images.githubusercontent.com/3982665/163749499-d22c3df9-42ef-404c-87f1-9aa515a1dc26.mov

tanmoyAtb:
> Hey @tanmoyAtb, I found an issue that I've demonstrated in the video below. I had two pages of cids but after sorting one of the columns the paging system broke.
> 
>  breaking.paging.mov

Ahh yes, that makes sense, That's a great catch @asnaith. I'll be looking into it.

tanmoyAtb:
> Hey @tanmoyAtb, I found an issue that I've demonstrated in the video below. I had two pages of cids but after sorting one of the columns the paging system broke.
> 
>  breaking.paging.mov

The sort directions should work properly now. @asnaith

packages/storage-ui/src/Components/Pages/CidsPage.tsx
```diff
@@ -233,6 +262,20 @@ const CidsPage = () => {
           </TableBody>
         </Table>
       </div>
+      {!!pins.length &&
```
Tbaut: I feel we should hide it when there's just 1 page.
```suggestion
      {!!pins.length && pins.length > 1 &&
```

packages/storage-ui/src/Contexts/StorageContext.tsx
```diff
@@ -102,19 +135,115 @@ const StorageProvider = ({ children }: StorageContextProps) => {
     }
   }, [storageApiClient])
 
-  const refreshPins = useCallback(({ searchedCid = undefined, searchedName = "" }: RefreshPinParams = {}) => {
+  const refreshPins = useCallback(() => {
     storageApiClient.listPins(
-      searchedCid ? [searchedCid] : undefined,
-      searchedName,
+      pinsParams.searchedCID ? [pinsParams.searchedCID] : undefined,
+      pinsParams.searchedName || undefined,
```
Tbaut: I guess this should work as well
```suggestion
      pinsParams.searchedName,
```

packages/storage-ui/src/Contexts/StorageContext.tsx
```diff
@@ -102,19 +135,115 @@ const StorageProvider = ({ children }: StorageContextProps) => {
...
+        if (
+          pinsResult.results?.length &&
+          pinsResult.count &&
+          pinsResult.results.length < pinsResult.count
+        ) {
+          setIsPreviousPins(true)
+        } else {
+          setIsPreviousPins(false)
+        }
+      }
```
Tbaut: This is very very verbose and I'm not sure I can follow everything. I tried to sum it up here, but I'm not 100% sure I nailed it.

```js
const showNextOrPrevious = pinsResult.results?.length && pinsResult.count &&  pinsResult.results.length < pinsResult.count

pinsParams.pinsRange.before
  ? setIsNextPins(showNextOrPrevious)
  : setIsPreviousPins(showNextOrPrevious)
```

Tbaut (approved):
I've had a hard time concentrate to properly understand the logic happening here, I'm particularly confused by the dates, I bet this is what we base our sorting on, rather than any id, also flags like `isPreviousPin`, were confusing maybe it's a naming thing.
In any case I made a couple suggestions, but playing with it was flawless :rocket:

packages/storage-ui/src/Contexts/StorageContext.tsx
tanmoyAtb: You're right, Didn't realise I ended up with the same logic for both cases in the end, with the logic changing so many times. Going to work around this part.

tanmoyAtb:
> I've had a hard time concentrate to properly understand the logic happening here, I'm particularly confused by the dates, I bet this is what we base our sorting on, rather than any id, also flags like `isPreviousPin`, were confusing maybe it's a naming thing. In any case I made a couple suggestions, but playing with it was flawless 🚀

Agreed, Going to improve the variable names here. 

So with the dates. There has been a lot of back and forth with this logic to get here and its a bit tricky.
To fetch next pins -> we pass a `before` date based on the oldest pin date available in view, we fetch them in date `descending` order, so the we get pins closest to the `before` date. 
For previous pins -> We pass an `after` date based on the latest pin in view, and fetch in date `ascending` order, again to get the latest pins closest the the `after` date.

FSM1 (approved):

PR #1896, opened 2022-01-31
No search results response handle

### LINKED ISSUE #1894: Handle API response when no results with search.
In files `/Drive` page, search for a random string that has no results. 

The App crashes.

tanmoyAtb:
closes #1894

### DIFF
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/SearchFileBrowser.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/SearchFileBrowser.tsx
@@ -17,6 +17,8 @@ const SearchFileBrowser: React.FC<IFileBrowserModuleProps> = ({ controls = false
   const { filesApiClient } = useFilesApi()
   const { buckets } = useFiles()
   const searchTerm = useMemo(() => decodeURIComponent(pathname.split("/").slice(2)[0]), [pathname])
+  const [loadingSearchResults, setLoadingSearchResults] = useState(true)
+  const [searchResults, setSearchResults] = useState<SearchEntry[]>([])
   const { redirect } = useHistory()
 
   const { addToast } = useToasts()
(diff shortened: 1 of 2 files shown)


asnaith:
Nice fix, I'm no longer seeing the crash. 👍

 I am seeing duplicate toasts though, is this related or should I create a separate issue for this? 

https://user-images.githubusercontent.com/3982665/151848615-7352370d-58a8-4f6c-a469-8d13e81728fe.mov

FSM1 (approved):

FSM1:
> I am seeing duplicate toasts though, is this related or should I create a separate issue for this?

Yeah also just noticed that. At first I thought it was purely the two consecutive calls failing, but turns out the error is thrown because of a empty file system.

tanmoyAtb:
> Nice fix, I'm no longer seeing the crash. 👍
> 
> I am seeing duplicate toasts though, is this related or should I create a separate issue for this?
> 
>  Screen.Recording.2022-01-31.at.10.04.33.AM.mov

Good find. I'll check into this as well.

Tbaut:
> TypeError: null has no properties

That's the error in the console

tanmoyAtb:
> > TypeError: null has no properties
> 
> That's the error in the console

Updates are in.

FSM1 (approved):

packages/files-ui/src/Components/Modules/FileBrowsers/SearchFileBrowser.tsx
```diff
@@ -63,8 +55,7 @@ const SearchFileBrowser: React.FC<IFileBrowserModuleProps> = ({ controls = false
       }
     }
     onSearch()
-    // eslint-disable-next-line
-  }, [searchTerm])
+  }, [searchTerm, getSearchResults])
```
Tbaut, line 58: :heart:

Tbaut (approved):
awesome

PR #1903, opened 2022-02-01
Fix time left in crypto payment

Tbaut:
closes a bug that I noticed recently. If you start a crypto payment, then go back to it 20min later (using the "see payment info" button for instance), the time remaining will correctly show that you have 40min left, but the loading progression will actually be 100%. I fixed this here.

---
Submission checklist: 

   

#### Layout

#### Theme


tanmoyAtb (approved):
💯

asnaith:
Nice fix, looks good. I had no idea this element was supposed to work like that! 😄

asnaith (approved):

PR #1910, opened 2022-02-02
Style the error boundaries component

### LINKED ISSUE #41: Create a styled Error Fallback component
Currently the Error Boundary fallback component is hard-coded and unstyled. This should be updated to use the styling solution we are using for the remainder of the app. See `files-ui/src/app.tsx` for the current implementation.

Tbaut:
closes #41 ooooh that feels good!

I wasn't able to use `createStyle`, the app kept crashing, saying I wasn't able to use `useTheme`. so I did this all manually :man_shrugging:.

Edit: error boundaries are a weird kind of component based on old fashion classes where you can't use hooks basically https://blog.lookitskris.com/error-boundaries-hooks/

image
image

---
Submission checklist: 

   

#### Layout

#### Theme

### DIFF
--- a/packages/files-ui/src/App.tsx
+++ b/packages/files-ui/src/App.tsx
@@ -1,9 +1,9 @@
-import React, { useCallback } from "react"
-import { init as initSentry, ErrorBoundary, showReportDialog } from "@sentry/react"
+import React from "react"
+import { init as initSentry, ErrorBoundary } from "@sentry/react"
 import { Web3Provider } from "@chainsafe/web3-context"
 import { ThemeSwitcher } from "@chainsafe/common-theme"
 import "@chainsafe/common-theme/dist/font-faces.css"
-import { Button, CssBaseline, Modal, Router, ToastProvider, Typography } from "@chainsafe/common-components"
+import { CssBaseline, Router, ToastProvider } from "@chainsafe/common-components"
 import { FilesProvider } from "./Contexts/FilesContext"
 import FilesRoutes from "./Components/FilesRoutes"
 import AppWrapper from "./Components/Layouts/AppWrapper"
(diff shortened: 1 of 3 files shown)


asnaith:
I ran this branch locally and temporarily reverted Tanmoy's search fix to test this out. Looking good in both themes. 

The close button was not responding to my click action and I also noticed that after around 6 seconds of being displayed the styling for the error disappears and the unstyled error is displayed. Video below: 

https://user-images.githubusercontent.com/3982665/152240861-ea121367-a076-48f8-9258-204890411afa.mov

Tbaut:
What you see after some time is the react development error (which is still nice actually) because you run a dev server. This would not be shown in prod. or on render. The reason it is lagging is because this error is being prepared. Close it with the cross at the top right and you'll see the error boundaries component again, and the close button should then work.

asnaith:
> What you see after some time is the react development error (which is still nice actually) because you run a dev server. This would not be shown in prod. or on render. The reason it is lagging is because this error is being prepared. Close it with the cross at the top right and you'll see the error boundaries component again, and the close button should then work. 

Sounds good 👍 thanks for the explanation. I wasn't sure what was happening there.

asnaith (approved):
Re-tested and closed the react development error, all is beautiful 🔥

tanmoyAtb:
Its amazing how the light and dark theme `vars` worked out. 👍 

I'm not sure how useful the stack part of the modal would be. This is what I got on a local build version.

Screenshot 2022-02-04 at 11 54 29 PM

@asnaith thanks for giving the idea on how to best test this PR :)

Tbaut:
Yeah, the stack can be useful, as well as useless, it'll really depend on the error itself!
In my case (and Andrew's as well) it's actually super helpful because we know which component the error is coming from, but it may be obscure in a build indeed..

FSM1 (approved):

PR #1919, opened 2022-02-07
Add ui test for downgrading subscription plan

### LINKED ISSUE #1918: Add ui test for subscription plan downgrade
Add a new test that covers downgrading to a lower tier plan after the account has been upgraded. 

Note: It's not possible to begin a test on a specific plan programmatically so the upgrade process will have to be part of the test.

asnaith:
closes #1918

---

### DIFF
--- a/packages/files-ui/cypress/tests/subscription-plan-spec.ts
+++ b/packages/files-ui/cypress/tests/subscription-plan-spec.ts
@@ -13,6 +13,7 @@ import { selectPaymentMethodModal } from "../support/page-objects/modals/billing
 import { planChangeConfirmationModal } from "../support/page-objects/modals/billing/planChangeConfirmationModal"
 import { planChangeSuccessModal } from "../support/page-objects/modals/billing/planChangeSuccessModal"
 import { billingHistoryPage } from "../support/page-objects/billingHistoryPage"
+import { downgradeDetailsModal } from "../support/page-objects/modals/billing/downgradeDetailsModal"
 
 describe("Subscription Plan", () => {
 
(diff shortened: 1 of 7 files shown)


asnaith:
This covers downgrading from a premium plan to a standard plan. 

Should we supplement this with a test to move from a standard plan back to the free plan or would that be redundant?

packages/files-ui/cypress/tests/subscription-plan-spec.ts
```diff
@@ -369,5 +372,90 @@ describe("Subscription Plan", () => {
         })
       })
     })
+
+    it("can downgrade from premium plan to standard plan", () => {
+      cy.web3Login({ deleteCreditCard: true, resetToFreePlan: true })
+
+      // upgrade the account first
+      navigationMenu.settingsNavButton().click()
```
Tbaut, line 371: We've done this before, so instead of re-typing it all. Can we maybe extract this into a function and re-use it, so that it's only present once in the code-base?

Tbaut (commented):
Just a nit to avoid duplicating code

packages/files-ui/cypress/tests/subscription-plan-spec.ts
asnaith: Yep I can do. My original thinking was that referencing the cypress aliases created in a function stored from the page object could be confusing to someone reading/modifying the test but I will leave comments explaining.

asnaith: I've pushed an update that includes more re-usable code.

I've also added a second test for downgrading from standard to a free plan as some of the elements / steps in the flow are different. 

I've reduced the repetition, especially in the upgrade part, but there are still some similarities to both tests because they are similar in behaviour (but I think it makes sense for them to be individual tests).

FSM1 (approved):

Tbaut (approved):
:100:

Issue #1911, opened 2022-02-02
Add some information regarding the crypto time lock


Tbaut:
We should add some information to tell our users that the cryto price will be locked for 60min

serenaho:
Here is my suggested solution/copy. Let me know if it makes sense:

![Change Plan _ Review Plan Change.png](https://images.zenhubusercontent.com/615b27919165edef4fb3f145/72ef15cb-5961-4dc6-b699-c04d9be4cf5b)

Tbaut:
That's perfect, thank you so much

PR #1924, opened 2022-02-08
Add info messages before payment and on downgrade

### LINKED ISSUE #1912: Add Final Sale/Non-Refundable warnings to Billing Flow
On the downgrade modal
- for credit card we should tell that the users will get unused credit
- for crypto we should tell that there is no refund.

Tbaut:
closes #1912

---
Submission checklist: 

   

#### Layout

#### Theme


FSM1:
Overall looks great, though I did notice that one of the info icons is grey while the other is red
image
image

Tbaut:
yeah I followed the design https://github.com/ChainSafe/ui-monorepo/issues/1911 that's from another PR
but I agree we shouldn't do things differently. I'll make it the same.

FSM1 (approved):

asnaith:
I pushed a commit to this branch to update the tests to check for the presence of these additional elements (just the downgrade and change plan warnings, for now, I haven't automated the pay with crypto modal yet).

asnaith (approved):

PR #1925, opened 2022-02-08
Fix User Profile Button text overflow

### LINKED ISSUE #1920: Fix user sign-out button
The button looks fine using a wallet address but touches the edge when credentials are different like so:

We should try and fix this with other credential considerations in mind such as Files IDs and usernames. Perhaps we could set a max width for the button and just add a “…” at the end of the credential when it gets too long.

FSM1:
closes #1920

image

---
Submission checklist: 

#### Layout

#### Theme


Tbaut (approved):
:100:

asnaith (approved):

PR #1934, opened 2022-02-10
Adapt Settings to the design and split the Display and Profile settings

### LINKED ISSUE #1907: Consistent Header + Menu Font Sizing
Small thing that I noticed while looking through the current stage is the inconsistency of text-sizing as pictured below:

**How it should be:**

We don't have to strictly follow the 16px or 20px, it's just a concern of consistent/relevant sizing. The settings tab headers should also be bolded depending on which tab they're on.

Tbaut:
closes #1907
The easiest is to play with it, and see if you find anything inconsistent.

---
Submission checklist: 

   

#### Layout

#### Theme


asnaith (approved):
Looks great and makes sense. It's nicer to use on mobile too 👍

packages/files-ui/src/Components/Elements/InvoiceLines.tsx
```diff
@@ -56,11 +57,16 @@ const useStyles = makeStyles(
         color: palette.additional["volcano"][7]
       },
       price: {
-        fontWeight: "bold !important" as "bold"
+        fontWeight: "bold !important" as "bold",
```
tanmoyAtb, line 60: Weird that the types don't recognize `!important` here

Tbaut: Yup, the type checking is dumb here.

tanmoyAtb (approved):
Amazing amazing work, UX is much better. 

Just found this small margin offset, we might not keep this section anyway, still putting it here. 
The button on first and last name,

Screenshot 2022-02-11 at 8 23 19 PM

Tbaut:
Thanks, whether or not we keep it, I guess we would only comment it out, so I'll def fix that now

FSM1 (approved):
Great work, this feels much smoother an experience🎉

Tbaut:
Little heads up, I added a last minute small change to my settings PR, by removing the lock icon to the "Save change" button because IMHO a lock doesn't fit with saving a user name (this same icon is used on the security tab) and we don't use icons on buttons in general. lmk if you disagree. Also a note to myself, when disabled, this button has the same color as the background and looks totally out of place :) 

image

PR #1936, opened 2022-02-11
Cross position in subscription modal fix

tanmoyAtb:
We had a small bug on the subscription modal with some screen sizes. Here's a screenshot.

Screenshot 2022-02-11 at 8 48 28 PM

This PR fixes the problem.


Tbaut (commented):
Nice, while we're here, I noticed there's the same css used in `PayInvoiceModal.tsx` with the same bug, fixed by the same 480 instead of `undefined`, can you add it there too?

FSM1:
This change does mess with the padding that the modal usually has:
Before:
image

After:
image

asnaith:
Looks good 🔥

I will re-test when crypto payment modal gets the same fix applied

tanmoyAtb:
All right, this should be good now. 👍

asnaith (approved):

Tbaut:
I'll let @FSM1 confirm it's looking good

FSM1 (approved):
Works well 🎉

PR #1940, opened 2022-02-11
owner removed from shared with list

### LINKED ISSUE #1929: Remove owner from the shared with list
When looking at the sharing overview, I had the impression that the owner would not be part of the shared user column. But it actually is, using its username, making it even more confusing. Here's an example, I'm the owner of this one, and the first person that's shared with is actually the same, myself:
image

tanmoyAtb:
closes #1929

This PR removes the `owner` from shared with list in shared folder overview


Tbaut (approved):
Code looks good :)

FSM1 (approved):
looks good 🎉

PR #1945, opened 2022-02-14
Hash icons in shared user list

### LINKED ISSUE #1928: Use icon in sharing overview
Now that we have nice icons, we should use them instead of the user icon I believe. I don't think we'd need design for this.
IMHO these should be updated
image

Only use identicon for the primary user, and have the `+n` in the same style as it is now

tanmoyAtb:
closes #1928

---
Submission checklist: 

   

#### Layout

#### Theme


Tbaut (changes requested):
The key for the `hashIcon` need to be the uuid I think, otherwise the icon don't match, e.g I have different icons in the manage modal, and in the overview

https://user-images.githubusercontent.com/33178835/153918001-94cf1d6f-ea8d-4243-9cb4-735efe470336.mp4

tanmoyAtb:
> The key for the `hashIcon` need to be the uuid I think, otherwise the icon don't match, e.g I have different icons in the manage modal, and in the overview
> 
>  sharing.mp4

hmm, yes, that's right. 👍

FSM1 (approved):
Working really well, and looks great 💯

tanmoyAtb:
> The key for the `hashIcon` need to be the uuid I think, otherwise the icon don't match, e.g I have different icons in the manage modal, and in the overview
> 
>  sharing.mp4

Thanks for hint. PR is updated. All the hashicons now use `uuid`, user list, user dropdown and shared users in modal. 👍

packages/files-ui/src/Components/Elements/UserBubble.tsx
```diff
@@ -63,39 +66,47 @@ const useStyles = makeStyles(({ zIndex, animation, constants, palette }: CSFThem
 interface Props {
   text?: string
   tooltip: string | string[]
+  showHashIcon?: boolean
```
Tbaut: To me this prop is redundant with the `hashIconValue`. Either both are set, or not isn't it?

packages/files-ui/src/Components/Elements/SharedUsers.tsx
```diff
@@ -25,12 +25,8 @@ const SharedUsers = ({ bucket, showOwners }: Props) => {
   const { desktop } = useThemeSwitcher()
   const { owners, readers, writers } = bucket
 
-  const getUserLabels = useCallback((users: RichUserInfo[]): string[] => {
-    return users.reduce((acc: string[], user): string[] => {
-      const displayName = getUserDisplayName(user)
-
-      return [...acc, displayName]
-    }, [] as string[])
+  const getUserLabels = useCallback((users: RichUserInfo[]) => {
```
Tbaut: That's somehow of a nit, but this function (and many other "label" variables), aren't only about labels any more. They are more about user display info. But I can close my eyes on this :D although having the return type defined would be nice I think. `interface UserDisplayInfo { displayName: string; uuid: string } `

Tbaut (approved):
Looking good now, I left a couple comments. Feel free to ignore

tanmoyAtb:
> Looking good now, I left a couple comments. Feel free to ignore

Updates are in. :D

PR #1947, opened 2022-02-14
exit and refresh the billing plan when the time is up

### LINKED ISSUE #1935: Crypto payment timer doesn't stay at 0.00 when time elapses
**Description**
Edge case: Leaving the modal open for longer than 60 minutes will result in the timer going into negative figures. 

**Steps to Reproduce**
1. Choose a plan 
2. Choose yearly duration 
3. Choose cypto as payment type
4. Leave crypto payment modal in view for over 60 minutes

**Observed behavior**
The timer does not end and instead goes into negative figures. 

It's still possible to access the QR code after payment time has elapsed if the modal was left in view. 

When the expired modal is closed the "Pay now"  button is still accessible but will result in a crypto payment modal with a never-ending spinner (because the API is returning code 400 ""invoice status other than open"). 

**Expected behavior**

TBD

It should not be possible to access this modal (mainly the QR code) when the payment will no longer be valid. 

**Screenshots**
Screen Shot 2022-02-10 at 3 33 07 PM

Screen Shot 2022-02-10 at 3 56 32 PM

Tbaut:
closes #1935

---
Submission checklist: 

No layout change


Tbaut:
while testing with and really waiting for 60min, I saw that the subscription refresh seemed to be fired too quickly, the api didn't have the time to figure that the time was up. I'll try to handle that differently. In any case testing is pretty annoying :D

asnaith:
> while testing with and really waiting for 60min, I saw that the subscription refresh seemed to be fired too quickly, the api didn't have the time to figure that the time was up. I'll try to handle that differently. In any case testing is pretty annoying :D

Ok, I was just about to check this out but I'll wait for the changes 👍

Is there any way we can temporarily decrease the timeout on our local instances for testing?

FSM1:
> while testing with and really waiting for 60min, I saw that the subscription refresh seemed to be fired too quickly, the api didn't have the time to figure that the time was up. I'll try to handle that differently. In any case testing is pretty annoying :D

One approach to take with this is to render a separate contents for `timeRemaining <= 0` with a button wired up to request a new coinbase charge (manually). This will prevent the request being made too soon.

Tbaut:
Good point with the content regarding the times up. My main problem is that the refresh subscription did not change anything immediately e.g, right now:
- time is up I refresh the subscription
- I still have a pending crypto payment according to the api
- if my UI relies on this, we still show the outstanding invoice, and would keep showing the same modal.

So if the user requests a new charge right away, it may not work.

An idea I had is to show a loader with some text that the time is up, and refresh the current subscription in a setTimeout of 10s, until we have no open invoice any more... I'll try to come up with something a little more elegant than that. Testing it is a pretty time intensive thing anyway :d

Tbaut:
I tried something out, will test it now. Basically we'll keep polling the api when the time is up, until the current subscription is not a pending one any more.

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/Common/CryptoPayment.tsx
```diff
@@ -255,6 +256,21 @@ const CryptoPayment = ({ planPrice }: ICryptoPayment) => {
   , [invoices])
   const currencies = useMemo(() => cryptoPayment?.payment_methods.map(c => c.currency), [cryptoPayment])
   const [selectedCurrency, setSelectedCurrency] = useState<string | undefined>(undefined)
+  const [isFethingSubscription, setIsFethingSubscription] = useState(false)
```
tanmoyAtb:
```suggestion
  const [isFetchingSubscription, setIsFetchingSubscription] = useState(false)
```

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/ChangePlan/ChangePlanModal.tsx
```diff
@@ -208,7 +208,11 @@ const ChangeProductModal = ({ onClose }: IChangeProductModal) => {
         paymentMethod={selectedPaymentMethod}
       />
       }
-      {slide === "cryptoPayment" && <CryptoPayment planPrice={selectedPrice} />}
+      {slide === "cryptoPayment" &&
+      <CryptoPayment
+        planPrice={selectedPrice}
+        onClose={onClose}
+      />}
```
tanmoyAtb: nit
```suggestion
      {slide === "cryptoPayment" && <CryptoPayment
        planPrice={selectedPrice}
        onClose={onClose}
      />}
```

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/Common/CryptoPayment.tsx
```diff
@@ -255,6 +256,21 @@ const CryptoPayment = ({ planPrice }: ICryptoPayment) => {
...
+      setIsFethingSubscription(true)
+      fetchCurrentSubscription()
+        .then((subscription) => {
+          if (subscription && subscription.status !== "pending_update"){
+            onClose()
+          }
+        })
+        .finally(() => setIsFethingSubscription(false))
+    }
+  }, [fetchCurrentSubscription, isFethingSubscription, onClose, timeRemaining])
```
tanmoyAtb:
```suggestion
    if(!isFetchingSubscription && timeRemaining && timeRemaining?.asSeconds() <= 0){
      setIsFetchingSubscription(true)
      fetchCurrentSubscription()
        .then((subscription) => {
          if (subscription && subscription.status !== "pending_update"){
            onClose()
          }
        })
        .finally(() => setIsFetchingSubscription(false))
    }
  }, [fetchCurrentSubscription, isFetchingSubscription, onClose, timeRemaining])
```

tanmoyAtb (approved):
Just a few nits but the code looks good, Haven't tested this. Going to let it run now.

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/Common/CryptoPayment.tsx
Tbaut: Thanks, I applied it

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/Common/CryptoPayment.tsx
Tbaut: Thanks, I applied it

Tbaut:
Last I managed to see a time up, some effect ran, and the subscription was renewed automatically.. which is I think dangerous, so I'll trying to prevent this now. This is how it looked like

https://user-images.githubusercontent.com/33178835/154161853-f07d8373-c42e-476c-82a4-4aea69d5914d.mp4

asnaith:
~~Seems to be working nicely.~~

~~After the time expired this loading spinner and info label was in view for 1 minute and 20 seconds before it automatically transitioned to a new crypto payment modal with 1 minute 20 seconds already deducted from the total time.~~

~~Think that's a pretty elegant solution.~~
 
Thought the renewal of the time period was intentional and part of the fix 😅

asnaith:
> Last I managed to see a time up, some effect ran, and the subscription was renewed automatically.. which is I think dangerous, so I'll trying to prevent this now. This is how it looked like
> 
>  weird.mp4

Good catch. I didn't navigate away from the tab when I was testing, not sure if that's relevant to it. 

For me, after the 60 min time-period had expired and renewed when I closed the crypto payment modal and re-opened it via the "Pay Now" link the timer had continued as expected and didn't renew.

Tbaut:
> For me, after the 60 min time-period had expired and renewed when I closed the crypto payment modal and re-opened it via the "Pay Now" link the timer had continued as expected and didn't renew.

Yeah, this automatic renew would only happen if the modal was open. What I want to do is let the user wait in front of a loader while we
- refresh everything (subscription/plans etc) to make sure there's no more outstanding payment shown
- close the modal once it's done

Tbaut:
ok, now I'm happy

https://user-images.githubusercontent.com/33178835/154166393-3ba4da01-d9af-4fce-adea-46d66b384f63.mp4

asnaith:
> ok, now I'm happy
> 
>  yes.mp4

This looks great now 🙌

asnaith (approved):

PR #1948, opened 2022-02-15
Files Billing soft launch 

### LINKED ISSUE #1930: Billing Soft Launch capability
To enable us to `feature flag` the billing UI only for users who are whitelisted for billing, the API will expose an endpoint to allow us to retrieve a flag for whether the current user has billing enabled (API issue [#1952](https://github.com/chainsafe/files-api/issues/1952)).

The Subscription tab of settings as well as the subscription route should be rendered only when this flag is returned as true by the API.

The API response should be stubbed out in tests so that the billing automation tests still work as expected.

FSM1:
Since the actual tab content does not get rendered, I decided not to mess with the router switch at all, as at worst a blank tab is rendered, if the user does decide to mess with the path directly.

closes #1930

---
Submission checklist: 

   

#### Layout

#### Theme


tanmoyAtb (approved):
Looks good! Having the tab conditionally rendered should do the trick without fiddling with the routes.

packages/common-components/src/Tabs/Tabs.tsx
```diff
@@ -59,29 +59,31 @@ export interface ITabsProps<TabKey> {
...
           ? children.map((elem, index) => {
-            return (
-              <li
-                data-testid={elem.props.testId}
-                key={index}
-                className={
-                  clsx(
-                    elem.props.tabKey === activeKey && "selected",
-                    classes.tabBar,
+            return (elem !== null)
```
Tbaut: Could we rather have an early `if(!elem) return null` rather than a ternary here?
I'd find it more readable.

packages/files-ui/src/Components/Modules/Settings/index.tsx
```diff
@@ -190,26 +202,18 @@ const Settings: React.FC = () => {
...
-              className={clsx(classes.tabPane, (!desktop && !path) ? classes.hideTabPane : "")}
-              title={t`Display`}
-              tabKey="display"
-              testId="tab-display"
-              icon={<ProfileIcon className={classes.profileIcon} />}
-              iconRight={<CaretRightIcon/>}
-            >
-              <DisplayTab />
-            </TabPane>
+            {isBillingEnabled
```
Tbaut, line 218: I think `isBillingEnabled && <TabPane...` should do the trick. Trying to avoid a ternary again :D

packages/files-ui/src/Contexts/BillingContext.tsx
```diff
@@ -87,6 +89,12 @@ const BillingProvider = ({ children }: BillingContextProps) => {
     refreshInvoices()
   }, [refreshInvoices])
 
+  useEffect(() => {
+    if (!isLoggedIn) return
+
+    filesApiClient.getEligibility().then(res => setIsBillingEnabled(res.is_eligible))
```
Tbaut: to be on the safe side
```suggestion
    filesApiClient.getEligibility()
      .then(res => setIsBillingEnabled(res.is_eligible))
      .catch(console.error)
```

Tbaut (commented):
Found a couple nits :)
Happy with not fiddling with the route either if it looks ok

packages/files-ui/src/Components/Modules/Settings/index.tsx
FSM1: unfortunately not, since TabPane children need to be of a specific type, which doesnt include boolean values

packages/common-components/src/Tabs/Tabs.tsx
```diff
@@ -75,7 +76,7 @@ const Tabs = <TabKey, >({ className, children, activeKey, injectedClass, onTabSe
                   clsx(
                     elem.props.tabKey === activeKey && "selected",
                     classes.tabBar,
-                    injectedClass?.tabBar
+                      injectedClass?.tabBar
```
Tbaut: Tiny nit, committing right away to merge
```suggestion
                    injectedClass?.tabBar
```

Tbaut (approved):
awesome!

PR #1953, opened 2022-02-16
Refactor shared folder creation

### LINKED ISSUE #1943: Shared folder creation should allow link creation right away
As discussed:
- Shared folder creation should first ask for the shared folder name, and nothing else
- upon creation we should show the manage permission modal, allowing to add users/ create links

Tbaut:
closes #1943

---
Submission checklist: 

   

#### Layout

#### Theme


Tbaut:
The test should now pass, note @asnaith that due to my refactoring, the tests aren't as nice as they used to be. For instance to create a shared folder (the helper function that is only used once), I do the following:
- navigate to the shared folder
- click on the create button
- type in a name
- click on "Create"
- click on "Cancel" on the next screen, because I don't care about adding any user. I find this step somewhat inelegant.

asnaith (approved):
@Tbaut the new modals are working really well and they're a big ux improvement especially on mobile 🚀

The test edits look great too, I pushed a few tiny commits for minor things but that's all. As for the extra step, I don't think it's too inelegant - it's 1 of only 2 valid paths that the user can take at that point of the share flow 🙂. 

It did make me realize that we are lacking automated test coverage for adding by name / wallet / address / ens though. I'll create a ticket for that!

FSM1 (approved):
working really nicely 🎉

FSM1:
> @Tbaut the new modals are working really well and they're a big ux improvement especially on mobile 🚀
> 
> The test edits look great too, I pushed a few tiny commits for minor things but that's all. As for the extra step, I don't think it's too inelegant - it's 1 of only 2 valid paths that the user can take at that point of the share flow 🙂.
> 
> It did make me realize that we are lacking automated test coverage for adding by name / wallet / address / ens though. I'll create a ticket for that!

hahah Ryan's staging user is going to have ALL the things shared with him 😂

PR #1957, opened 2022-02-17
Comment out name inputs in user profile

### LINKED ISSUE #1956: Remove name and lastname input in display
If the name and last name aren't used anywhere other than the login button we should probably remove them to clear things up in the profile settings (there are a lot of other things to set here already) and if we don't have good use for them, for privacy reason we should probably not ask for it.

What's your opinion?

Tbaut:
closes #1956

---
Submission checklist: 

   

#### Layout

#### Theme


asnaith (approved):
👍 Makes sense to comment out the tests for now in case we re-introduce these inputs when we have a use case for the name data. 

We could have used `it.skip` as an alternative for this too.

FSM1 (approved):

PR #1958, opened 2022-02-18
Add ui test for downgrade plan selection 

### LINKED ISSUE #1949: Add ui test coverage for plan selection availability
Add test coverage for checking the availability of the "Select" button on the plan selection modal based on different circumstances. 

If possible, also look into the possibility of intercepting the summary endpoint and stubbing the quota to ensure a downgrade is not possible if the user is using more storage than the plan allows.

asnaith:
closes #1949

- A check to ensure that a user can't choose a plan if their storage quota is above that plan's allowance


packages/files-ui/cypress/tests/subscription-plan-spec.ts
```diff
@@ -214,6 +214,40 @@ describe("Subscription Plan", () => {
...
+      navigationMenu.settingsNavButton().click()
+      settingsPage.upgradeSubscription("premium")
+
+      // setup intercepter, stub the used storage value to be higher
+      cy.intercept("GET", "**/buckets/summary", (req) => {
+        req.on("response", (res) => {
+          res.body.used_storage = "9704775608"
+        })
+      })
+
```
Tbaut: This isn't needed right? It's just for the UI to be consistent, but since it's a test, I don't think we should keep it
```suggestion
```

Tbaut (approved):
Looks good beside my suggestion :)

asnaith:
@Tbaut yeah you're right, a bit of an oversight - I had left this in as originally I believed this would be all that was necessary to invoke the plan warning (before I found about `is_update_allowed`). 

It does put the UI in the state that the user would have if really in this scenario but it really only changes the display label and progress bar, it's not needed to invoke the state we're asserting on so I agree it can be removed.

packages/files-ui/cypress/tests/subscription-plan-spec.ts
asnaith: @Tbaut yeah you're right, a bit of an oversight - I had left this in as originally I believed this would be all that was necessary to invoke the plan warning (before I found about is_update_allowed).

It does put the UI in the state that the user would have if really in this scenario but it really only changes the display label and progress bar, it's not needed to invoke the state we're asserting on so I agree it can be removed.

FSM1 (approved):

PR #1960, opened 2022-02-21
File preview updates

### LINKED ISSUE #1909: General feedback on the file preview
@Tbaut commented on [Tue Mar 02 2021](https://github.com/ChainSafe/files-design/issues/91)

I have had this feeling since the beginning, I just realized what I thought was a bug, was actually on purpose.
The top menu stopping _almost_ in the middle of the screen looks weird to me, also having the options button here is not where I'd expect it to be (in a corner).
 
image

Another remark, I've felt lost a often times looking for a way to close this full screen modal, looking for a cross in the top right, but ended up realizing that the left arrow at the top left was achieving this function. I think this is not intuitive and we should have a cross, in the top right hand corner.

---

@Tbaut commented on [Fri Mar 05 2021](https://github.com/ChainSafe/files-design/issues/91#issuecomment-791421653)

Transfered to the design repo as discussed, so that we can have new mocks for it, since it's not a bug.

FSM1:
closes #1909 

---
Submission checklist: 

   

#### Layout

#### Theme


packages/files-ui/src/Components/Modules/FilePreviewModal.tsx
```diff
@@ -294,16 +273,19 @@ const FilePreviewModal = ({ file, nextFile, previousFile, closePreview, filePath
...
-          style={{
-            focusVisible: classes.focusVisible,
-            menuWrapper: classes.menuWrapper,
-            root: classes.menuRoot
-          }}
-        />
+        <Button
+          onClick={handleDownload}
+          variant="text"
+          data-cy="button-download-previewed-file">
```
Tbaut:
```suggestion
          data-cy="button-download-previewed-file"
        >
```

packages/files-ui/src/Components/Modules/FilePreviewModal.tsx
```diff
@@ -294,16 +273,19 @@ const FilePreviewModal = ({ file, nextFile, previousFile, closePreview, filePath
...
+          <DownloadIcon className={classes.downloadIcon}/>
+        </Button>
+        <Button
+          onClick={closePreview}
+          variant='text'
+          className={classes.closePreviewButton}
+          data-cy="button-close-preview">
```
Tbaut:
```suggestion
          data-cy="button-close-preview"
        >
```

packages/files-ui/src/Components/Modules/PreviewRenderers/PDFPreview.tsx
```diff
@@ -2,32 +2,35 @@ import React, { useEffect, useState } from "react"
...
       position: "absolute",
       zIndex: zIndex?.layer1,
       display: "flex",
       flexDirection: "row",
       alignItems: "center",
-      bottom: 0,
-      backgroundColor: palette.additional["gray"][9],
-      color: palette.additional["gray"][3],
+      bottom: 34,
+      backgroundColor: "#262626",
```
Tbaut, line 18: Can't we use the `palette.additional["gray"][x]` for these? I believe that's the origin of the problems I had in dark mode

Tbaut (commented):
I've had some small issues with the "active" state of buttons.
like below on the "go back to overall view" button
image

Or the next button



https://user-images.githubusercontent.com/33178835/154969151-2a873d0c-9e61-4f15-a817-cdd5d2624ce0.mp4

packages/files-ui/src/Components/Modules/PreviewRenderers/PDFPreview.tsx
FSM1: problem with using the `["gray"].[x]` is that these colors change depending on which theme is chosen, whereas the Preview components should be insensitive to theme changes. Not seeing any problems with dark mode from the screenshots below

Light mode:
image

Dark Mode:
image

FSM1:
> I've had some small issues with the "active" state of buttons. like below on the "go back to overall view" button image
> 
> Or the next button
> 
>  next.mp4

Yeah the active state does mess things up, but this is out of the scope of this issue. The problem should be resolved in the button component in my mind.

Tbaut:
yeah the problem I mentioned are all on the active state of buttons, both the arrow and the "back to overall view"

Tbaut (approved):

tanmoyAtb:
All previews seem to be working great, 

Only for `html` files, I found the zoom buttons on top right appearing behind, 
Here's a preview - 
Screenshot 2022-02-22 at 11 37 00 PM

FSM1:
> All previews seem to be working great,
> 
> Only for `html` files, I found the zoom buttons on top right appearing behind, Here's a preview - Screenshot 2022-02-22 at 11 37 00 PM

ahhh whoops, I may have completely forgotten to look at the text file renderer. let me check that out again. will ping you when I have tested that

FSM1:
> > All previews seem to be working great,
> > Only for `html` files, I found the zoom buttons on top right appearing behind, Here's a preview - Screenshot 2022-02-22 at 11 37 00 PM
> 
> ahhh whoops, I may have completely forgotten to look at the text file renderer. let me check that out again. will ping you when I have tested that

Fixed in the latest commit. Great catch @tanmoyAtb.

asnaith:
Looking good for the most part on the files I tested but I did find a few issues (1 of which is pre-existing though so I've logged a separate issue for it). 

1) One issue that is an introduction is the text size options on mobile - they're not working as intended. It looks like we don't show the increase/decrease options on our current release so perhaps the fix is just to omit them from being displayed on mobile?

Video:

https://user-images.githubusercontent.com/3982665/155213893-8230e98e-0faf-4bac-bbca-ccc227399117.MP4

2) Another introduction is specific to iOS. When I opened an .mp4 and then tried to close the preview modal it would show me empty video controls from the OS instead of the drive page. I recorded a video of the actual interaction because it's easier to see rather than on a screen recording. It doesn't do this on iOS with the current stage / prod builds. 

Video:

https://user-images.githubusercontent.com/3982665/155360111-f8d84832-1469-42ac-ba9b-dd84a6bf83cf.mov

FSM1:
> 1. One issue that is an introduction is the text size options on mobile - they're not working as intended. It looks like we don't show the increase/decrease options on our current release so perhaps the fix is just to omit them from being displayed on mobile?
 
Resolved this. Great catch.

> 2. Another introduction is specific to iOS. When I opened an .mp4 and then tried to close the preview modal it would show me empty video controls from the OS instead of the drive page. I recorded a video of the actual interaction because it's easier to see rather than on a screen recording. It doesn't do this on iOS with the current stage / prod builds.

The video uploaded does not seem to be showing anything. Which browser were you using on iOS when testing this? There were no changes made to the video preview so I am not too sure why the behavior would be any different.

tanmoyAtb (approved):
Working great on all formats I could think of.

asnaith (approved):
As we discussed in planning, I've created a separate issue for the iOS / Safari video issue https://github.com/ChainSafe/ui-monorepo/issues/1970. 

Everything else looks good.

PR #1963, opened 2022-02-21
Circular progress bar fix.

### LINKED ISSUE #1904: Circular progress bar doesn't reflect the real progress
It's somewhat a nit, but it annoyed me enough to log it :D

If you play with the Circular progress on Storybook, and set the progress to 40 (click 3 times on "add 10"), it'll look like 50%, e.g
image

This is 50% (clicking 4 times on add 10)
image

I figured this is directly related to the width, here's the same 40%, and 50% with a 100px width - it's looking good:
image
image

I tried to dig a bit, but couldn't figure out quickly enough.

Either one of you has an idea, or we can forget about it for now.

tanmoyAtb:
closes #1904

The basic problem was adjustment due the progress stroke width.
See the storybooks too check.

---
Submission checklist: 

   

#### Layout

#### Theme

### DIFF
--- a/packages/common-components/src/ProgressBar/CircularProgressBar.tsx
+++ b/packages/common-components/src/ProgressBar/CircularProgressBar.tsx
@@ -29,8 +29,7 @@ const useStyles = makeStyles((theme: ITheme) =>
       }
     },
     progressBar: (props: IStyleProps) => ({
-      strokeLinecap: "round",
-      strokeDasharray: `${props.circumference}px ${props.circumference}px`,
+      strokeDasharray: `${props.circumference}px`,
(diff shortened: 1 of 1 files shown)


Tbaut (approved):
the fix is over my head, but it's working great!

FSM1 (approved):
this is 🔥

asnaith:
I must admit I hadn't previously set a timer and then checked to see if 25% was actually 25% etc...but I did now 😄 

Screen Shot 2022-02-22 at 1 35 53 PM

Really nice find and fix 🔥

asnaith (approved):

PR #1965, opened 2022-02-22
Make Storage tests pass

### LINKED ISSUE #1434: Improve apiTestHelper for Storage UI tests
_Related to the enabling of Storage UI tests in #1283_

Currently the **storage ui apiTestHelper** is not working reliably enough for us to build CID pinning tests and we need to add the ability to delete buckets too so that we can begin a bucket creation test with a clean data state.

Task: 

1) Investigate and determine wether `clearPins` is currently broken (and fix if it is) or clarify it is working but the deletion of CIDs is not instant because of the queuing system that we have with CIDs. 

2) Add the ability to delete buckets in the test helper so that we can clear bucket state prior to the test beginning for reliable bucket tests.

Tbaut:
closes #1434


packages/storage-ui/cypress/tests/cid-management-spec.ts
```diff
@@ -22,14 +22,14 @@ describe("CID management", () => {
...
-    it.skip("can unpin a cid", () => {
+    it("can unpin a cid", () => {
       cy.web3Login({ clearPins: true })
 
       // pin and then unpin a CID
       cidsPage.addPinnedCid()
       cidsPage.cidRowKebabButton().click()
       cidsPage.unpinMenuOption().click()
-      cidsPage.cidItemRow().should("not.exist")
+      cidsPage.cidItemRow().should("contain.text", "queued")
```
Tbaut, line 34: This is super tough to test because the pins are in a "queued" state when they are deleted. So in this test, if you let it run, what happens is that:
- the first test runs and created a pin say with cid 123
- the second test runs and starts by deleting pins -> pin with cid 123 goes into "queued" status
- the second test creates a pin with cid 123 -> then deletes it and we have this `should("contain.text", "queued")` that actually targets the pin being deleted from the first test erroneously.

How to go around this:
- we definitely need to have 2 different cid between the tests
- we need a way to target the new line, from the second test, with the second cid. Maybe you have a clue how to do this @asnaith I thought with an alias like `cidsPage.cidItemRow().should("contain.text", cid2).invoke("node").should("contain.text", "queued")`

asnaith: 👍 @Tbaut I need to refresh myself with this test and see but I suspect cypress alias will help us out. I'll pull your branch and take a look at this later on today

asnaith: Looking back at this feature and test with a fresher mind - to be honest, I'm starting to doubt that it's a great candidate for automation. Due to the nature of how the feature behaves, with pinning / unpinning taking an undeterminable amount of time, it's not feasible to achieve full e2e with reliability. 

Even if we keep the two tests separate and use unique CID's - in the unpin test we'll be attempting to unpin before the cid has pinned. 

I've tried intercepting and stubbing some of the GET `/pins?=status...` responses to force the UI to show that one is pinned (when really it's still queued) but I'm actually having issues unpinning one for real at the moment so I cannot mock the successful unpinning, I will look into this more tomorrow. Also, It looks like I might have to do something with the DELETE too (because the CID is actually queued I'm not going to be authorized to remove it).

Once I get this working though, because of the stubbing all we're really proving is that a row is added after a CID is entered on the modal and that a row is removed when the delete button on the kebab menu is pressed, everything else will be faked.

Tbaut: good point. Let's just do the pinning then, and that's probably the best we can do for a start.

asnaith (approved):

FSM1 (approved):
looks good 👍🏼

PR #1971, opened 2022-02-23
Release - 2022-02-23

FSM1:
Release of latest to PROD


Tbaut (approved):
Tests are passing, nothing can go wrong :D

asnaith (approved):
There's lots of changes here 😅 but I've reviewed the commit history and performed some additional smoke testing on stage for things I was unsure about. 

With our whitelist, test automation suite, and manual testing at the PR stage, we are well covered from a test/risk perspective. 

Lots of nice improvements and fixes are included in here 😃

tanmoyAtb (approved):
This is happenin !!

packages/files-ui/src/locales/de/messages.po
```diff
@@ -586,6 +748,9 @@ msgstr "Datei melden"
 msgid "Report a bug"
 msgstr "Fehler melden"
 
+msgid "Request access!"
+msgstr ""
+
```
Tbaut, line 753: I'm not sure what's wrong with the way we want to merge, but that seems to be the annoyance...

FSM1: ahhh dammit... no idea why there is a merge conflict... 🤔

Tbaut: I'll make a PR for this specifically.. let's see

Tbaut: ok that was a failed attempt, I wonder what would happen if we nuke this.
Will it make conflicts later too..
```suggestion
```

PR #1973, opened 2022-02-23
Fix colors and qr code border

### LINKED ISSUE #1967: Difficulty scanning QR code in dark theme (Android)
This is not a bug per se but a consideration for perhaps a better approach on the below modal: 

Screen Shot 2022-02-22 at 3 42 01 PM

On an iPhone 12 I have no difficulties scanning the QR code in both themes but on an Android device (Pixel 6 Pro) I cannot scan the QR code in dark mode. Scanning the code only works for me on Android in the light theme. 

I don't have a wider pool of devices to test this out on but as the Android phone I used is one of the latest I suspect it could be a problem for many more, especially lower-end devices. 

Update text colors for warning to ensure higher contrast

Tbaut:
closes #1967

image
image

---
Submission checklist: 

   

#### Layout

#### Theme

### DIFF
--- a/packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/Common/CryptoPayment.tsx
+++ b/packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/Common/CryptoPayment.tsx
@@ -110,11 +110,16 @@ const useStyles = makeStyles(({ constants, palette, zIndex, animation, breakpoin
       display: "flex",
       justifyContent: "center"
     },
+    qrCodeInside: {
+      padding: 2,
+      backgroundColor: "white",
+      height: 132
+    },
     qrCodeLabel: {
       display: "flex",
       justifyContent: "center",
       textAlign: "center",
-      color: "#AFAFAF"
+      color: palette.additional["gray"][8]
     },
     availableCurrencies: {
       display: "flex",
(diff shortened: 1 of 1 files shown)


FSM1 (approved):

packages/files-ui/src/Components/Modules/Settings/SubscriptionTab/Common/CryptoPayment.tsx
```diff
@@ -110,11 +110,16 @@ const useStyles = makeStyles(({ constants, palette, zIndex, animation, breakpoin
       display: "flex",
       justifyContent: "center"
     },
+    qrCodeInside: {
+      padding: 2,
+      backgroundColor: "white",
+      height: 132
```
FSM1, line 116: any reason for having to specify a height here?

Tbaut: yeah, for some reason, the div was a little higher than this per default.. maybe because of the svg inside, there was no pading/marging, the svg is 128, and I add 2x2px so that's where the 132 comes from.. maybe a comment would help.

Note that adding a border to the DVG directly actually makes it overflow the div :shrug:

asnaith (approved):
Nice! This now scans as expected on my Android device

Issue #1968, opened 2022-02-23
[Storage] Update app styling to align with the Files


FSM1:
- [x] Add identicon based on user UUID
- [x] Scaling of nav bar at low height values
- [x] Check all modals for styling consistency
- [x] Settings tabs font sizes
- [x] Login scaling

FSM1:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Files-5eb5ea92d698464ed9543f62/issues/ChainSafe/ui-monorepo/1968?planning-poker) with ZenHub @tanmoyAtb @Tbaut

serenaho:
@FSM1 Here are the options we discussed :)

**Option A: Notification Menu Item / Page:**

![Page](https://user-images.githubusercontent.com/91699405/156844811-4d38dbe2-6cdd-4309-b76e-6244016923d9.png)

**Option B: Notification Popup:**
Note: For this option I opted for a Log out button instead of the user's name since it's the only option in the dropdown. IMO it looks awkward to have a dropdown at the bottom of the page. However, a third option could be the username opening to the right like in this example.

![Notification Popup](https://user-images.githubusercontent.com/91699405/156844818-a5d251bf-9fdc-4428-a017-7978489d4e80.png)

PR #1983, opened 2022-02-25
Billing storage

### LINKED ISSUE #1969: [Storage] Enable Billing for Storage
Add the Billing module to Storage settings - only enabled for whitelisted users
Billing flows should be the same as those in Files.

* Do not include the account restricted mode in this change

Tbaut:
closes #1969
I left out the notifications so that it doesn't interfere too much with the design refactor #1968

---
Submission checklist: 

   

#### Layout

#### Theme
- [x] Components / elements inspected in light mode (there's no dark mode on Storage)


packages/storage-ui/src/Components/Pages/SettingsPage.tsx
```diff
@@ -174,6 +178,16 @@ const SettingsPage: React.FC = () => {
             >
               <ApiKeys />
             </TabPane>
+            <TabPane
```
Tbaut: This needs to be rendered conditionally with `{isBillingEnabled ? <TabXYX /> : null}`

asnaith:
Just noticed a couple of issues

1) The link colour is the wrong shade of blue 
Screen Shot 2022-02-25 at 2 39 56 PM

Screen Shot 2022-02-25 at 2 50 32 PM

2) Add / update card modals are not working because the inputs are not getting the response from stripe to activate. 

I was able to test the upgrade and download flows successfully by adding a card to the profile via **Files** and then logging back into Storage. I had no issues upgrading with credit card or downgrading. I didn't pay via crypto but everything else with the flow and ability to return etc is all good.

Tbaut:
Thanks Andrew for the review. I fixed the color issue for the links, and checked with the CC, from the Storage interface I was able to delete one (that I added from Files), and add one. Also note that you should make sure to have the Stripe pk env variable in you local environment. I just added to Render as it wasn't present. This should solve your issues.

Other than that things worked well for my tests.

tanmoyAtb (approved):
Great ! All the flows are working nicely ! 👍

asnaith:
>  Also note that you should make sure to have the Stripe pk env variable in you local environment. I just added to Render as it wasn't present. This should solve your issues.

Ah yes, completely forgot about the .env file! Thanks, I'll try again with that in place

asnaith:
I was able to test properly locally after adding the stripe key to the Storage UI .env 😅. 

The feature is working well, just two minor points remaining:

- The whitelist for the subscription tab access, at the moment all can access it
- The stripe key for render might need looking at (it still wasn't working for me)

Tbaut:
I relaunched the deployment for render.. and checked, the env variable is definitely set :shrug: 

Pushing the whitelist check so that we don't forget it. Any reviewer can comment it.

asnaith (approved):
Yep, I can see the latest change has implemented the whitelist 👍

Render is still not working with stripe elements for me but approving as that has nothing to do with code in this PR.

Tbaut:
I'll check the render issue more in details

PR #1991, opened 2022-03-01
Update product names

FSM1:
Update product names to align with the names on invoices, as well as add production product id's to the lookup


packages/files-ui/src/Contexts/BillingContext.tsx
```diff
@@ -39,17 +39,31 @@ const ProductMapping: {[key: string]: {
...
   prod_JwSGHB8qFx7rRM: {
-    name: t`Premium plan`,
-    description: t`Premium plan`
+    name: t`Files Max`,
+    description: ""
+  },
+  // Production Product Ids
+  prod_KRAq3CngQMKebw: {
+    name: t`Files Free`,
+    description: ""
```
Tbaut: Side comment, I wonder what the description is supposed to be here, it's empty for all..

Tbaut (approved):

packages/files-ui/src/Contexts/BillingContext.tsx
FSM1: also do not think it is used anywhere. Think we can nuke it...

Tbaut:
Tests are failing, looks like it's pretty consistent

FSM1:
> Tests are failing, looks like it's pretty consistent

yeah I am fixing these slowly... some of the selectors were using the plan names.

asnaith:
> > Tests are failing, looks like it's pretty consistent
> 
> yeah I am fixing these slowly... some of the selectors were using the plan names.

@FSM1 Ah yes, I know that part you are referring to. Unfortunately using some of the plan name text was a necessity in order to differentiate between all the plan containers and ensure we select a specific select plan button. 

If there's a way of assigning a unique identifier to each plan container that would be better but I wasn't able to find a way to do that.

tanmoyAtb:
Hey @FSM1 some of the `description` fields are still around. Not sure if it's still in progress.

FSM1:
> Hey @FSM1 some of the `description` fields are still around. Not sure if it's still in progress.

These have been removed from the mapping, and the processing of the various products, but `product.description` is still returned by the API so might appear elsewhere in the code.

tanmoyAtb (approved):
Looks good! 
On that note, we can also make the changes on storage side if needed. Storage has the old names and description field.

FSM1:
> Looks good! On that note, we can also make the changes on storage side if needed. Storage has the old names and description field.

Will leave that for later, when the storage product names are finalized and ready...

asnaith (approved):
Thanks for all the test updates 🙏

PR #1996, opened 2022-03-01
Refactor shared folder cretation and edition

### LINKED ISSUE #1984: "Manage Shared Folder" modal does not display address / keys correctly
**Description**
The layout of the "Manage Shared Folder" modal cannot properly accommodate an Ethereum address or share key, in the case of share key it makes the access rights dropdown inaccessible. 

**Steps to Reproduce**
1. Share a folder to ethereum address or share key

**Observed behavior**

Screen Shot 2022-02-25 at 1 10 03 PM

Screen Shot 2022-02-25 at 1 06 20 PM

**Expected behavior**
I guess we should truncate or resize the address / key string so that it fits correctly

**Console Error**

**Screenshots**

**Desktop:**
 - OS: macOS
 - Browser: Chrome (not browser specific)
 - Version : 98.0.4758.109

Tbaut:
closes #1984 

There were a bunch of custom types which made the handling of shared folder user quite difficult to wrap our head around.
I refactored everything to play with `LookupUser` only, which has all the info we need. Since this is much bigger than just fixing the bug in the issue, please test thoroughly the shared folder creation and edition.

---
Submission checklist: 

   

#### Layout

#### Theme


asnaith:
I had a thorough look at the feature on desktop and mobile. The fix for the original issue is working nicely. 

I found a few issues, most were pre-existing and unrelated to these changes so I've created separate issues for them. 

The only issue that I could find related to these changes is with ENS addresses, when you add a user via their ens it is displaying the wallet address in the modal.

Tbaut:
Aaah I didn't test this one. Thanks, I'll check it out.

Tbaut:
@asnaith I had a look at ens, and there was yet some more to refactor. If you have an ens name with a reverse lookup setup, then please let me know, the one from ryan meowzit.eth doesn't have it, which means that from the ens name you get the address, but from the address we don't get the ens name.

What this means is that well we can't show the ens name in the future in the UI, if the ens isn't setup. We had a hack so far, that allowed to show it in the shared folder management modal right when we added it. But once you click on "update", you'll never see the ens name again, which is inconsistent. Now we would have never looked up for the ens name for any address, this PR adds this. Now here's what happens:
- you look for meowzit.eth
- it shows you the associated address, unless there's a reverse lookup for the name.
- when you add it, you see either the address (in our case) or the ens name if it's setup.

Now if an ens name has the reverse lookup setup, then we'll show the ens name everywhere (this is a feature brought by my refactoring) lmk your thoughts.

packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -723,24 +714,10 @@ const FilesProvider = ({ children }: FilesContextProps) => {
...
-    }))
-  }, [encryptForPublicKey])
+  }, [encryptForPublicKey, filesApiClient, getKeyForBucket, getPermissionForBucket, getUsersWithEncryptionKey, publicKey, refreshBuckets])
 
   const editSharedFolder = useCallback(
-    async (bucket: BucketKeyPermission, writerUsers?: UpdateSharedFolderUser[], readerUsers?: UpdateSharedFolderUser[]) => {
+    async (bucket: BucketKeyPermission, writerUsers?: LookupUser[], readerUsers?: LookupUser[]) => {
       if (!publicKey) return
 
       if (!readerUsers || !writerUsers) return
```
tanmoyAtb, line 723: nit; We should have the readers and writers required params here as well like `createSharedFolder`

asnaith:
> 
> Now if an ens name has the reverse lookup setup, then we'll show the ens name everywhere (this is a feature brought by my refactoring) lmk your thoughts.

Makes sense to me 👍. I've made a note of this behavior in the user scenarios too. 

Unfortunately, I don't have an ens name with a reverse lookup to use for a test.

tanmoyAtb (approved):
So much cleaner ! 🚀

Tbaut:
@FSM1 I added the check you mentioned about ENS in [163f701](https://github.com/ChainSafe/ui-monorepo/pull/1996/commits/163f70112e89b9bd85b80db08b7e60ee1bb54a49)

FSM1 (approved):
working great 💯

PR #2001, opened 2022-03-03
Use toggle for account visibility

### LINKED ISSUE #1994: "Allow lookup" checkbox should be its own menu
When looking at this screenshot, the checkbox to allow lookup seems to be only related to the username, although it's a general lookup agreement.

I'd suggest to use a toggle, and make it big, and separate from the others (maybe even first wdyt?)

image

Tbaut:
closes #1994

---
Submission checklist: 

   

#### Layout

#### Theme

### DIFF
--- a/packages/common-components/src/ToggleSwitch/ToggleSwitch.tsx
+++ b/packages/common-components/src/ToggleSwitch/ToggleSwitch.tsx
@@ -1,6 +1,6 @@
 import { createStyles, ITheme, makeStyles } from "@chainsafe/common-theme"
 import clsx from "clsx"
-import React, { ReactNode, useCallback, useMemo, useState } from "react"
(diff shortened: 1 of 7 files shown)


asnaith:
This is working great and looks a lot nicer. 

I just thought it seemed a bit odd having the toggle switch to the left of the label when all the other actionable elements (copy buttons and set username) are on the right.

tanmoyAtb (approved):
Looks great !

FSM1 (approved):

Tbaut:
@asnaith tbh I'm not really sold on one way or another. The copy buttons for the address are helper, and the input for the username would actually be the first to be clicked, on the left. I think it's ok to leave it there. lmk if you have a strong opinion on that.

PR #2002, opened 2022-03-03
Access right options while lookup - manage shared folder

### LINKED ISSUE #1998: "Manage Shared Folder" access right option obscured when lookup results are open
**Description**
The access rights dropdown becomes partially obscured when the lookup results are open

**Steps to Reproduce**
1. Create a new shared folder
2. Enter the name of a user and leave lookup results open
3. Click on the access permission dropdown

**Expected behavior**
The options should never be obscured by another element.

Perhaps we should close the lookup results if the access rights dropdown is clicked? 

**Video**

https://user-images.githubusercontent.com/3982665/156274761-38e5e2b1-2a5c-41c8-833e-0f0524c55650.mov

**Desktop:**
 - OS: macOS
 Browser: Chrome (not browser specific)
 - Version : 98.0.4758.109

tanmoyAtb:
closes #1998 

Just changing z-index feels a good UX to me.

### DIFF
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/ManageSharedFolder.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/ManageSharedFolder.tsx
@@ -15,7 +15,7 @@ import PermissionsDropdown from "./Sharing/PermissionsDropdown"
 import { UserName } from "../../Elements/UserName"
 
 const useStyles = makeStyles(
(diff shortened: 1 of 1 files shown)


asnaith (approved):
> Just changing z-index feels a good UX to me.

Agreed, this fix works nicely 👍

Tbaut (approved):
:100:

PR #2004, opened 2022-03-04
Show 'me' instead of the user icon

Tbaut:
In a change I made earlier this week, I made this mistake where you'd show the user identicon, even if it's us. This gets back to what we have previously, I think it helps knowing where we are the owner right away.


FSM1:
I feel like the identicon makes for much better visual consistency and would leave it as that. The tooltip on the bubble still allows the user to see that they are the owner.

tanmoyAtb:
> I feel like the identicon makes for much better visual consistency and would leave it as that. The tooltip on the bubble still allows the user to see that they are the owner.

Agreed with this. I actually thought that the change was intentional in the other PR.

Tbaut:
ah well, then I leave it there :) I agree it's nicer with the icons, but it was not intentional :)

PR #2008, opened 2022-03-04
Resolve yarn.lock issue in gh actions test jobs

asnaith:
~~Please ignore, just using this branch for troubleshooting~~

This fixes the issue we are seeing in CI on failed tests 

``Error: The template is not valid. .github/workflows/test-files.yml (Line: 27, Col: 16): hashFiles('**/yarn.lock') failed. Fail to hash files under directory '/home/runner/work/ui-monorepo/ui-monorepo``

I suspect the wildcarding of the yarn.lock file was not a tight enough match and it was picking up a yarn.lock in one of the dependencies instead. Now we specifically target the yarn.lock in the root dir.

### DIFF
--- a/.github/workflows/test-files-on-demand.yml
+++ b/.github/workflows/test-files-on-demand.yml
@@ -15,7 +15,7 @@ jobs:
             **/node_modules
             ~/.cache/Cypress
             **/build
-          key: ${{ runner.os }}-node_modules-files-build-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-node_modules-files-build-${{ hashFiles('./yarn.lock') }}
           restore-keys: |
             ${{ runner.os }}-node_modules-build-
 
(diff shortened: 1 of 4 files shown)


Tbaut:
Nice, looks like it worked!

asnaith:
> Nice, looks like it worked!

Yeah now I'm reverting back to how it was to see if it fails again to ensure I didn't just get lucky. If I get consistent results when switching between the two changes we can say with confidence this is the issue. 

I'm suspecting it's a bug with the matching of the lock file when using wildcarding. Perhaps it's finding more than one (from a dependency or something) and getting confused. 

Let's see.

asnaith:
Yep, consistent behavior. Works fine with the yarn.lock wildcarding off, breaks with it on.

FSM1 (approved):

tanmoyAtb (approved):

PR #2019, opened 2022-03-16
Add ui tests for notifications

### LINKED ISSUE #2018: Add ui test coverage for notifications
We show a few different types of notifications in the files and storage UI. 

- Investigate what responses can be stubbed in order to show the notification 
- add test coverage for both Files and Storage

asnaith:
closes #2018

Adds tests for 
- card expiring soon notification
- open crypto invoice notification

### DIFF
--- a/packages/files-ui/cypress/fixtures/cardData.ts
+++ b/packages/files-ui/cypress/fixtures/cardData.ts
@@ -1,3 +1,5 @@
+import dayjs from "dayjs"
+
 export const visaNumber = "4242424242424242"
 export const visaCvc = "123"
 export const visaExpiry = "12/30"
@@ -7,4 +9,4 @@ export const mastercardExpiry = "01/31"
 export const invalidCardNumber = "6242424242424255"
 export const invalidCvc = "11"
 export const invalidExpiry = "02/21"
-export const currentDateExpiry = new Date().getMonth() + 1 + "/" + new Date().getFullYear().toString().substr(-2)
+export const currentDateExpiry = dayjs().format("MM/YY")
--- a/packages/files-ui/cypress/fixtures/cardData.ts
+++ b/packages/files-ui/cypress/fixtures/cardData.ts
@@ -7,3 +7,4 @@ export const mastercardExpiry = "01/31"
 export const invalidCardNumber = "6242424242424255"
 export const invalidCvc = "11"
 export const invalidExpiry = "02/21"
+export const currentDateExpiry = new Date().getMonth() + 1 + "/" + new Date().getFullYear().toString().substr(-2)
--- a/packages/files-ui/cypress/support/page-objects/settingsPage.ts
+++ b/packages/files-ui/cypress/support/page-objects/settingsPage.ts
@@ -1,5 +1,5 @@
 import { basePage } from "./basePage"
-import { visaNumber, visaExpiry, visaCvc } from "../../fixtures/cardData"
+import { visaNumber, visaExpiry, visaCvc, currentDateExpiry } from "../../fixtures/cardData"
 import { cardAddedToast } from "../../support/page-objects/toasts/cardAddedToast"
 import { selectPlanModal } from "../../support/page-objects/modals/billing/selectPlanModal"
 import { planDetailsModal } from "../../support/page-objects/modals/billing/planDetailsModal"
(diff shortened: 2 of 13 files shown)


packages/files-ui/cypress/fixtures/cardData.ts
```diff
@@ -6,4 +6,5 @@ export const mastercardCvc = "456"
 export const mastercardExpiry = "01/31"
 export const invalidCardNumber = "6242424242424255"
 export const invalidCvc = "11"
-export const invalidExpiry = "02/21"
\ No newline at end of file
+export const invalidExpiry = "02/21"
+export const currentDateExpiry = new Date().getMonth() + 1 + "/" + new Date().getFullYear().toString().substr(-2)
```
asnaith: This gives me what I need `03/22` but wondering if there's a better / preferred way?

Tbaut: we use dayjs elsewhere, so I believe you can do `dayjs().format('MM/YY')` which is a little shorter

asnaith: @Tbaut thank you, that looks cleaner and works just as well 👍

asnaith:
I realized that by introducing the test for the crypto invoice notification I have affected another test in the subscription plan spec related to crypto payments (it's not expecting an unpaid invoice row to be there at the beginning of the test). 

I might have to see if there's a way to use separate sessions for each test to prevent a conflict as there's not a way to clear down that state afaik.

Tbaut:
I pushed  commit to add a flag to generate a random session id, and force a new session.

packages/files-ui/cypress/support/page-objects/settingsPage.ts
```diff
@@ -40,8 +40,9 @@ export const settingsPage = {
   payNowButton: () => cy.get("[data-testid=button-pay-invoice]"),
 
   // use this convenience function when an upgraded account is required as a test requisite
-  upgradeSubscription(plan: "pro" | "max") {
+  upgradeSubscription(plan: "pro" | "max", card?: "expiring") {
```
Tbaut: I would use a boolean here `isCardExpired` instead
If you want to make this easier to read, then instead of having 2 params, you can change this for object 1 param, with the type `{plan: "pro"|"max", isCardExpired: boolean}`. Then when you call this function, the params are easy to read, rather than `func("pro", true)`, it'd be `func({plan: "pro", isCardExpired: true}) `

edit: Maybe what you wanted to test is the fact that it's expiring this month, the name would change, but not the idea.

Tbaut (commented):
Looks good, just left a small comment

Tbaut:
It failed with the following, but it passed locally :(
 ```
 |       1) can initiate and return to a crypto payment flow within 60 minutes
 | 
 | 
 |   9 passing (3m)
 |   1 failing
 | 
 |   1) Subscription Plan
 |        desktop
 |          can initiate and return to a crypto payment flow within 60 minutes:
 |      CypressError: Timed out retrying after 4050ms: `cy.click()` failed because this element is `disabled`:
 ```
 
edit: This looks like an api issue. The `returnToFreePlan` flag fails.

packages/files-ui/cypress/support/page-objects/settingsPage.ts
asnaith: @Tbaut thanks for the explanation and letting me know about this approach, I've updated this now 🙏.

asnaith:
> edit: This looks like an api issue. The `returnToFreePlan` flag fails.

@Tbaut Unfortunately I got stuck on this and was unable to resolve it but I'm fairly confident it's not an API issue (well, it seems related to the API but not caused by the API). I've tried many things but am not fully understanding the underlying issue. 

I can replicate the issue outside of CI though. Currently, with the two tests marked as `it.only`, the first time that you run them both they will pass as they both are using new sessions. Afterwards, if you then click the "run all tests" button to do a second run...

Screen Shot 2022-03-17 at 8 30 34 PM

The test that is restoring the session will fail on the resetToFreePlan every time, whilst the other test continues to run and pass as it is always using a new session. If you clear the sessions everything works again too (guess that's obvious 😄). So it seems like something weird is going on when using apiTestHelper after a session restore (maybe with the token refresh?). 

Screen Shot 2022-03-17 at 8 36 26 PM


I'll continue to look at it tomorrow with a fresh brain but if you have any ideas please let me know (and thanks for your help so far with this!).

Tbaut:
Hey Andrew, from the suite of calls perspective, there's 100% something weird with the api. This should be tackled in this PR https://github.com/ChainSafe/files-api/pull/2082 I couldn't test with the api preview unfortunately so we'll have to wait until it's merged.

The problem is with the "return to free subscription" helper, that is failing for no apparent reason.

Tbaut:
ok I have some news. The "back to free plan" logic does not work if the user that we have is currently having a subscription awaiting payment, which is the case when we have a crypto test right before. There is no way we can go around this api limitation.

So the solution is to use my "newSesion" flag, every time we do something with crypto.

asnaith:
@Tbaut as mentioned on slack, the issues described above are now resolved by the latest api release on stage :)

Tbaut (approved):

FSM1 (approved):

Issue #2022, opened 2022-03-17
Allow users to add a name when creating a pin


Tbaut:
Since the api allows it, we should take advantage of it.

PR #2027, opened 2022-03-17
[Storage] Add warning when pinning a cid twice

### LINKED ISSUE #1212: ChainSafe Storage: User can pin the same file multiple times
**Description**
It is possible for the user to repeatedly pin the same file. 

**To Reproduce**
1. Login to ChainSafe storage 
2. Navigate to the Cids page
3. Click on the Pin CID button
4. Paste a CID into the input box
5. Click Pin
6. Repeat steps 2 -5 with the same CID

**Expected behavior**
I don't think there's a use case for this behaviour so perhaps it would be better to inform the user that they have already pinned that CID?

**Screenshots**
Screen Shot 2021-06-30 at 11 57 15 AM

 - OS: macOS
 - Browser Chrome version 91.0.4472.114, Firefox version 89.0.2

Tbaut:
closes #1212

### DIFF
--- a/packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
+++ b/packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
@@ -1,15 +1,14 @@
-import React, { useCallback, useRef, useState } from "react"
+import React, { useCallback, useState } from "react"
 import { makeStyles, createStyles } from "@chainsafe/common-theme"
-import { Button, FormikTextInput, Grid } from "@chainsafe/common-components"
+import { Button, Grid, TextInput, Typography } from "@chainsafe/common-components"
 import CustomModal from "../Elements/CustomModal"
 import { CSSTheme } from "../../Themes/types"
 import CustomButton from "../Elements/CustomButton"
 import { t, Trans } from "@lingui/macro"
-import { Formik, Form } from "formik"
 import { useStorage } from "../../Contexts/StorageContext"
 import { cidValidator } from "../../Utils/validationSchema"
 
-const useStyles = makeStyles(({ constants, breakpoints, zIndex }: CSSTheme) =>
+const useStyles = makeStyles(({ constants, breakpoints, zIndex, palette }: CSSTheme) =>
   createStyles({
     root: {
       padding: constants.generalUnit * 4,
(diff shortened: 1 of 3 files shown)


asnaith (approved):
Good solution for the scenario, warning over prevention makes sense :)

packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
```diff
@@ -63,97 +75,136 @@ interface IAddCIDModuleProps {
...
+          setShowWarning(true)
+        }
+      })
+  }, [cid, searchCid])
+
+  const onCidChange = useCallback((cid?: string | number) => {
+    setCidError("")
+    setShowWarning(false)
+
+    cid && setCid(cid?.toString())
```
tanmoyAtb: This is preventing erasing the first character of the input. This should fix it.
```suggestion
    setCid(cid?.toString() || "")
```

tanmoyAtb (changes requested):
So i tried this approach ->
I copied a CID without the last character,
typed in the last character
but I did not get the warning.

While we are at it, On a UI note, I would prefer the CID modal to be wider than what we have.

packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
```diff
@@ -63,97 +75,136 @@ interface IAddCIDModuleProps {
...
+    cidValidator
+      .validate({ cid })
+      .then(() => {
+        setCidError("")
+        warnforDuplicatesCid()
+      })
+      .catch((e: Error) => {
+        setCidError(e.message)
+      })
+  }, [warnforDuplicatesCid])
```
FSM1:
```suggestion
  const checkForDuplicateCid = useCallback(() => {
    searchCid(cid)
      .then((res) => {
        if(res && res.results?.length){
          setShowWarning(true)
        }
      })
  }, [cid, searchCid])

  const onCidChange = useCallback((cid?: string | number) => {
    setCidError("")
    setShowWarning(false)

    setCid(cid?.toString() || "")

    cidValidator
      .validate({ cid })
      .then(() => {
        setCidError("")
        checkForDuplicateCid()
      })
      .catch((e: Error) => {
        setCidError(e.message)
      })
  }, [checkForDuplicateCid])
```

packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
```diff
@@ -63,97 +75,136 @@ interface IAddCIDModuleProps {
...
-                loading={accessingCID}
-                data-cy="button-submit-pin"
-              >
-                <Trans>Pin</Trans>
-              </Button>
-            </Grid>
-          </div>
-        </Form>
-      </Formik>
+              <Trans>Warning: cid already pinned</Trans>
```
FSM1:
```suggestion
              <Trans>Warning: CID already pinned</Trans>
```

FSM1 (approved):
Just a few cosmetic changes, but working great otherwise 👍🏼

packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
Tbaut: :+1:

packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
Tbaut: refactored that part :)

Tbaut:
Thanks Tanmoy and Mike. My logic was a little meh for the check. I re-implemented that in a nicer way. The problem you mentioned @tanmoyAtb is now fixed, great catch.

Tbaut:
BTW I agree with the modal style that's a little weird. I'll change that when adding a name together with https://github.com/ChainSafe/ui-monorepo/issues/2022

tanmoyAtb (approved):
Looking 💯

PR #2031, opened 2022-03-21
[Storage] Restricted mode banner and flow

### LINKED ISSUE #2020: [Storage] Add support for restricted accounts
Implement restricted account UI elements (banner and disable uploads)

Tbaut:
closes #2020

---
Submission checklist: 

   

#### Layout

### DIFF
--- a/packages/storage-ui/src/Components/Pages/BucketPage.tsx
+++ b/packages/storage-ui/src/Components/Pages/BucketPage.tsx
@@ -22,7 +22,7 @@ import { DISMISSED_SURVEY_KEY } from "../Modules/SurveyBanner"
 
 const BucketPage: React.FC<IFileBrowserModuleProps> = () => {
   const { storageBuckets, uploadFiles, uploadsInProgress, getStorageSummary, downloadFile } = useStorage()
-  const { storageApiClient } = useStorageApi()
+  const { storageApiClient, accountRestricted } = useStorageApi()
   const { addToast } = useToasts()
   const [loadingCurrentPath, setLoadingCurrentPath] = useState(false)
   const [pathContents, setPathContents] = useState<FileSystemItem[]>([])
@@ -160,9 +160,19 @@ const BucketPage: React.FC<IFileBrowserModuleProps> = () => {
 
   const handleUploadOnDrop = useCallback(async (files: File[], fileItems: DataTransferItemList, path: string) => {
     if (!bucket) return
+
+    if (accountRestricted) {
+      addToast({
+        type:"error",
+        title: t`Uploads disabled`,
+        subtitle: t`Oops! You need to pay for this month to upload more content.`
+      })
+      return
+    }
+
     let hasFolder = false
     for (let i = 0; i < files.length; i++) {
-      if (fileItems[i].webkitGetAsEntry().isDirectory) {
+      if (fileItems[i].webkitGetAsEntry()?.isDirectory) {
         hasFolder = true
       }
     }
@@ -176,7 +186,7 @@ const BucketPage: React.FC<IFileBrowserModuleProps> = () => {
         .then(() => refreshContents())
         .catch(console.error)
     }
-  }, [addToast, uploadFiles, bucket, refreshContents])
+  }, [bucket, accountRestricted, addToast, uploadFiles, refreshContents])
 
   const viewFolder = useCallback((toView: ISelectedFile) => {
     const fileSystemItem = pathContents.find(f => f.cid === toView.cid && f.name === toView.name)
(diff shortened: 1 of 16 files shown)


packages/storage-ui/src/Contexts/StorageApiContext.tsx
```diff
@@ -228,6 +230,22 @@ const StorageApiProvider = ({ apiUrl, withLocalStorage = true, children }: Stora
...
+    if (accessToken && accessToken.token && storageApiClient) {
+      storageApiClient?.setToken(accessToken.token)
+      const decodedAccessToken = jwtDecode<{ perm: { secured?: string; files?: string } }>(
+        accessToken.token
+      )
+
+      if (decodedAccessToken.perm.files === "restricted") {
+        setAccountRestricted(true)
+      } else {
+        setAccountRestricted(false)
```
Tbaut, line 244: FYI @asnaith, change this to `true` to test, and your account will be in restricted mode.

packages/storage-ui/src/Components/Pages/BucketPage.tsx
```diff
@@ -160,9 +160,19 @@ const BucketPage: React.FC<IFileBrowserModuleProps> = () => {
 
   const handleUploadOnDrop = useCallback(async (files: File[], fileItems: DataTransferItemList, path: string) => {
     if (!bucket) return
+
+    if (accountRestricted) {
+      addToast({
+        type:"error",
+        title: t`Uploads disabled`,
+        subtitle: t`Oops! You need to pay for this month to upload more content.`
```
tanmoyAtb: Something in the line of "Please clear due payments to upload more content" like what we have in the restricted banner sounds more appropriate to me.

tanmoyAtb (approved):
Looks great :)

packages/storage-ui/src/Components/Pages/BucketPage.tsx
Tbaut: What do you think about
```suggestion
        subtitle: t`Your account is restricted. Until you&apos;ve settled up, you can&apos;t upload any new content`
```

tanmoyAtb: Yup. This is better.

FSM1: Can you apply the same update to Files while you're at it?

packages/storage-ui/src/Contexts/StorageApiContext.tsx
asnaith: @Tbaut  awesome. Thank you for this tip 👍

asnaith:
With the modification of the code, I was able to see the banner and the disabled bucket and pin buttons 👍

The hover color of the button is the same as the banner background. 
Screen Shot 2022-03-21 at 2 34 35 PM

As I am forcing restricted mode through code and not the api response I am unable to test the full functionality and see if the banner disappears and the buttons are re-enabled when a user re-adds a card/upgrades their subscription. 

I'm wondering if there is any additional info displayed to the user on the settings page, informing them of the specific action they need to take? I don't see additional elements on the settings page and expected something to further explain the reason for the restriction / required action but perhaps because I'm using the code mod to test and not genuine api responses for a restricted user etc.

packages/storage-ui/src/Components/Pages/BucketPage.tsx
Tbaut: yup, that was my plan :D

FSM1:
> With the modification of the code, I was able to see the banner and the disabled bucket and pin buttons 👍
> 
> The hover color of the button is the same as the banner background. Screen Shot 2022-03-21 at 2 34 35 PM
> 
> As I am forcing restricted mode through code and not the api response I am unable to test the full functionality and see if the banner disappears and the buttons are re-enabled when a user re-adds a card/upgrades their subscription.
> 
> I'm wondering if there is any additional info displayed to the user on the settings page, informing them of the specific action they need to take? I don't see additional elements on the settings page and expected something to further explain the reason for the restriction / required action but perhaps because I'm using the code mod to test and not genuine api responses for a restricted user etc.

There is currently no additional info displayed. The user would see an unpaid invoice on the `settings/billing` and I feel that this should be enough for now.

Tbaut:
@asnaith re the hover background, the fact that the text remains white, and therefor visible makes me this that it isn't really a problem :shrug: Let me know your thoughts.

Agreed with Mike that the unpaid invoice (in red) that is visible when you've something to pay, should be enough.

FSM1 (approved):

asnaith:
> @asnaith re the hover background, the fact that the text remains white, and therefor visible makes me this that it isn't really a problem 🤷 Let me know your thoughts.
> 

@Tbaut yeah it's extremely minor and agree it's not a problem at all as it stays visible and functions as intended. I was just pointing it out in case we wanted to add that extra layer of polish to the presentation (I assumed changing it would be as simple as just changing a color hex).

PR #2048, opened 2022-03-25
Update CID ui tests

### LINKED ISSUE #2044: Update CID tests
- Update CID tests to use unique sessions and prevent deletion lag from causing test conflicts 
- Update the CID modal to follow the same pattern we transitioned to in Files (define modals as their own test objects instead of members of a page).
- Add test coverage for dupe CID warning
- Add test coverage for adding a name to CID

asnaith:
closes #2044


packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
```diff
@@ -135,7 +135,7 @@ const AddCIDModal = ({ modalOpen = false, close }: IAddCIDModuleProps) => {
   }, [])
 
   const onNameChange = useCallback((name?: string | number) => {
-    setName(name?.toString().trim() || "")
+    setName(name?.toString() || "")
```
asnaith: Attempted a fix here - it seemed necessary to avoid spaces in a CID name from getting trimmed (I noticed "cute cat" was saved as "cutecat")

Tbaut: Ah nice catch thanks, this is indeed the problem, but we need to trim the name before the upload, so that we don't send to the server any useless space at the end of the name. I'll add that to the relevant function.

packages/storage-ui/src/Components/Elements/CidRow.tsx
```diff
@@ -80,7 +80,8 @@ const CidRow = ({ pinStatus }: Props) => {
     >
       <TableCell
         className={classes.cid}
-        align='left'>
+        align='left'
+        data-cy="cell-pin-name">
```
Tbaut:
```suggestion
        data-cy="cell-pin-name"
      >
```

packages/storage-ui/src/Components/Modules/AddCIDModal.tsx
Tbaut: done

Tbaut (approved):
:rocket:

FSM1 (approved):

Issue #2051, opened 2022-03-28
the new file name when renaming can't be selected


Tbaut:
How to reproduce:
- rename a file
- try to put the cursor at another place than the end with the mouse, or select part of the text: you can't

Expected behavior:
- The renaming field is focused
- I can change part of the name by navigating with the mouse

Tbaut:
This is on Firefox BTW

PR #2054, opened 2022-03-29
File rename refactor

### LINKED ISSUE #2050: Click outside should save a renaming action
Based on me shadowing a Files user, and checking the behavior of a normal OS (windows or linux but I bet mac is the same).
When you click out after renaming a file, it actually saves it. On files it cancels the renaming action.

Tbaut:
closes #2050

Since the edition isn't exactly instant even with good internet, I find it good to have a spinner to show something is happening

---
Submission checklist: 

   

#### Layout

#### Theme


asnaith:
I think this feels more natural 👍. 

There is no way for a user to cancel once they have made their change now though. 

We also need to apply the same change to the Shared page, when renaming a Shared Folder it did not save when I clicked outside the input. 

It would be good to apply the same change to files within buckets on Storage UI too (or create a seperate issue), what do you think? 

I wasn't able to see the spinner when I throttled the connection in the dev tools.

Tbaut:
> There is no way for a user to cancel once they have made their change now though.

There is by pressing the `esc` key, like on an os.

Thanks for checking it out, I'll apply these changes to the other areas where we rename stuff.

Tbaut:
Storage needed some small refactor. This is being a big PR now. I tried to test all cases, and couldn't spot anything missing, feel free to play around and rename anything you can rename :)

I did not change the mobile way with a loader on the modal, just to keep things from exploding totally.

asnaith:
Just checked out the latest changes. Share folder renaming in Files UI and file renaming in Storage UI buckets are both good now. 

I did notice on storage ui that we're not doing the same thing with file extensions when we're renaming so I'll create a separate issue as it would be good to have them behaving the same. 

I realized I was seeing the spinner before but in dark mode it's really difficult to see: 

Screen Shot 2022-03-30 at 2 55 00 PM

tanmoyAtb:
The implementation is really great, here are some thoughts that came into my mind.
These are really borderline nitpicking though. 

1. on hitting rename, the editing field should get in focus (I start typing right after hitting rename).
2. If you clear the field, and click outside, this should cancel the rename operation (This is common OS behaviour as well, sth I noticed when I didn't want to hit `esc`).

Tbaut:
@tanmoyAtb 
> the editing field should get in focus

Agreed, I looked for this, the `autofocus` is set properly, but there's generally a focus problem linked to the drag and drop IMHO, so formik fails at focusing the field. We should double check this with https://github.com/ChainSafe/ui-monorepo/issues/2051


> If you clear the field, and click outside, this should cancel the rename operation (This is common OS behaviour as well, sth I noticed when I didn't want to hit esc).

Ah good point! Definitely something I can do


@asnaith I hadn't tested dark mode yet (and didn't check the box in the first post :D ), it's fixed now :)
image

asnaith:
> If you clear the field, and click outside, this should cancel the rename operation (This is common OS behaviour as well, sth I noticed when I didn't want to hit esc).

>>Ah good point! Definitely something I can do

This seems implemented already :) 

https://user-images.githubusercontent.com/3982665/161147068-72c4fb35-40ce-4c80-9c22-23f3031178a1.mov

> @asnaith I hadn't tested dark mode yet (and didn't check the box in the first post :D ), it's fixed now :)

That's why I switched to dark mode straight away 😂. Looks great now.

asnaith (approved):

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemTableItem.tsx
```diff
@@ -178,10 +181,14 @@ const FileSystemTableItem = React.forwardRef(
       initialValues: { name: fileName },
       validationSchema: nameValidator,
       onSubmit: (values: { name: string }) => {
+
         const newName = extension !== "" ? `${values.name.trim()}.${extension}` : values.name.trim()
 
-        if (newName !== name) {
-          newName && handleRename && handleRename(file.cid, newName)
+        if (newName !== name && !!newName && handleRename) {
```
Tbaut, line 187: yup that's in the code, if `newName` is "" it'll be falsy and go to the `else` and stop editing.

Tbaut:
Thanks Andrew for checking, it's indeed already supported :)

tanmoyAtb (approved):
The rename functionality feels fully intuitive now. 💯

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemGridItem.tsx
```diff
@@ -291,7 +297,13 @@ const FileSystemGridItem = React.forwardRef(
                 </Form>
               </FormikProvider>
             )
-            : <div className={classes.gridFolderName}>{name}</div>
+            : <div className={classes.gridFolderName}>
+              {name}{!isEditingLoading && <Loading
```
tanmoyAtb: caught a small glitch
```suggestion
              {name}{isEditingLoading && <Loading
```

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemGridItem.tsx
```diff
@@ -196,8 +199,11 @@ const FileSystemGridItem = React.forwardRef(
       onSubmit: (values: { name: string }) => {
         const newName = extension !== "" ? `${values.name.trim()}.${extension}` : values.name.trim()
 
-        if (newName !== name) {
-          newName && handleRename && handleRename(file.cid, newName)
+        if (newName !== name && newName && handleRename) {
```
Tbaut: To align with the rest
```suggestion
        if (newName !== name && !!newName && handleRename) {
```

PR #2056, opened 2022-03-30
stage url update

asnaith:
Replace references of `https://stage.imploy.site/api/v1` with `https://stage-api.chainsafe.io/api/v1` + update tests


asnaith:
I noticed after the URL change that I had to recreate some accounts to fix some failing tests, the accounts being used for lookups no longer existed and none of the share links were valid either. I'm not sure if that was expected after the url switch so I'm leaving this as a draft, maybe this data update was unnecessary?

There are also some legitimate test fails still happening due to the Max plan not being returned in `/products` atm and most of the subscription-related tests are expecting it.

asnaith:
Hmmm I updated the account being used for the storage tests (thought this was the issue) and it passes locally for me but it's still failing in CI. Will look into this more tomorrow. 

Bucket management -- desktop -- can create a bucket (failed)

Tbaut (approved):
Thanks, having some cors error now, but the changes look good.

FSM1 (approved):

tanmoyAtb (approved):

Tbaut:
There was an env set for the github action with the old url. I removed this and it should be fine now.

Tbaut:
many tests are failing.. can't tell why, it seems random..

asnaith:
> many tests are failing.. can't tell why, it seems random..

@Tbaut  Yeah I think it's maybe because the data update - seems like original data was put back since I made the changes and now the new data I created yesterday was wiped. I will try reverting those changes and re-test soon.

asnaith:
@FSM1 @Tbaut @tanmoyAtb As expected, I had to revert the data changes I made because the old data returned and the new stuff I had created no longer existed 🙃

The only failures happening now in this branch are related to issues raised on our channel 

- file uploading failing (causing all but one of the fails)
- invoice response is blank due to missing key in header (don't believe this is something from our side). 

Hopefully, once those issues are fixed we can merge this into dev, and then we can update all of the other PRs to bring in these changes.

Tbaut:
Tbh I'd merge. This will be one pr less. If there's anything we can do later, let's do it in another one.

asnaith:
> Tbh I'd merge. This will be one pr less. If there's anything we can do later, let's do it in another one.

Sounds good :)

PR #2067, opened 2022-04-01
Prevent drag when rename

### LINKED ISSUE #2051: the new file name when renaming can't be selected

tanmoyAtb:
closes #2051
Preventing drag events from the textinput works better than dynamically removing refs.


Tbaut (commented):
Unfortunately this doesn't solve the problem I mentioned in the issue. I still can't select some text in the rename input with the mouse (that's only on FF), and the input isn't focuses automatically as it should either.


Idk the drag logic well, but I don't really understand why we'd need to make this draggable, and then cancel anything that's happening, this looks hacky to me (maybe it's not though?)

Could we add this for storage as well

tanmoyAtb:
> Unfortunately this doesn't solve the problem I mentioned in the issue. I still can't select some text in the rename input with the mouse (that's only on FF), and the input isn't focuses automatically as it should either.
> 
> Idk the drag logic well, but I don't really understand why we'd need to make this draggable, and then cancel anything that's happening, this looks hacky to me (maybe it's not though?)
> 
> Could we add this for storage as well

Interesting, 

I found this discussion here, says its a known issue with FF, there are also some interesting hacks - looking into them
[Firefox issue and hacks](https://stackoverflow.com/questions/21680363/prevent-drag-event-to-interfere-with-input-elements-in-firefox-using-html5-drag)

Also the draggable implementation is something I found in the React DnD repo discussions. 
[React DnD discussion](https://github.com/react-dnd/react-dnd/issues/335)

Looking into coming up with a more solid approach to this. 

Also the storage side of DnD is old code, I'm going to create a separate issue to recheck the DnD in storage overall.

asnaith:
This change is a great UX improvement, the only thing I found was the FF issue Thibaut already mentioned. 

I will check it out again when FF improvements are ready.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemItem.tsx
```diff
@@ -409,6 +418,7 @@ const FileSystemItem = ({
     }
   }
 
+
```
tanmoyAtb:
```suggestion
```

tanmoyAtb:
Finally found nice hacks to keep FF happy as well as get the autofocus.

- Text in rename field selectable,
- Rename textfield autofocuses.

Created #2071 to update storage side of DnD

asnaith (approved):
Latest changes look great, working well in firefox now

Tbaut (approved):
Awesome, I added the hack for shared folder renaming as well to get the focus. I don't really know what's going on here with the text input, like why it needs the hack, maybe the fact that it appears only after the page load..

Also I saw that the trick isn't implemented on Storage but I guess it should be part of https://github.com/ChainSafe/ui-monorepo/issues/2071

PR #2072, opened 2022-04-04
Add ui test for blocked uploading when storage exceeded

### LINKED ISSUE #2066: Add new ui test coverage for blocked uploads
Add ui test coverage to ensure a user cannot upload a file if it exceeds their storage capacity. 

Stub the buckets/summary data to set available storage.

asnaith:
closes #2066


packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -47,6 +47,26 @@ describe("File management", () => {
...
+          res.body.total_storage = "107374182400"
+          res.body.used_storage = "107374181400"
+        })
+      })
+
+      cy.web3Login()
+      homePage.uploadButton().click()
+      fileUploadModal.body().attachFile("../fixtures/uploadedFiles/text-file.txt")
+      // we attach files via automation differently than a real user, click is necessary to invoke error
+      fileUploadModal.uploadButton().should("be.enabled").click()
```
Tbaut: somehow this click doesn't happen, or doesn't have the effect we want. The test is failing after this, on the next line

asnaith: Yeah I've noticed some inconsistent behavior after using `attachFile` which is different than what a real user does.

Seems like a timing issue. Sometimes the button is still enabled and requires a click to show the error (a user never sees this behavior) and other times the error is shown and the button is disabled like a user normally sees. Will look further tomorrow.

asnaith:
@Tbaut I noticed I messed up the branch name when I pushed this draft (had been working in a branch with those error messages before they were in `dev`). Should I create a new one or just keep this as is?

Tbaut:
Up to you, but I guess it's OK, don't worry.

packages/files-ui/cypress/tests/file-management-spec.ts
asnaith: I've fixed this now, I wrote a small helper function to add some reliability to this part of the test. A hardcoded wait was necessary 😓

packages/files-ui/cypress/support/commands.ts
```diff
@@ -147,15 +147,18 @@ Cypress.Commands.add(
...
   const click = ($el: JQuery<HTMLElement>) => $el.trigger("click")
 
   return cy
     .wrap($element)
     .should("be.visible")
     .should("be.enabled")
     .pipe(click)
-    .should($el => expect($el).to.not.be.visible)
+    .should($el => expectDisable
+      ?  expect($el).to.be.disabled
```
FSM1:
```suggestion
      ? expect($el).to.be.disabled
```

FSM1 (approved):

tanmoyAtb (approved):
Looks good 👍

PR #2080, opened 2022-04-07
Enable shared folder update button to close the modal

### LINKED ISSUE #2068: Confusing disabled "Update" button when only a link got added/changed to a shared bucket
How to reproduce:
- create or manage a shared bucket
- add a sharing link
- the "Update" button is disabled, and only the "Close" button is enabled.

This is confusing because the user did something (add a link) but it feels like the Update button would make it valid, although it's already valid.

To make this change, we should maybe have a "Done" button enabled anytime something was done, and if it's just a link creation, then it does the same as close. If a user was added, then it actually updates the bucket

image

Tbaut:
closes #2068

When only a link was generated, the Update button will be enabled but will just close the modal.


asnaith (approved):

FSM1 (approved):
Working well, made a couple small changes, mainly ensuring that setTouchedLinksList is called when removing a link.

PR #2084, opened 2022-04-10
DnD on storage

### LINKED ISSUE #2071: DnD updates in storage
These implementations need updates for storage ui

- Files and folders can be dragged into folder rows to move them
- Files and folders can be dragged into breadcrumbs to move them
- Check proper previews are created while dragging
- Renaming should prevent drags and rename textfield should allow text selections (Check FF)

tanmoyAtb:
closes #2071 

---
Submission checklist: 

   

#### Layout

#### Theme

### DIFF
--- a/packages/storage-ui/src/Components/Modules/FilesList/DragPreviewLayer.tsx
+++ b/packages/storage-ui/src/Components/Modules/FilesList/DragPreviewLayer.tsx
@@ -1,0 +1,193 @@
+import { FolderFilledSvg, FileImageSvg, FilePdfSvg, FileTextSvg, Typography } from "@chainsafe/common-components"
+import {  makeStyles } from "@chainsafe/common-theme"
+import clsx from "clsx"
+import React from "react"
+import { useDragLayer, XYCoord } from "react-dnd"
+import { FileSystemItem } from "../../../Contexts/StorageContext"
+import { CSSTheme } from "../../../Themes/types"
+import { DragTypes } from "./DragConstants"
+import { BrowserView } from "../../../Contexts/types"
+import { ISelectedFile } from "../../../Contexts/FileBrowserContext"
+
+const useStyles = makeStyles(({ breakpoints, constants, palette }: CSSTheme) => {
+  return ({
+    rowItem: {
+      display: "flex",
+      height: 70,
+      border: "2px solid transparent"
+    },
+    fileIcon: {
+      "& svg": {
+        width: constants.generalUnit * 2.5,
+        fill: constants.fileSystemItemRow.icon
+      },
+      [breakpoints.up("md")]: {
+        paddingLeft: constants.generalUnit * 8.5
+      },
+      paddingRight: constants.generalUnit * 6
+    },
+    folderIcon: {
+      "& svg": {
+        fill: palette.additional.gray[9]
+      }
+    },
+    filename: {
+      whiteSpace: "nowrap",
+      textOverflow: "ellipsis",
+      overflow: "hidden"
+    },
+    previewDragLayer: {
+      position: "fixed",
+      pointerEvents: "none",
+      zIndex: 100,
+      left: 0,
+      top: 0,
+      bottom: 0,
+      right: 0
+    },
+    gridViewContainer: {
+      display: "flex",
+      flex: 1,
+      maxWidth: constants.generalUnit * 24
+    },
+    gridFolderName: {
+      textAlign: "center",
+      wordBreak: "break-all",
+      overflowWrap: "break-word",
+      padding: constants.generalUnit
+    },
+    gridViewIconNameBox: {
+      display: "flex",
+      flexDirection: "column",
+      width: "100%",
+      cursor: "pointer"
+    },
+    gridIcon: {
+      display: "flex",
+      justifyContent: "center",
+      alignItems: "center",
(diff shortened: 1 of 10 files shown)


asnaith:
Hey Tanmoy, I noticed a few issues in this branch after these changes: 

- The breadcrumb positioning/layout breaks once you start to nest folders. See video / screenshot below: 

https://user-images.githubusercontent.com/3982665/162825402-4e61a1cb-dcab-4a6c-9c22-b218f17df2a2.mov

Screen Shot 2022-04-11 at 1 25 23 PM

- The file browser is only allowing drag and drop on folders if one of them contains a file. If you try and drag and drop an empty folder into another empty folder then it doesn't work (this works on files). 

https://user-images.githubusercontent.com/3982665/162827173-9e318444-c43c-40e4-a8e9-ef1f75ab990e.mov

- I'm not seeing any DND behaviour on the mobile layout

tanmoyAtb:
> Hey Tanmoy, I noticed a few issues in this branch after these changes:
> 
> * The breadcrumb positioning/layout breaks once you start to nest folders. See video / screenshot below:
> 
>  breadcrumb.issues.mov 
> Screen Shot 2022-04-11 at 1 25 23 PM
> 
> * The file browser is only allowing drag and drop on folders if one of them contains a file. If you try and drag and drop an empty folder into another empty folder then it doesn't work (this works on files).
> 
>  empty.folder.issue.mov 
> * I'm not seeing any DND behaviour on the mobile layout

Hey @asnaith, thanks for the check. Interesting stuff,
Can you give me the browser and screen size at which you are testing the breadcrumb breaking ?

asnaith:
> Hey @asnaith, thanks for the check. Interesting stuff, Can you give me the browser and screen size at which you are testing the breadcrumb breaking ?

@tanmoyAtb This was at 1280 x 720 resolution on Chrome / Firefox. 

The easiest way to reproduce this is to create a folder, double click on it, add a new folder inside and repeat. After a couple of times doing that you will see that things start to go awry with the breadcrumb.

tanmoyAtb:
All right, we had a css system missing on storage. It's all updated.
We don't really have DnD for mobile layouts on files either. So we're not going to have it here I think.

asnaith:
Hey @tanmoyAtb this is looking great after the css fix. The only DND issue remaining is the one I mentioned before with blank folders. 

> The file browser is only allowing drag and drop on folders if one of them contains a file. If you try and drag and drop an empty folder into another empty folder then it doesn't work (this works on files).

I'm assuming we'd want to fix it (because it's allowed on Files) but I'm not sure. What do you think?

tanmoyAtb:
> #2071

Hey @asnaith I checked on the empty folder moving functionality. It seems to be working fine on my end. 

https://user-images.githubusercontent.com/10713353/163410722-146c0b12-758f-4a6f-9110-49c7062b5948.mov

tanmoyAtb:
> > #2071
> 
> Hey @asnaith I checked on the empty folder moving functionality. It seems to be working fine on my end.
> 
>  Screen.Recording.2022-04-14.at.8.20.56.PM.mov

I just realized the move doesn't work with IPFS file system. Looking into it.

asnaith:
> > > #2071
> > 
> > 
> > Hey @asnaith I checked on the empty folder moving functionality. It seems to be working fine on my end.
> > Screen.Recording.2022-04-14.at.8.20.56.PM.mov
> 
> I just realized the move doesn't work with IPFS file system. Looking into it.

@tanmoyAtb Oh! Good catch. I hadn't yet realized it was specifically related to bucket type

tanmoyAtb:
> > > > #2071
> > > 
> > > 
> > > Hey @asnaith I checked on the empty folder moving functionality. It seems to be working fine on my end.
> > > Screen.Recording.2022-04-14.at.8.20.56.PM.mov
> > 
> > 
> > I just realized the move doesn't work with IPFS file system. Looking into it.
> 
> @tanmoyAtb Oh! Good catch. I hadn't yet realized it was specifically related to bucket type

All right. It should be good to go now.

asnaith:
> All right. It should be good to go now.

👍 Working great for both bucket types now

asnaith (approved):

packages/storage-ui/src/Components/Modules/FilesList/DragPreviewLayer.tsx
```diff
@@ -0,0 +1,193 @@
...
+            let Icon
+            if (previewItem.isFolder) {
+              Icon = FolderFilledSvg
+            } else if (previewItem.content_type.includes("image")) {
+              Icon = FileImageSvg
+            } else if (previewItem.content_type.includes("pdf")) {
+              Icon = FilePdfSvg
+            } else {
+              Icon = FileTextSvg
+            }
```
FSM1, line 175: Can you not re-use the existing `getIcon` logic here instead?

FSM1 (approved):
Working really well. I left one quick suggestion, that will make for neater code, but besides that this is good to go.

packages/storage-ui/src/Components/Modules/FilesList/DragPreviewLayer.tsx
tanmoyAtb: all updated !

PR #2086, opened 2022-04-11
set default bucket filesystem

FSM1:
Sets default file system to Chainsafe, when creating a new bucket.


asnaith (approved):

tanmoyAtb (approved):
Definitely should be this way. 👍

PR #2106, opened 2022-04-19
Fix double slash when uploading a file

Tbaut:
I feel we've seen this issue before, reported by Andrew. As I was debugging something else, I found out why it happens.
How to reproduce:
- click on upload
- drag a folder with files 
- check the POST upload path and see that the path is `//some dir/some-file` with a double slash

This should fix it. 

@asnaith although I think this is a pretty safe change IMHO, this helper function is used a lot, and it is potentially breaking things. I've spent quite some time to test things, it's used in the preview overview, move files, create directories etc..


FSM1 (approved):
Seems to be working well. Makes more sense now that we have figured out the root cause 🎉

tanmoyAtb (approved):
There was definitely discussions around the double slash before.
Working well here.

asnaith:
@Tbaut nice work on pinpointing the true cause of this! Great to get it fixed 🔥

PR #2109, opened 2022-04-25
shift key functionality for selections

### LINKED ISSUE #2094: When holding the `SHIFT` key, the user should be able to select all files within a range.
When selecting files in any kind of file explorer, I have grown accustomed to being able to use the `CTRL` key to select particular files in no specific order, and the `SHIFT` key to select files within a certain frame.  However, when holding the `SHIFT` key, nothing happens.

When holding the `SHIFT` key while selecting files, the selection of files should be extended from the farthermost file to the selected files.

tanmoyAtb:
closes #2094

I've also made small adjustments to click and `ctrl+click`,
Please see if the behaviour feels intuitive.

Submission checklist: 
No UI change

### DIFF
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
@@ -527,7 +527,6 @@ const FilesList = ({ isShared = false }: Props) => {
       const mySelectedItems = selectedItems
       while (countIndex !== currentIndex) {
         // filter out if item already selected
-        console.log(countIndex)
         const currentCID = items[countIndex].cid
         mySelectedItems.filter((s) => s.cid !== currentCID)
         mySelectedItems.push(items[countIndex])
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
@@ -510,10 +510,10 @@ const FilesList = ({ isShared = false }: Props) => {
 
       const lastIndex = selectedItems.length
         ? items.findIndex((i) => i.cid === selectedItems[selectedItems.length - 1].cid)
-        : undefined
+        : -1
 
       // first item
-      if (!lastIndex) {
+      if (lastIndex === -1) {
         setSelectedItems([item])
         return
       }
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
@@ -483,13 +483,9 @@ const FilesList = ({ isShared = false }: Props) => {
   // Selection logic
   const handleSelectItem = useCallback(
     (item: FileSystemItemType) => {
-      if (selectedCids.includes(item.cid)) {
-        setSelectedItems([])
-      } else {
-        setSelectedItems([item])
-      }
+      setSelectedItems([item])
     },
-    [selectedCids]
+    []
   )
 
   const handleAddToSelectedItems = useCallback(
@@ -507,16 +503,42 @@ const FilesList = ({ isShared = false }: Props) => {
 
   // select item with SHIFT pressed
   const handleSelectItemWithShift = useCallback(
-    (itemToAdd: FileSystemItemType) => {
-      if (selectedCids.includes(itemToAdd.cid)) {
-        setSelectedItems(
-          selectedItems.filter((selectedItem: FileSystemItemType) => selectedItem.cid !== itemToAdd.cid)
-        )
-      } else {
-        setSelectedItems([...selectedItems, itemToAdd])
+    (item: FileSystemItemType) => {
+      // item already selected
+      const isItemAlreadySelected = selectedItems.find((s) => s.cid === item.cid)
+      if (isItemAlreadySelected) return
+
+      const lastIndex = selectedItems.length
+        ? items.findIndex((i) => i.cid === selectedItems[selectedItems.length - 1].cid)
+        : undefined
+
+      // first item
+      if (!lastIndex) {
+        setSelectedItems([item])
+        return
       }
+
+      const currentIndex = items.findIndex((i) => i.cid === item.cid)
+      // unavailable item
+      if (currentIndex === -1) return
+
+      // new item, with selected items 
+      let countIndex = lastIndex
+      const mySelectedItems = selectedItems
+      while (countIndex !== currentIndex) {
+        // filter out if item already selected
+        console.log(countIndex)
+        const currentCID = items[countIndex].cid
+        mySelectedItems.filter((s) => s.cid !== currentCID)
+        mySelectedItems.push(items[countIndex])
+        if (currentIndex > lastIndex) countIndex++
+        else countIndex--
+      }
+      // add the current item
+      mySelectedItems.push(items[currentIndex])
+      setSelectedItems([...mySelectedItems])
     },
-    [selectedCids, selectedItems]
+    [selectedItems, items]
   )
 
   const toggleAll = useCallback(() => {
--- a/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
+++ b/packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
@@ -505,7 +505,7 @@ const FilesList = ({ isShared = false }: Props) => {
   const handleSelectItemWithShift = useCallback(
     (item: FileSystemItemType) => {
       // item already selected
-      const isItemAlreadySelected = selectedItems.find((s) => s.cid === item.cid)
+      const isItemAlreadySelected = selectedItems.includes(item)
       if (isItemAlreadySelected) return
 
       const lastIndex = selectedItems.length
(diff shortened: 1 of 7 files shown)


asnaith:
@tanmoyAtb I really like this improvement. Definitely feels nice and intuitive (command + shift on mac). Working the same as macOS Finder now 🚀

juans-chainsafe:
@tanmoyAtb it works perfect! I tested on storage and files, the issue is fixed.
Also, I rerun the failed tests in cypress because there was an issue with the API when you try to upload a file.

Just one comment: the `ctrl+click` (cmd+shift or cmd+click on mac) is working in Chrome and Safari nicely, but i**s not working on Firefox**

juans-chainsafe (approved):

juans-chainsafe (changes requested):
ctrl+click (cmd+shift or cmd+click on mac) is working in Chrome and Safari nicely, but is not working on Firefox

Tbaut:
indeed, I tried it locally on FF and it didn't work.

tanmoyAtb:
Hey guys, I tried out both `cmd+click` and `shift+click` with firefox with the render links. Seems to be working my side. Here's a screen recording.

https://user-images.githubusercontent.com/10713353/166635790-6922593e-7c20-4a13-af23-366bb0dd4a46.mov

Tbaut:
checked the render link, didn't work for me :( I'll check it out a little

FSM1 (approved):
lgtm 🎉 also tested (the render link) in FF and it seems to be working just fine there as well 🤷🏼

juans-chainsafe:
I tried again but it doesn't work for me in Firefox, not sure why

https://user-images.githubusercontent.com/104365141/166678510-a1dff579-85c6-4497-af1d-17bf7aebfd2e.mov

Tbaut:
I figured what was the problem. Juan and I actually click on the checkboxes, not on the file themselves, in that case we don't use the same handler. This is done in `FileSystemTableItem` (and similar for grid items) and the checkbox only does `handleAddToSelectedItems`.

tanmoyAtb:
> I figured what was the problem. Juan and I actually click on the checkboxes, not on the file themselves, in that case we don't use the same handler. This is done in `FileSystemTableItem` (and similar for grid items) and the checkbox only does `handleAddToSelectedItems`.

cool, so do you think we should expand the checkbox click behaviour to handle `shift`/`cmd` ?

juans-chainsafe:
> > I figured what was the problem. Juan and I actually click on the checkboxes, not on the file themselves, in that case we don't use the same handler. This is done in `FileSystemTableItem` (and similar for grid items) and the checkbox only does `handleAddToSelectedItems`.
> 
> cool, so do you think we should expand the checkbox click behaviour to handle `shift`/`cmd` ?

I think that will be a good update, because as a user I was trying to to do it on the checkbox

tanmoyAtb:
> > > I figured what was the problem. Juan and I actually click on the checkboxes, not on the file themselves, in that case we don't use the same handler. This is done in `FileSystemTableItem` (and similar for grid items) and the checkbox only does `handleAddToSelectedItems`.
> > 
> > 
> > cool, so do you think we should expand the checkbox click behaviour to handle `shift`/`cmd` ?
> 
> I think that will be a good update, because as a user I was trying to to do it on the checkbox

I've made the updates to handle the `shift/cmd/ctrl` behaviour with checkbox clicks 👍

Tbaut:
did you test it on Firefox? it's not working for me. I added some logs in the `handleItemSelectOnCheck` but it looks like it's not called when ctrl or shift are pressed while clicking :(.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -505,6 +501,46 @@ const FilesList = ({ isShared = false }: Props) => {
     [selectedCids, selectedItems]
   )
 
+  // select item with SHIFT pressed
+  const handleSelectItemWithShift = useCallback(
+    (item: FileSystemItemType) => {
+      // item already selected
+      const isItemAlreadySelected = selectedItems.find((s) => s.cid === item.cid)
```
Tbaut: You can use `selectedItems.~~contains~~includes(item.cid)` here.

Tbaut (commented):
Small comment in the code. the code looks good otherwise.
Re. the checkbox annoyance, I'm with Juan, as a user, I'm used to click on the checkboxes, on gmail for instance. I did look at how they do that, and they actually don't use real checkboxes, they use a trick with `div`, anyway it's a whole other ballpark.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
tanmoyAtb: This implementation becomes a problem really, as selectedItems is an array of objects. simply checking the cid match works.

tanmoyAtb:
I found a number of reports of this problem. like this one
https://stackoverflow.com/questions/37803218/javascript-ctrl-click-not-working-in-firefox

FF really becomes an annoyance sometimes. 
For now the implementation we have should be good enough for most cases. We can tackle the FF problem in another issue.

juans-chainsafe (approved):
I see, then for me seems ok to merge it like this, because the functionality is working, but not in the checkbox like in other browsers, but yes, the problem seems to be only in FF

juans-chainsafe:
Issue created https://github.com/ChainSafe/ui-monorepo/issues/2123

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
Tbaut: sorry, I should have said `.includes` instead of contains. It's more elegant than `.find` in this case. 
```suggestion
      const isItemAlreadySelected = selectedItems.includes(item.cid)
```

and then you get a boolean, and you can check for `isItemAlreadySelected` instead of `isItemAlreadySelected.length`

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -505,6 +501,46 @@ const FilesList = ({ isShared = false }: Props) => {
...
+      if (isItemAlreadySelected) return
+
+      const lastIndex = selectedItems.length
```
Tbaut, line 511: here
```suggestion
      const lastIndex = selectedItems
```

Tbaut:
Having 2 different browsers having 2 implementations is rare, and I agree we can tackle this in a new issue, but I wouldn't call it a bug right away. This PR made a change from `onChange` to `onClick`, take a look at https://stackoverflow.com/questions/26925469/on-click-event-in-firefox where the `onChange` event seem to yield the right effect.

A personal opinion, Firefox isn't perfect, but it's an open source and much more privacy preserving software than any of its competitors, and as much as I'm sad to see some things don't work as we want, we should try everything we can to provide the best experience for FF Files users.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
tanmoyAtb: Yup, using the `includes` now.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
tanmoyAtb: not sure about this one. we need to keep the `length` part to sure selectedItems is not empty as we are trying access elements inside it.

tanmoyAtb:
All right. I found a solution to our situation with FF. The problem is mainly handling `checkbox` type inputs, but in our checkbox component we already have a `div` component that is actually displayed (this is probably what was the gmail implementation as well). So using the `onClick` with the `div` element does the trick. :)

packages/storage-ui/src/Components/Modules/FilesList/FilesList.tsx
```diff
@@ -429,6 +425,51 @@ const FilesList = () => {
     [selectedCids]
   )
 
+  // select item with SHIFT pressed
+  const handleSelectItemWithShift = useCallback(
+    (item: ISelectedFile) => {
+      // item already selected
+      const isItemAlreadySelected = selectedItems.find((s) => s.cid === item.cid && s.name === item.name)
```
Tbaut: Given the name, I'd consider making this a boolean
```suggestion
      const isItemAlreadySelected = selectedItems
        .find((s) => s.cid === item.cid && s.name === item.name)
        .length > 0
```

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
Tbaut: sorry my mistake, I mixed this one and `isItemAlreadySelected`.

Tbaut (approved):
This is awesome, thanks a lot for your perseverance, it's working well!


~~I left another small comment, to make a `is***` variable a boolean.~~
edit: already accepted :)

tanmoyAtb:
> This is awesome, thanks a lot for your perseverance, it's working well!
> 
> ~I left another small comment, to make a `is***` variable a boolean.~ edit: already accepted :)

Thanks for pushing me into looking for the solution :) feels really good to have the full solution.

FSM1:
> > This is awesome, thanks a lot for your perseverance, it's working well!
> > ~I left another small comment, to make a `is***` variable a boolean.~ edit: already accepted :)
> 
> Thanks for pushing me into looking for the solution :) feels really good to have the full solution.

Does this PR also close #2123 ?

Tbaut:
Yup :)

PR #2117, opened 2022-05-04
Improve bucket upload ui tests

### LINKED ISSUE #2085: Expand ui tests for storage ui file upload
At the moment, our automated tests are just uploading to whatever bucket type is selected by default. 

We need to expand this current test coverage so that we are uploading to both supported file system types of bucket (ipfs and ChainSafe)

asnaith:
closes #2085 

Our automation tests in this area were pretty basic, only creating and deleting the default bucket. If bucket uploading was broken for a specific file system type then we couldn't detect it. 

These updated tests now provide more robust coverage for the bucket create, upload, and delete flow and utilize both file system types, ChainSafe and IPFS.


packages/storage-ui/cypress/fixtures/storageTestData.ts
```diff
@@ -1,4 +1,5 @@
-export const bucketName = "test bucket"
+export const chainSafeBucketName = `cs bucket ${Date.now()}`
+export const ipfsBucketName = `ipfs bucket ${Date.now()}`
```
asnaith, line 2: We remove the bucket from the UI instantly and don't wait for any confirmation so to avoid issues I'm using unique names here.

packages/storage-ui/cypress/support/page-objects/bucketContentsPage.ts
```diff
@@ -0,0 +1,27 @@
+import { basePage } from "./basePage"
+
+export const bucketContentsPage = {
+  ...basePage,
+
+  // bucket content browser elements
+  bucketHeaderLabel: () => cy.get("[data-cy=header-bucket]", { timeout: 20000 }),
```
juans-chainsafe: this timeout is because this type of web elements take longer to load than usual, right?

packages/storage-ui/src/Components/Modules/UploadFileModal/UploadFileModal.tsx
```diff
@@ -136,7 +136,7 @@ const UploadFileModal = ({ modalOpen, close }: IUploadFileModuleProps) => {
           />
           <footer className={classes.footer}>
             <Button
-              data-cy="upload-cancel-button"
+              testId="cancel-upload"
```
juans-chainsafe, line 139: why testId and not data-cy? this modal es reusable?

juans-chainsafe (approved):
Looks great, you had a lot of work here! awesome

packages/storage-ui/src/Components/Modules/UploadFileModal/UploadFileModal.tsx
asnaith: @juans-chainsafe In this example specifically, the button component already has a testId prop so I'm just passing something through to that, as opposed to writing a custom attribute directly on the element. It's a little cleaner / more readable ("testId" is instantly obvious to all reading the code what it's for). 

Screen Shot 2022-05-04 at 10 06 30 AM

For buttons, either type of locator will work, but there are instances with other components where this is not the case:  

- Due to the manner in which we are setting up the element in code (maybe using the default component without further customizations) it's sometimes hard to add a custom data-cy attribute to it (off the top of my head I think toasts in Files is a good example). 

- I've also seen instances where even though we have set up a data-cy tag Cypress hasn't been able to find it but by setting up a testId on the component and passing something through to it does work (saw this recently with radio buttons). 

I am certain you will encounter this too at some point. Hopefully, this explanation makes sense but we can also discuss it further if you want :)

packages/storage-ui/cypress/support/page-objects/bucketContentsPage.ts
asnaith: Actually...this was unnecessary 😄. It was left there by accident. I have pushed a commit to remove it and some others which aren't needed. 

The default is 6 seconds and that will be ok.

tanmoyAtb (approved):
Looks good to me !

PR #2122, opened 2022-05-09
[Storage] Add sign out test

### LINKED ISSUE #2121: Add sign out test case in Storage
As described, add test coverage for the below scenarios in Storage:

- User can sign out from Base Page Component - Account / Wallet Address Dropdown

juans-chainsafe:
closes #2121

### DIFF
--- a/packages/storage-ui/src/Components/Layouts/AppNav.tsx
+++ b/packages/storage-ui/src/Components/Layouts/AppNav.tsx
@@ -337,6 +337,7 @@ const AppNav: React.FC<IAppNav> = ({ navOpen, setNavOpen }: IAppNav) => {
 
               <div style={{ display: "flex" }}>
                 <Button
+                  data-cy="button-sign-out"
                   variant='secondary'
                   onClick={() => {
                     handleOnClick()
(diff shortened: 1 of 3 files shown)


packages/storage-ui/src/Components/Layouts/AppNav.tsx
```diff
@@ -337,6 +337,7 @@ const AppNav: React.FC<IAppNav> = ({ navOpen, setNavOpen }: IAppNav) => {
 
               <div style={{ display: "flex" }}>
                 <Button
+                  data-cy="button-sign-out"
```
asnaith, line 340: The button component has a testId prop so instead of the custom tag data-cy tag you could also write 

`testId="sign-out"` and your locator would be `[data-testid=button-sign-out]`

It's a minor thing but I think when the component supports testId props it's probably better to use them as it's simpler, cleaner and makes sense to people reading the code for the first time. What do you think? 

PS there may be other examples in the framework where we aren't using the available props as in some cases the component was updated after, we can clean those up as we encounter them.

Tbaut: My 2cts here, `testId` props should only be available for things where the `data-cy` doesn't work. I think we should use the `data-cy` as much as possible. To me, this is a mistake to have a `testId` prop if the data-cy works. This is definitely something we should look into.

Tbaut: what `testId` does is that it passes down a `data-testId` identifier to the underlying html component, if the html component isn't accessible from the code from Files. This is the case for some component from our library, but there are few afaik.

asnaith: Ok! Let's exclusively only use testId when regular data-cy doesn't work then, sorry for any confusion @juans-chainsafe

asnaith (approved):

FSM1 (approved):
lgtm

Issue #1931, opened 2022-02-09
[Storage] NFT metadata creation UI


FSM1:
Implement a dynamic Formik form with a default metadata selector (ERC721/ERC1155) which will control the initial schema provided to the Formik HoC.

The user should be able to modify the schema of the Formik HoC by adding/removing fields by selecting the field type:
- Primitive (string), 
- Object (collection of primitives, files, other objects)
- Files.

Upon submission the resulting Formik HoC `Values` object should be submitted to the `apiClient.uploadNft` method with the resulting CID being displayed to the user.

FSM1:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Files-5eb5ea92d698464ed9543f62/issues/ChainSafe/ui-monorepo/1931?planning-poker) with ZenHub @tanmoyAtb @Tbaut

PR #2125, opened 2022-05-11
NFT Metadata form

FSM1:
This is a very basic MVP of the NFT Metadata creation. The main objective here was to ensure that the basic requirements can be met. The UX can still do with a lot of improvement.

This addresses the initial part of #1931


packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -227,35 +231,52 @@ const FileInput = ({
...
+                  : <li className={clsx(classes.item, classNames?.item)}>
+                    <span className={classes.itemText}>{value.path}</span>
                     <Button
                       testId="remove-from-file-list"
                       className={clsx(classes.crossIcon, classNames?.closeIcon)}
                       type="button"
                       onClick={(e) => {
                         e.stopPropagation()
-                        removeItem(i)
+                        removeItem(1)
```
Tbaut: should we make the removeItem param optional and have no param here?? This `1` disturbed me although it's actually ignored for single File items.

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -227,35 +231,52 @@ const FileInput = ({
...
+                          e.stopPropagation()
+                          removeItem(i)
+                        }}
+                        size="small"
+                      >
+                        <CrossIcon fontSize="small" />
+                      </Button>
+                    </li>
+                  ))
+                  : <li className={clsx(classes.item, classNames?.item)}>
```
Tbaut: I'm generally wary of duplicating the code. it'd be great to extract it in a variable if that's not too annoying. I often had endless loops using useCallback though, we can avoid them I guess.

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -227,35 +231,52 @@ const FileInput = ({
...
-                ))}
+                }
               </ul>
             </ScrollbarWrapper>
           </div>
         )
       ) : (
         <>
-          {value?.length === 0
+          {value === null || Array.isArray(value) && value?.length === 0
```
Tbaut: with parenthesis?
```suggestion
          {value === null || (Array.isArray(value) && value?.length === 0)
```

packages/storage-ui/src/Components/Elements/FormikImageInput.tsx
```diff
@@ -0,0 +1,204 @@
...
+      name,
+      ...rest
+    }: FormikImageInputProps,
+    forwardedRef: any
+  ) => {
+    const classes = useStyles()
+    const [{ value }, meta, helpers] = useField<File>(name)
+
+    const [isDropActive, setIsDropActive] = useState(false)
+    const [preview, setPreview] = useState<any>(undefined)
```
Tbaut: Any way we can define this? Like `Blob | MediaSource | undefined`

packages/storage-ui/src/Components/Elements/FormikImageInput.tsx
```diff
@@ -0,0 +1,204 @@
...
+          {...getRootProps({
+            onClick: event => event.stopPropagation()
+          })}
+          className={clsx(
+            classes.imageContainer,
+            className
+          )}
+          {...rest}
+        >
+          {value ? <>
```
Tbaut: can we have the `?` and `:` at the beginning of their line for consistency?

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+const primitives = ["boolean", "string", "number"]
+
+const ObjectInput: React.FC<{obj: any; namespace: string}> =
+  ({ obj, namespace }) => {
+    const [newFieldName, setNewFieldName] = useState("")
+    const [newFieldType, setNewFieldType] = useState<"value" | "array" | "file" | "object">("value")
+    const formikBag = useFormikContext()
+
+    return <>
+      <div style={{ border: "1px #000 solid" }}>
```
Tbaut: Any reason not to use the usual classes for the style?

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <Typography>{namespace}</Typography><br />
+        {Object.keys(obj).map((key) => generateFormFields(obj[key], `${namespace}.${key}`))}
+        <TextInput
+          value={newFieldName}
+          onChange={(val) => setNewFieldName(val?.toString() || "")}
+          label="Property name" />
```
Tbaut: Trans and new line
```suggestion
          label=t`Property name`
        />
```

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+          }, {
+            label: "File",
+            value: "file"
+          }]}
+          onChange={setNewFieldType}
+          value={newFieldType}
+          label="Property type"/>
+        <Button
+          type="button"
+          onClick={() => formikBag.setFieldValue(`${namespace}.${newFieldName}`, (newFieldType === "array")
```
Tbaut: can we extract this function for better readability

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+    }
+  })
+)
+
+const primitives = ["boolean", "string", "number"]
+
+const ObjectInput: React.FC<{obj: any; namespace: string}> =
+  ({ obj, namespace }) => {
+    const [newFieldName, setNewFieldName] = useState("")
+    const [newFieldType, setNewFieldType] = useState<"value" | "array" | "file" | "object">("value")
```
Tbaut: can we name this type, it's being reused below.

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+      obj={val}
+      namespace={`${namespace}`}
+      key={namespace} />
+  }
+  return undefined
+}
+
+
+
+const CreateNFTPage: React.FC = () => {
```
Tbaut: There are many components in this file, would it be a little easier to read if we had them in separate files?

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -143,11 +143,11 @@ const FileInput = ({
   const classes = useStyles()
   const [previews, setPreviews] = useState<any[]>([])
   const [errors, setErrors] = useState<any[]>([])
-  const [{ value }, meta, helpers] = useField<Array<FileWithPath>>(name)
+  const [{ value }, meta, helpers] = useField<Array<FileWithPath> | FileWithPath | null>(name)
```
Tbaut: just for readability using `FileWithPath[]` could be cool
```suggestion
  const [{ value }, meta, helpers] = useField<FileWithPath[] | FileWithPath | null>(name)
```

packages/storage-ui/src/Components/Layouts/AppNav.tsx
```diff
@@ -278,6 +278,19 @@ const AppNav: React.FC<IAppNav> = ({ navOpen, setNavOpen }: IAppNav) => {
...
+              <Link
+                data-cy="create-nft-nav"
+                onClick={handleOnClick}
+                className={clsx(classes.navItem, appNavTab === "cids" && "selected")}
+                to={ROUTE_LINKS.UploadNFT}
+              >
+                <DatabaseSvg />
+                <Typography
+                  variant="h5"
+                >
```
Tbaut:
```suggestion
                <Typography variant="h5">
```

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+          }, {
+            label: "Array",
+            value: "array"
+          }, {
+            label: "File",
+            value: "file"
+          }]}
+          onChange={setNewFieldType}
+          value={newFieldType}
+          label="Property type"/>
```
Tbaut:
```suggestion
          label="Property type"
        />
```

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const { storageApiClient } = useStorageApi()
+
+  const formikProps = useFormik({
+    initialValues: initialValues,
+    // validationSchema={validationSchema}
+    onSubmit: async (val) => {
+      console.log("submitting")
+      console.log(val)
+      const result = await storageApiClient.uploadNFT(val)
+      console.log(result)
```
Tbaut:
```suggestion
      const result = await storageApiClient.uploadNFT(val)
```

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
```diff
@@ -0,0 +1,220 @@
...
+      </div>
+      {desktop && <Divider />}
+      <FormikProvider value={formikProps}>
+        <Form>
+          <FormikTextInput name='name' />
+          <FormikTextInput name='description' />
+          <FormikTextInput name='decimals' />
+          <FormikImageInput name='image' />
+          {generateFormFields(formikProps.values.properties, "properties")}
+          <Button type="submit">Upload</Button>
```
Tbaut:
```suggestion
          <Button type="submit"><Trans>Upload</Trans></Button>
```

Tbaut (commented):
Nice!! I left a couple comments in the code, I'll play with it now :)

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
FSM1: leaving these in here for now

packages/storage-ui/src/Components/Pages/UploadNFT.tsx
FSM1: Mostly just convenience factor while deving. Will be removed once proper styling is applied

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -143,11 +143,11 @@ const FileInput = ({
   const classes = useStyles()
   const [previews, setPreviews] = useState<any[]>([])
   const [errors, setErrors] = useState<any[]>([])
-  const [{ value }, meta, helpers] = useField<Array<FileWithPath>>(name)
+  const [{ value }, meta, helpers] = useField<FileWithPath[] | FileWithPath | null>(name)
 
   useEffect(() => {
-    onFileNumberChange && onFileNumberChange(value.length)
-  }, [onFileNumberChange, value.length])
+    onFileNumberChange && Array.isArray(value) && onFileNumberChange(value.length)
```
tanmoyAtb: when `value` is of type `FileWithPath`, this function won't be triggered, is that a problem ?

packages/storage-ui/src/Components/Elements/FormikImageInput.tsx
```diff
@@ -0,0 +1,204 @@
...
+    const [isDropActive, setIsDropActive] = useState(false)
+    const [preview, setPreview] = useState<any>(undefined)
+
+    useEffect(() => {
+      value && setPreview(URL.createObjectURL(value))
+
+      return () => {
+        URL.revokeObjectURL(preview)
+      }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tanmoyAtb: We could place the `return` part in another useEffect to avoid this eslint disabled

packages/storage-ui/src/Components/Layouts/AppNav.tsx
```diff
@@ -278,6 +278,17 @@ const AppNav: React.FC<IAppNav> = ({ navOpen, setNavOpen }: IAppNav) => {
                   <Trans>CIDs</Trans>
                 </Typography>
               </Link>
+              <Link
+                data-cy="create-nft-nav"
+                onClick={handleOnClick}
+                className={clsx(classes.navItem, appNavTab === "cids" && "selected")}
```
tanmoyAtb: `appNavTab` should be something like `uploadNFT` here

tanmoyAtb (changes requested):
Leaving a couple of comments here. 
Looking into the UI.

packages/common-components/src/FileInput/FileInput.tsx
FSM1: Are you referring to the `<li>...</li>` ?

Tbaut: yes, unless I missed something the `<li>` is the same on both branches of the ternary, just the `map` and the click differ.

packages/common-components/src/FileInput/FileInput.tsx
FSM1: I don't think so really. I honestly don't even remember why we added this handler rather than having the parent deal with this by interpreting the `value` that is being propagated through formik.

Tbaut (commented):
Played with it locally and all tests passed but one. (the one that's marked as only right now)

> files must be a `array` type, but the final value was: `null`. If "null" is intended as an empty value be sure to mark the schema as `.nullable()`

So I fixed this. The `yarn` issue was pretty weird, but I'll check it out more in detail if it keeps failing.

edit: I really think that's a red herring for a problem somewhere else. With a `git --version` it also fails to find git. But I don't think this is the origin of the problem, why would it appear now anyway.. see https://lifesaver.codes/answer/error-couldn-t-find-the-binary-git-2083

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -139,11 +169,15 @@ const FileInput = ({
   const classes = useStyles()
   const [previews, setPreviews] = useState<any[]>([])
   const [errors, setErrors] = useState<any[]>([])
-  const [{ value }, meta, helpers] = useField<FileWithPath[]>(name)
+  const [{ value }, meta, helpers] = useField<FileWithPath[] | FileWithPath | null>(name)
```
Tbaut, line 172: Is there a good reason to do this? I bet there is. This is messing up with the formik `schema` for file upload that expect an array (and then don't show an error any more because there's now a `null` or even a single file possibility). Also our `onSubmit` all expect an array. Should we modify them all? I find an array much easier to handle, it'd be nice if we could keep that.

FSM1: I needed to use a different `value` for an empty array property, that would be detectable at runtime. This is where `null` came in.

tanmoyAtb (approved):
Lets go with it 👍

Tbaut (approved):
uuuh yes

PR #2129, opened 2022-05-12
Extend timeout to improve storage test reliability

asnaith:
Just a quick one liner -

 I realized I was overzealous in another PR last week when I removed some timeouts. Re-adding this one as it caused a reliability issue with storage tests in CI (where network speed can be slower).


juans-chainsafe (approved):
Seems that the tests in Storage were fixed by this change. Apparently the lingui/lint errors are not from this change

Tbaut (approved):
Sure, it can't hurt

PR #2131, opened 2022-05-16
fix formik checkbox types

FSM1:


Tbaut (approved):
Looks harmless, I didn't test though and we'll find out quickly if it doesn't work

tanmoyAtb (approved):

PR #2134, opened 2022-05-17
Right click context menu

### LINKED ISSUE #2095: Right-clicking on the page should serve the user with a custom menu, rather than browser page settings.
Computer users are especially accustomed to working with menus shown after right-clicking.  In a file explorer, the user -- after right-clicking -- is likely to expect a menu to show up with a concise selection of functions which are usually also available elsewhere in the application.  However, currently, right-clicking causes the standard browser page manipulation menu to show up.

When right-clicking on the page, a menu themed similarly to the rest of the application should be shown to the user, allowing them to select from a concise menu of options.

tanmoyAtb:
closes #2095 

Some notes: 
We have a right click(context menu) implementation here. The overall menu system took a refactor. I really wanted to use the MaterialMenu we already had, but for some reason the material menu just does not fire on the second right click. 🤷 I toyed around with it, and then decided to write a simple `AnchorMenu`. some of the behaviour is inspired from drive and dropbox. 

~~I didn't add this to storage. If we feel like it should be there, I'll add that in.~~
Added to storage

---
Submission checklist: 

   

#### Layout

#### Theme


juans-chainsafe (approved):
Amazing job! is working as expected, I tested as well in mobile of course in there you don't have the right click but I checked that all is working correctly there.

PS: I think it will be a good idea to add this to storage as well, maybe not in this ticket but I can create another if you agree

tanmoyAtb:
I'm going to add this to storage and reopen this PR for review.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/itemOperations.tsx
```diff
@@ -0,0 +1,175 @@
...
+  ZoomInSvg,
+  EyeSvg,
+  ExclamationCircleInverseSvg
+} from "@chainsafe/common-components"
+import { t, Trans } from "@lingui/macro"
+import { FileOperation } from "../../types"
+import { FileSystemItem } from "../../../../../Contexts/FilesContext"
+import { getPathWithFile } from "../../../../../Utils/pathUtils"
+
+export function getItemMenuOptions(
```
Tbaut: Can we have this in an object? I think it's better when we have so many options. The functions can also be mentionned directly instead of having a separate `function` object.

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -851,10 +900,90 @@ const FilesList = ({ isShared = false }: Props) => {
     selectionContainsAFolder
   ])
 
-  const onShare = useCallback((fileSystemItem: FileSystemItemType) => {
-    setSelectedItems([fileSystemItem])
-    handleOpenShareDialog()
-  }, [handleOpenShareDialog])
+  const [contextMenuPosition, setContextMenuPosition] = useState<{
```
Tbaut: Can we extract this into a named interface?

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -876,6 +1006,19 @@ const FilesList = ({ isShared = false }: Props) => {
           <Trans>Drop to upload files</Trans>
         </Typography>
       </div>
+      <AnchorMenu
+        options={contextMenuOptions}
+        onClose={() => setContextMenuPosition(null)}
+        isOpen={!!contextMenuPosition}
+        anchorPosition={
```
Tbaut: To keep things clean and tidy we could have `const anchorPosition = useMemo(() => contextMenuPosition && {top:..., left: ...}, [])` above, and just use anchorPosition={anchorPosition} here.

Tbaut (commented):
Left some minor comments in the code. The refactor with the `getItemOperations` brings some clarity IMHO which is great, although it's a little messy with the million params :man_shrugging: .

The only main problem I see is that because we don't use the "clever" menu, we end up having an unusable right click menu when we are close to the bottom of the page, and scrolling doesn't help.

https://user-images.githubusercontent.com/33178835/169258561-304e2fda-334b-4e75-a682-f64f707ed587.mp4

tanmoyAtb:
> Left some minor comments in the code. The refactor with the `getItemOperations` brings some clarity IMHO which is great, although it's a little messy with the million params 🤷‍♂️ .
> 
> The only main problem I see is that because we don't use the "clever" menu, we end up having an unusable right click menu when we are close to the bottom of the page, and scrolling doesn't help.
> 
>  bottom.mp4

I have been able to make the `AnchorMenu` we have a little more clever. It should position better at the bottom and right sides now. Also, prevented scroll when menu is open (this is what the material menu does as well).

Cleaned up the item operations more and also added to storage. :D

packages/files-ui/src/Components/Modules/FileBrowsers/views/FileSystemItem/FileSystemItem.tsx
```diff
@@ -176,164 +167,46 @@ const FileSystemItem = ({
...
   const classes = useStyles()
-  const filePath = useMemo(() => getPathWithFile(currentPath, name), [currentPath, name])
 
-  const onFilePreview = useCallback(() => {
-    showPreview && showPreview(files.indexOf(file))
-  }, [file, files, showPreview])
+  const menuItems = useMemo(() => getItemMenuOptions({
+    menuIconClass: classes.menuIcon,
+    file: file,
+    inSharedFolder: inSharedFolder,
```
Tbaut:
```suggestion
    file,
    inSharedFolder,
```

packages/storage-ui/src/Components/Modules/FileSystemItem/itemOperations.tsx
```diff
@@ -0,0 +1,152 @@
...
+    file,
+    viewFolder,
+    editFile,
+    deleteFile,
+    downloadFile,
+    moveFile,
+    previewFile,
+    showFileInfo,
+    itemOperations
+  } = params
```
Tbaut: Nit, making the interface explicit and restructuring the params right away.
```suggestion

interface MenuOptionsParams {
  menuIconClass: string
  file: FileSystemItem
  inSharedFolder: boolean
  itemOperations: FileOperation[]
  editFile?: (item: FileSystemItem) => void
  deleteFile?: (item: FileSystemItem) => void
  downloadFile?: (item: FileSystemItem) => void
  moveFile?: (item: FileSystemItem) => void
  handleShare?: (item: FileSystemItem) => void
  recoverFile?: (item: FileSystemItem) => void
  previewFile?: (item: FileSystemItem) => void
  showFileInfo?: (item: FileSystemItem) => void
  reportFile?: (item: FileSystemItem) => void
  viewFolder?: (item: FileSystemItem) => void
}

export const getItemMenuOptions = ({
  menuIconClass,
  file,
  inSharedFolder,
  itemOperations,
  editFile,
  deleteFile,
  downloadFile,
  moveFile,
  handleShare,
  recoverFile,
  previewFile,
  showFileInfo,
  reportFile,
  viewFolder
}: MenuOptionsParams) => {
```

packages/files-ui/src/Components/Modules/FileBrowsers/views/FilesList.tsx
```diff
@@ -851,10 +912,88 @@ const FilesList = ({ isShared = false }: Props) => {
...
+    classes.menuIcon,
+    getItemOperations,
+    itemFunctions,
+    bulkActions
+  ])
+
+  const anchorPosition = useMemo(() => contextMenuPosition ? {
+    top: contextMenuPosition.mouseY,
+    left: contextMenuPosition.mouseX
+  } : undefined, [contextMenuPosition])
```
Tbaut: Nit with ternary. I didn't put the object on several lines to keep it short
```suggestion
  const anchorPosition = useMemo(() => contextMenuPosition
    ? { top: contextMenuPosition.mouseY, left: contextMenuPosition.mouseX}
    : undefined
, [contextMenuPosition])
```

packages/files-ui/src/UI-components/AnchorMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const MENU_RIGHT_SPACE_TOLERANCE = 180
+const MENU_BOTTOM_SPACE_TOLERANCE = 250
+
+export default function MenuDropdown({
+  isOpen,
+  options,
+  anchorPosition,
+  onClose
+}: Props) {
```
Tbaut:
```suggestion
export default function MenuDropdown({ isOpen, options, anchorPosition, onClose }: Props) {
```

Tbaut (commented):
Love the refactoring and new functionality. I checked mostly Files code, left a couple nits. I did play with Storage as well and noticed that the menu, when it appears is displayed a very brief moment at the top left before being displayed at the right position. See my video

https://user-images.githubusercontent.com/33178835/169535088-6011feb7-68af-48ae-8f25-e844906c4ba1.mp4

tanmoyAtb:
> Love the refactoring and new functionality. I checked mostly Files code, left a couple nits. I did play with Storage as well and noticed that the menu, when it appears is displayed a very brief moment at the top left before being displayed at the right position. See my video
> 
>  flashing.mp4

I was able to see the flash in Firefox a few times. Made some changes to make sure that the dropdown only appears when we have an anchor position. Please check that it works well your side.

packages/files-ui/src/UI-components/AnchorMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const MENU_RIGHT_SPACE_TOLERANCE = 180
+const MENU_BOTTOM_SPACE_TOLERANCE = 250
+
+export default function MenuDropdown({ isOpen, options, anchorPosition, onClose }: Props) {
+  const position = useMemo(() => {
+    const windowHeight = window.innerHeight
+    const windowWidth = window.innerWidth
+    const position: Position = {}
+    if (!anchorPosition) return {}
```
Tbaut: An empty object is truthy, so better return undefined, and we can return earlier, and only if the menu is open.

```suggestion
    if (!anchorPosition || !isOpen) return

    const windowHeight = window.innerHeight
    const windowWidth = window.innerWidth
    const position: Position = {}
```

packages/files-ui/src/UI-components/AnchorMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+    // calculate left or right
+    if (windowWidth - anchorPosition.left > MENU_RIGHT_SPACE_TOLERANCE) {
+      position.left = anchorPosition.left
+    } else {
+      position.right = windowWidth - anchorPosition.left
+    }
+    return position
+  }, [anchorPosition])
+
+  const classes = useStyles(position)
```
Tbaut, line 82: Otherwise TS isn't happy because it can be undefined now.
```suggestion
  const classes = useStyles(position || {})
```

packages/files-ui/src/UI-components/AnchorMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+      position.bottom = windowHeight - anchorPosition.top
+    }
+    // calculate left or right
+    if (windowWidth - anchorPosition.left > MENU_RIGHT_SPACE_TOLERANCE) {
+      position.left = anchorPosition.left
+    } else {
+      position.right = windowWidth - anchorPosition.left
+    }
+    return position
+  }, [anchorPosition])
```
Tbaut, line 76: exaustive deps
```suggestion
  }, [anchorPosition, isOpen])
```

packages/files-ui/src/UI-components/AnchorMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  useEffect(() => {
+    if (isOpen) {
+      document.body.style.overflowY = "hidden"
+    } else {
+      document.body.style.overflowY = "scroll"
+    }
+  }, [isOpen])
+
+  if (!position) return null
```
Tbaut, line 91: Now that position can be undedined, this will work. It didn't really in the past because if `position` was an empty object, it didn't return `null`

packages/files-ui/src/UI-components/AnchorMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return (
+    <div
+      className={clsx(classes.anchorRoot, isOpen && !!position && "open")}
```
Tbaut: Same here, `!!position` was always true. You can get convinced by doing the following in the dev console `console.log(!!{}) //true`

Now we could actually remove it, it's redundant since the `isOpen` is included in the position. Up to you.

Tbaut (commented):
It didn't quite make the cut and was still flashing occasionally. I added some comments to completely solve the issue, with explanations. Let me know if that makes sense, and we should replicate this in storage as well.

tanmoyAtb:
Thanks :) :). This was perfect insight into the problem. I've made the updates to files and storage.

Tbaut:
I was still having issue, so I fought a bit and arrived at a state where I really really can't reproduce in this PR https://github.com/ChainSafe/ui-monorepo/pull/2152 :)

Tbaut (approved):
awesome awesome work, the files view is feeling great thanks to this!

PR #2139, opened 2022-05-19
[Files] Add download file test

### LINKED ISSUE #2132: [Files] Add download file from file browser test

- API response from `/download` endpoint returned 200 status code (doesn't have any response message)
- Toast visible and closable
- File was download (check if this is possible with Cypress)

juans-chainsafe:
closes #2132

Probably is a good idea to add this test case to **Storage** too, if this PR goes well, I will add that case after that.

### DIFF
--- a/packages/files-ui/cypress/tests/file-management-spec.ts
+++ b/packages/files-ui/cypress/tests/file-management-spec.ts
@@ -523,10 +523,10 @@ describe("File management", () => {
       cy.intercept("POST", "**/bucket/*/download")
         .as("downloadRequest")
         .then(() => {
-          homePage.downloadMenuOption().eq(0).click();
+          homePage.downloadMenuOption().eq(0).click()
 
           cy.wait("@downloadRequest").should((download) => {
-            expect(download.response).to.have.property("statusCode", 200);
+            expect(download.response).to.have.property("statusCode", 200)
           })
         })
      
@@ -536,8 +536,8 @@ describe("File management", () => {
       cy.get<string>("@fileContent").then((fileContent) => {
         cy.readFile(path.join(downloadsFolder, fileName))
           .should("exist")
-          .should("eq", fileContent);
-      });
+          .should("eq", fileContent)
+      })
 
     })
   })
--- a/packages/files-ui/cypress/tests/file-management-spec.ts
+++ b/packages/files-ui/cypress/tests/file-management-spec.ts
@@ -526,8 +526,8 @@ describe("File management", () => {
 
           cy.wait("@downloadRequest").should((download) => {
             expect(download.response).to.have.property("statusCode", 200);
-          });
-        });
+          })
+        })
      
       // ensure the file was download
       downloadCompleteToast.body().should("be.visible")
--- a/packages/files-ui/cypress/tests/file-management-spec.ts
+++ b/packages/files-ui/cypress/tests/file-management-spec.ts
@@ -511,8 +511,9 @@ describe("File management", () => {
 
       cy.web3Login({ clearCSFBucket: true })
   
-      // upload a file
+      // upload a file and store file content
       homePage.uploadFile(`../fixtures/uploadedFiles/${fileName}`)
+      cy.readFile(`cypress/fixtures/uploadedFiles/${fileName}`).as("fileContent")
       homePage.fileItemRow().should("have.length", 1)
     
       // download file from kebab menu 
@@ -529,10 +530,15 @@ describe("File management", () => {
           })
         })
      
-      // ensure the file was download
+      // ensure the file was downloaded
       downloadCompleteToast.body().should("be.visible")
       downloadCompleteToast.closeButton().click()
-      cy.readFile(path.join(downloadsFolder, fileName)).should("exist")
+      cy.get<string>("@fileContent").then((fileContent) => {
+        cy.readFile(path.join(downloadsFolder, fileName))
+          .should("exist")
+          .should("eq", fileContent);
+      });
+
     })
   })
 })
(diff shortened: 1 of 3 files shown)


FSM1 (approved):
I guess testing whether the downloaded file contents match what was uploaded is a bit out of scope? Might be an interesting addition to this test.

juans-chainsafe:
> I guess testing whether the downloaded file contents match what was uploaded is a bit out of scope? Might be an interesting addition to this test.

I didn't though about that, but seems a nice validation, apparently Cypress has a function readFile so I think I can add it in this PR, doesn't seem very complicated.

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -502,5 +504,35 @@ describe("File management", () => {
...
+        .as("downloadRequest")
+        .then(() => {
+          homePage.downloadMenuOption().eq(0).click();
+
+          cy.wait("@downloadRequest").should((download) => {
+            expect(download.response).to.have.property("statusCode", 200);
+          })
+        })
+     
+      // ensure the file was download
```
asnaith:
```suggestion
      // ensure the file was downloaded
```

asnaith (approved):
Nicely written test :) I can see why you used the const instead of a cypress alias. Makes sense for the path assertion. 

Good use of the intercept and wait for reliability too 🚀

juans-chainsafe:
File validation added! @FSM1

tanmoyAtb (approved):
Great work. Testing that the content matches is really reassuring !

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -502,5 +504,41 @@ describe("File management", () => {
...
+          cy.wait("@downloadRequest").should((download) => {
+            expect(download.response).to.have.property("statusCode", 200)
+          })
+        })
+     
+      // ensure the file was downloaded
+      downloadCompleteToast.body().should("be.visible")
+      downloadCompleteToast.closeButton().click()
+      cy.get<string>("@fileContent").then((fileContent) => {
+        cy.readFile(path.join(downloadsFolder, fileName))
```
Tbaut: Since `path` isn't a direct dependency of this project, I wouldn't use it here, just to make sure nothing breaks unexpectedly. It's also super simple in this case. 
```suggestion
        cy.readFile(`${downloadsFolder}/${fileName}`)
```

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -502,5 +504,41 @@ describe("File management", () => {
       fileInfoModal.closeButton().click()
       fileInfoModal.body().should("not.exist")
     })
+
+    it("can download a file from file browser", () => {
+      const fileName = "text-file.txt"
+      const downloadsFolder = Cypress.config("downloadsFolder")
```
Tbaut, line 509: I would consider storing the relative path as well so that it can be reused
```suggestion
      const downloadsFolder = Cypress.config("downloadsFolder")
      const fileRelativePath = `../fixtures/uploadedFiles/${fileName}`
```

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -502,5 +504,41 @@ describe("File management", () => {
...
+
+      cy.web3Login({ clearCSFBucket: true })
+  
+      // upload a file and store file content
+      homePage.uploadFile(`../fixtures/uploadedFiles/${fileName}`)
```
Tbaut:
```suggestion
      homePage.uploadFile(fileRelativePath)
```

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -502,5 +504,41 @@ describe("File management", () => {
...
+      cy.readFile(`cypress/fixtures/uploadedFiles/${fileName}`).as("fileContent")
```
Tbaut:
```suggestion
      cy.readFile(fileRelativePath).as("fileContent")
```

Tbaut (commented):
Great test Juan! I've just been messing a bit with the code, trying to not bring in any unneeded dependency.

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -3,6 +3,7 @@ import { homePage } from "../support/page-objects/homePage"
 import { navigationMenu } from "../support/page-objects/navigationMenu"
 import { folderName, folderPath } from "../fixtures/filesTestData"
 import "cypress-pipe"
+import path from "path"
```
Tbaut: This can be then removed.
```suggestion
```

packages/files-ui/cypress/tests/file-management-spec.ts
juans-chainsafe: I made all the changes, seems really great! but I have a problem with this one, I didn't put in a single variable the relative path because cypress readFile didn't work with "../fixtures/uploadedFiles" I needed to add cypress and remove the dots: "cypress/fixtures/uploadedFiles"

juans-chainsafe: I manage to resolve doing this:

```
 const fileRelativePath = `uploadedFiles/${fileName}`
 homePage.uploadFile(`../fixtures/${fileRelativePath}`)
 cy.fixture(fileRelativePath).as("fileContent")
```

Tbaut: great

Tbaut:
I didn't want to drag this branch too long so I pushed to it directly to:
- remove the `.only`
- remove some stray empty spaces (I'll get in touch to help you with some setup of your IDE to prevent that)
- simplify a bit these name stuff

PR #2148, opened 2022-05-23
Add notification when storage is running low

### LINKED ISSUE #2146: Notify users when they reach 75% of their storage quota
**Task**

In order to provide awareness of our upgrade feature 

**For all users that have reached 75% of their storage quota** 

- Present a notification in the notification manager informing the user that they have reached 75% of their storage quota. 

**For users who are on the early access list AND have reached 75% of their storage quota** 

- Present a notification in the notification manager informing the user that they have reached 75% of their storage quota and the notification should have an upgrade link which when clicked takes the user to the Billing tab within Settings

- Change the "Need more storage" button to "Upgrade" which when clicked also takes the user to the Billing tab within Settings

Tbaut:
closes #2146
 
I added the ability to close "turn off" a notification on click, until the next refresh at least :)

### DIFF
--- a/packages/files-ui/src/Contexts/BillingContext.tsx
+++ b/packages/files-ui/src/Contexts/BillingContext.tsx
@@ -68,16 +68,21 @@ const BillingProvider = ({ children }: BillingContextProps) => {
   const { filesApiClient, isLoggedIn, accountRestricted } = useFilesApi()
   const { redirect } = useHistory()
   const { addNotification, removeNotification } = useNotifications()
-  const { refreshBuckets } = useFiles()
+  const { refreshBuckets, storageSummary } = useFiles()
   const [currentSubscription, setCurrentSubscription] = useState<CurrentSubscription | undefined>()
   const [defaultCard, setDefaultCard] = useState<Card | undefined>(undefined)
   const [invoices, setInvoices] = useState<InvoiceResponse[] | undefined>()
   const isPendingInvoice = useMemo(() => currentSubscription?.status === "pending_update", [currentSubscription])
   const openInvoice = useMemo(() => invoices?.find((i) => i.status === "open"), [invoices])
   const [restrictedNotification, setRestrictedNotification] = useState<string | undefined>()
+  const [upgradeNotification, setUpgradeNotification] = useState<string | undefined>()
   const [unpaidInvoiceNotification, setUnpaidInvoiceNotification] = useState<string | undefined>()
   const [cardExpiringNotification, setCardExpiringNotification] = useState<string | undefined>()
   const [isBillingEnabled, setIsBillingEnabled] = useState(false)
+  const shouldProposeUpgrade = useMemo(() => storageSummary
+    ? storageSummary.used_storage > storageSummary.total_storage * 0.005
+    : false
+  , [storageSummary])
 
   const refreshInvoices = useCallback(() => {
     if (!currentSubscription) return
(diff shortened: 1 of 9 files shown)


packages/files-ui/src/Contexts/BillingContext.tsx
```diff
@@ -68,16 +68,21 @@ const BillingProvider = ({ children }: BillingContextProps) => {
...
   const [invoices, setInvoices] = useState<InvoiceResponse[] | undefined>()
   const isPendingInvoice = useMemo(() => currentSubscription?.status === "pending_update", [currentSubscription])
   const openInvoice = useMemo(() => invoices?.find((i) => i.status === "open"), [invoices])
   const [restrictedNotification, setRestrictedNotification] = useState<string | undefined>()
+  const [upgradeNotification, setUpgradeNotification] = useState<string | undefined>()
   const [unpaidInvoiceNotification, setUnpaidInvoiceNotification] = useState<string | undefined>()
   const [cardExpiringNotification, setCardExpiringNotification] = useState<string | undefined>()
   const [isBillingEnabled, setIsBillingEnabled] = useState(false)
+  const shouldProposeUpgrade = useMemo(() => storageSummary
+    ? storageSummary.used_storage > storageSummary.total_storage * 0.75
```
Tbaut, line 83: @asnaith or @juans-chainsafe if you want to test, just change this line to whatever percentage you want, to see the notification appear.

juans-chainsafe (approved):
Nicely done! and yep I tested locally because I needed to change the percentage, is working correctly! based on Andrew ticket.

PS: I will do the translations with Weblate to see how it works

packages/files-ui/src/Contexts/BillingContext.tsx
```diff
@@ -118,6 +123,20 @@ const BillingProvider = ({ children }: BillingContextProps) => {
     }
   }, [accountRestricted, addNotification, redirect, removeNotification, restrictedNotification])
 
+  useEffect(() => {
+    if (shouldProposeUpgrade && !upgradeNotification) {
```
tanmoyAtb, line 127: we should add `isBillingEnabled` as a condition here, otherwise notification shows up to go to billing page for all users.

tanmoyAtb (commented):
Looks good, leaving a comment to check `isBillingEnable` before showing notification.

asnaith:
Thanks for the tips on how to test! 

This is working well for most parts, just noticed the same thing as Tanmoy - users not whitelisted for billing can still click the notification but as they don't have access to billing it takes them to an empty settings page. 

I think we should just tell the user their storage is low and remove the "upgrade here" text / click through if they're not whitelisted. 

Screen Shot 2022-05-24 at 11 24 50 PM

packages/files-ui/src/Contexts/BillingContext.tsx
Tbaut: ooh well spotted, actually we should redirect them to our beta email collection

Tbaut:
I added a proper redirection to our form, and as a side note, I asked the design team to change the copy a bit to give more context.

packages/files-ui/src/Contexts/BillingContext.tsx
tanmoyAtb: Yupp, this looks more appropriate.

tanmoyAtb (approved):
Looks good to go !

asnaith:
> I added a proper redirection to our form, and as a side note, I asked the design team to change the copy a bit to give more context.

Nice change, looks good to go now :)

asnaith (approved):

PR #2149, opened 2022-05-23
Toasts update for uploads and downloads [storage] 

### LINKED ISSUE #2118: Port dismissible toast behaviour to Storage UI
We did some work in the Files UI to make toasts dismissible. It would be good to port this across to Storage UI too.

tanmoyAtb:
closes #2118 

Moved old code and added in toasts system to storage.


packages/storage-ui/src/Components/Modules/UploadProgressModals/UploadBox.tsx
```diff
@@ -1,114 +0,0 @@
...
-    progress,
-    errorMessage
-  } = uploadInProgress
-  const classes = useStyles()
-
-  return (
-    <>
-      <div
-        className={clsx(classes.appearBox, classes.boxContainer)}
-        data-cy="upload_status_toast_message" >
```
juans-chainsafe, line 72: @tanmoyAtb there are 2 tests failing because they can't find this id in the toast (after uploading a file in chainsafe or ipfs), in the new file this id changed or is not present?

packages/storage-ui/src/Contexts/StorageContext.tsx
```diff
@@ -340,17 +310,22 @@ const StorageProvider = ({ children }: StorageContextProps) => {
...
-      error: false,
-      noOfFiles: files.length,
+    const cancelSource = axios.CancelToken.source()
+    const cancelToken = cancelSource.token
+
+    const toastParams: ToastParams = {
+      title: plural(files.length, {
+        one: `Uploading ${files.length} file`,
+        other: `Uploading ${files.length} files`
+      }) as string,
```
Tbaut: plural returns a string, so that casting isn't needed.
```suggestion
      }),
```

Tbaut (commented):
Nice, that's a great cleanup! just left a comment (that is probably also in Files I imagine).

The tests are failing on the bucket management, which sounds like it could be a valid point. Let's check it out if it fails after a new dev merge

packages/storage-ui/src/Components/Modules/UploadProgressModals/UploadBox.tsx
asnaith: As discussed (but typing here for everyone else to see) the upload status toast in storage was more generic and it was using the same identifier for all uploading states. 

With the new style of toasts we can add an identifier for each state so all we should need to do here is: 

- remove the `uploadStatusToast` and add a `uploadCompleteToast` (define the same as the toast being used for Files tests) and then update the spec.

juans-chainsafe: refactor done! also I moved "bucketContentsPage.awaitBucketRefresh()" before the toast appears and after the click in upload file modal, because otherwise it fails because it doesn't find that request (I checked in localhost and the request is there but in the old way can't find it because it already happened) so, now looks like this:

```
      fileUploadModal.uploadButton().safeClick()
      fileUploadModal.body().should("not.exist")
      bucketContentsPage.awaitBucketRefresh()
      uploadCompleteToast.body().should("be.visible")
      uploadCompleteToast.closeButton().click()
```

tanmoyAtb: Thanks a lot @juans-chainsafe, really appreciate it.

juans-chainsafe (approved):

Tbaut (approved):
Awesome work

PR #2151, opened 2022-05-24
Add ui tests for change password

### LINKED ISSUE #2137: Add ui test coverage for changing password
Add new ui tests that cover the below scenarios:

- Changing the account password
- Ensuring errors are displayed when the password criteria is not met

(_Previously we were unable to test this critical functionality without breaking subsequent tests but now that we have the ability to use a new session we can do it without any conflicts_ 😃).

asnaith:
closes #2137


packages/files-ui/cypress/tests/settings-spec.ts
```diff
@@ -55,12 +58,72 @@ describe("Settings", () => {
...
+      settingsPage.securityTabButton().click()
+      settingsPage.changePasswordLink().click()
+
+      // ensure an error is displayed if password fields are blank
+      settingsPage.passwordInput().type("{selectall}{del}{esc}")
+      settingsPage.confirmPasswordInput().type("{selectall}{del}{esc}")
+      settingsPage.changePasswordButton().click()
+      settingsPage.changePasswordErrorLabel()
+        .should("be.visible")
+        .should("have.length.at.least", 2)
```
asnaith: I'm not able to add an identifier to the actual error message text so I added an identifier to the form and try to locate it with this: `[data-cy=form-change-password] span.error`

However, I am having to use ` .should("have.length.at.least", 2)` here instead of  `.should("have.length", 2)` because non error elements are matching my identifier and so I'm seeing 4 elements (the actual errors and the input title text). 

Example:
Screen Shot 2022-05-24 at 1 13 41 PM

Wondering if anyone has any suggestions on how to better approach this? I think I need to either fix up the non error elements so they don't appear as errors or add an identifer directly on the error text, at the moment I'm not sure how to do either.

packages/files-ui/cypress/tests/settings-spec.ts
```diff
@@ -55,12 +58,72 @@ describe("Settings", () => {
...
+        .should("have.length.at.least", 1)
+
+      // should see an error if confirm password does not match password
+      settingsPage.passwordInput().type(complexPassword)
+      settingsPage.confirmPasswordInput().type(simplePassword)
+      settingsPage.changePasswordButton().click()
+      settingsPage.changePasswordErrorLabel()
+        .should("be.visible")
+        .should("have.length.at.least", 1)
+    })
```
asnaith, line 126: I wanted to expand this test to include logging out and back in with the new password. However, **whilst under test**  if I do this I'm prompted to create (not enter) a password (and If I do that I'm logged in with a different address). 

It seems like the wallet address is randomized after signing out and I'm not too sure what's going on with the window injection etc? It works normally when I do the steps manually in a browser only an issue in test. Any tips? 

Example when the test is finished and I try to do the steps manually in the cypress window. I'm logged in with a different account: 

https://user-images.githubusercontent.com/3982665/170146078-1cf364fb-3932-4181-868c-cf20a008af94.mov

packages/files-ui/cypress/tests/settings-spec.ts
juans-chainsafe: I didn't see an option to add a identifier to the error label, hmm the only thing that I saw is to change the identifier to "[data-cy=form-change-password] span.default.error" (i saw that the "default" value is always added to the error message label

packages/files-ui/cypress/tests/settings-spec.ts
juans-chainsafe: so, if you try to do it in localhost:3000 outside Cypress window, works? if not I think is some configuration when running locally

asnaith: Yep, it works as normal on localhost, only an issue whilst under test. 

Pretty sure it's something to do with the way we fake the metamask signing but not quite sure where it gets the different address from. I'll re-visit with a fresh brain next week.

packages/files-ui/cypress/tests/settings-spec.ts
asnaith: @juans-chainsafe that was it! Thank you 🙏 I hadn't noticed the difference. As soon as I added that it was all good. I've updated the assertions and pushed an update

asnaith: @juans-chainsafe that was it! Thank you 🙏 I hadn't noticed the difference. As soon as I added that it was all good. I've updated the assertions and pushed an update

packages/files-ui/cypress/tests/settings-spec.ts
```diff
@@ -55,12 +58,71 @@ describe("Settings", () => {
...
+      // change password
+      settingsPage.passwordInput().type(complexPassword)
+      settingsPage.confirmPasswordInput().type(complexPassword)
+      settingsPage.changePasswordButton().click()
+
+      // ensure change password elements are not displayed after change
+      settingsPage.passwordInput().should("not.exist")
+      settingsPage.confirmPasswordInput().should("not.exist")
+      settingsPage.changePasswordButton().should("not.exist")
+    })
```
asnaith, line 86: I wanted to expand this check to include logging out and logging back in with the new password but I'm experiencing some weirdness with the fake metamask window / test wallet address. I'll look at that in a separate issue (#2156) so that this can be merged before I'm back online.

Tbaut (approved):
nice, as suggested in the followup issue, we may want to make sure the call to the api is made, with the right data, but we can add this later when closing the said issue.

juans-chainsafe (approved):
Awesome! agree with Tbaut with the assertion of the api call, but if we can do it in the new issue, then we can close this one

PR #2152, opened 2022-05-25
Fix context menu flash in Firefox

Tbaut:
I was still having issues despite my earlier suggestions, so I refactored a tiny bit and this time I can't reproduce any more.

This PR gets rid of the open/isOpen props and css. Unify the `anchorPosition` and `contextMenuPosition` which are the same.

How I could reproduce on Firefox before:
- refresh the page (with cache clean with ctrl+R)
- scroll
- right-click -> the first time it was flashing consistently.


tanmoyAtb (approved):
yup, I don't see any flashes or lag in the menu opening anymore.
I have a feeling this is because of the transition and slight delay in css picking up the position. 

This looks to the point, going to merge it in.

PR #2154, opened 2022-05-25
Fix context menu on modals

tanmoyAtb:
With the addition of context menu, Right click on any modal that is inside our file browser was opening the context menu. 

This PR fixes the issue. 
The simple fix was to move the modals out of the `article` component that responds to right click event.


asnaith:
Looks good on the modals now (these right-click changes are so awesome btw!). 

I noticed the same issue happens on the home page buttons too. Not sure if you want to apply the fix for that in this PR or should we create a new ticket? 

https://user-images.githubusercontent.com/3982665/170328318-83c4d00e-8420-4806-b292-12c0b30ecb9e.mov

juans-chainsafe:
Awesome! nice, I agree with Andrew, I think the right click should only trigger the context menu INSIDE the table, if we can add that will be great

tanmoyAtb:
> Looks good on the modals now (these right-click changes are so awesome btw!).
> 
> I noticed the same issue happens on the home page buttons too. Not sure if you want to apply the fix for that in this PR or should we create a new ticket?
> 
>  Screen.Recording.2022-05-25.at.10.35.20.AM.mov

definitely in this PR, looking into this.

tanmoyAtb:
All right. I've prevented right click menu on
The breadcrumbs
The action buttons in the header. 
I think this should cover where the context menu should be prevented.

juans-chainsafe:
> All right. I've prevented right click menu on The breadcrumbs The action buttons in the header. I think this should cover where the context menu should be prevented.

I can't see a commit for this, did you push the changes? also I checked the UI but is still the same as before

tanmoyAtb:
> > All right. I've prevented right click menu on The breadcrumbs The action buttons in the header. I think this should cover where the context menu should be prevented.
> 
> I can't see a commit for this, did you push the changes? also I checked the UI but is still the same as before

My bad. I pushed the changes now.

Tbaut (approved):
Some conflicts snuck in, but I could review and it's working well from my test :100:

juans-chainsafe (approved):
Working better now, thanks for the updates! I think once the PR conflict is resolve we can merge it

PR #2157, opened 2022-05-27
`no-loop-func` lint warning fix

tanmoyAtb:
We had a eslint warning about an `uploadedSize` variable.
Screenshot 2022-05-27 at 2 34 10 PM

This PR fixes the warning.
Alternatively I wanted to ignore the eslint rule. But this works as well 🤷


packages/files-ui/src/Contexts/FilesContext.tsx
```diff
@@ -384,6 +384,7 @@ const FilesProvider = ({ children }: FilesContextProps) => {
       for (const path of paths) {
         const filesToUpload = files.filter((f => getParentPathFromFilePath(f.path) === path))
         const batchSize = filesToUpload.reduce((sum, f) => sum += f.size, 0)
+        const uploadedSizeRef = uploadedSize
```
Tbaut, line 388:
```suggestion
        // prevent unsafe references warning on uploadedSize
        const uploadedSizeRef = uploadedSize
```

Tbaut (approved):
Nice one. I think the original warning doesn't make sense though, the warning acts as if the `for` loop wasn't waiting for the `encryptAndUpload` to be back, and the `uploadedSize` would be out of sync. The reality though is that the `for` is actually blocking because of the await, and the `uploadedSize` is always in sync.

Regardless, it's always better without any warning. I'd just add a comment for us to remember in the future.

tanmoyAtb:
> Nice one. I think the original warning doesn't make sense though, the warning acts as if the `for` loop wasn't waiting for the `encryptAndUpload` to be back, and the `uploadedSize` would be out of sync. The reality though is that the `for` is actually blocking because of the await, and the `uploadedSize` is always in sync.
> 
> Regardless, it's always better without any warning. I'd just add a comment for us to remember in the future.

Absolutely agreed, wanted to ignore the rule in eslint config but decided to let it stay. Might be of use some day.

juans-chainsafe (approved):
Checked locally and was fixed, nice!!

Issue #2112, opened 2022-04-27
[Files] Drag and drop a folder on file browser is uploading OS files


juans-chainsafe:
**Describe the bug**
When you drag and drop a folder containing files on the file browser it uploads the folders and files correctly but it also uploads OS files, eg. DS_Store. 

This only happens if you drag and drop on file browser, **it doesn't happen if you drag and drop on the modal**

**To Reproduce**
1. Login to Files
2. Drag and drop a folder with files to 'My Files'
3. Double click in the folder uploaded
4. See files imported

**Expected behavior**
Only regular files should be uploaded, system files should not be included

**Video**

https://user-images.githubusercontent.com/104365141/165584729-3ec36951-77c7-4040-9e70-4e91562a56be.mov

**Desktop (please complete the following information):**
Mac

**Additional context**
This only happens if you drag and drop on the file browser, **it doesn't happen if you drag and drop on the modal**

FSM1:
Investigate how the modal determines whether a file is hidden and attempt to reimplement this in the file browser DnD handler.

PR #2159, opened 2022-05-30
Filter out os hidden files

Tbaut:
closes #https://github.com/ChainSafe/ui-monorepo/issues/2112

The reason of the behavior is simple, we use 2 different packages. 
- `react-dropzone` in the modal, filter out the files (not the hidden ones, it's only based on the name)
https://github.com/react-dropzone/file-selector/blob/master/src/file-selector.ts#L4-L8
- `react-dnd` for the drag and drop doesn't filter out files.

This PR adds the filtering at the root level.


packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -147,6 +147,8 @@ const FileInput = ({
 
   const onDrop = useCallback(
     async (acceptedFiles: FileWithPath[], fileRejections: FileRejection[]) => {
+      console.log("acceptedFiles", acceptedFiles)
+      console.log("fileRejections", fileRejections)
```
Tbaut:
```suggestion
```

juans-chainsafe (approved):
Amazing, thanks for this update, working perfect with OS hidden files in Mac, both in Upload modal and doing drag and drop in the file browser

tanmoyAtb (approved):
Working great !

PR #2160, opened 2022-05-31
[Files] Copy to clipboard Wallet address and Files sharing key test

### LINKED ISSUE #2140: [Files] Add copy to clipboard fields in Profile test

- Copy to clipboard **Wallet address**
- Copy to clipboard **File sharing key**

juans-chainsafe:
closes #2140


packages/files-ui/cypress/tests/settings-spec.ts
```diff
@@ -123,6 +123,33 @@ describe("Settings", () => {
...
+      navigationMenu.settingsNavButton().click()
+      settingsPage.profileTabButton().click()
+      cy.url().should("include", "/settings/profile")
+
+      // ensure the correct wallet address is being copied to the clipboard
+      settingsPage.walletAddressLabel().click()
+      cy.window().its("navigator.clipboard").invoke("readText").then((text) => {
+        settingsPage.walletAddressLabel().should((element) => {
+          const walletAddress = element.text().split("...")
+          expect(text.startsWith(walletAddress[0])).to.be.true
```
juans-chainsafe, line 138: Here I had a problem, because the content that is copied to the clipboard is different from the text displayed in the UI (because is not complete and it adds "..." in the middle.

So, I manage to create two assertions, doing a split and then assert the start of the string, and second assert the end of the string

tanmoyAtb: Sounds reasonable to me !

tanmoyAtb (approved):

Tbaut (approved):
Looks good, nice workaround :)

PR #2162, opened 2022-06-01
[Files] Change to dark theme and go back to light theme test

### LINKED ISSUE #2141: [Files] Add change to dark theme test

In Settings -> Display change to dark theme and validate: 
- In Local Storage key `csf.themeKey` must be `dark`
- Dark theme checkbox must be selected
- Light theme checkbox must not be selected

juans-chainsafe:
closes #2141


packages/files-ui/src/locales/es/messages.po
```diff
@@ -5,8 +5,7 @@ msgstr ""
 "POT-Creation-Date: 2021-05-06 20:51+0200\n"
 "PO-Revision-Date: 2022-06-01 12:17+0000\n"
 "Last-Translator: Juan Manuel Spoleti <juan@chainsafe.io>\n"
-"Language-Team: Spanish <https://hosted.weblate.org/projects/chainsafe-files/"
```
juans-chainsafe: Weird, this is not from my commit, lingui automatically did this, but I didnt change these files

Tbaut (commented):
This test checks whether the local storage string is set or not, but whether or not this actually reflect on the UI, and that it does switch to dark/light mode is not tested. I didn't look in details how to do this, but I'm pretty sure we could check things like the background color of a div for instance.

packages/files-ui/src/locales/es/messages.po
Tbaut: don't worry, those are small things. Weblate does a formatting that lingui isn't happy with sometimes.

juans-chainsafe:
> This test checks whether the local storage string is set or not, but whether or not this actually reflect on the UI, and that it does switch to dark/light mode is not tested. I didn't look in details how to do this, but I'm pretty sure we could check things like the background color of a div for instance.

yes, you are right, and we do this validation as well:
`settingsPage.darkThemeLabel().get("div").should("have.class", "checked")`
to check that the div of the dark theme has the class "checked"

But we don't really check the UI (we though with Andrew of checking a background color but if that changes, the test will fail, but well, maybe those color hardly ever change) I will add that validation

Tbaut:
yup, if the test fails, then we'll fix it, it should be easy enough

juans-chainsafe:
> yup, if the test fails, then we'll fix it, it should be easy enough

changes uploaded! I validated the change of color in the display header

```
      //this is when it change to dark
      settingsPage.displayTabHeader().should("have.css", "color", "rgb(219, 219, 219)")
      
      //this is when it change to light
      settingsPage.displayTabHeader().should("have.css", "color", "rgb(15, 15, 15)")
```

packages/files-ui/cypress/support/page-objects/settingsPage.ts
```diff
@@ -24,6 +24,12 @@ export const settingsPage = {
   setUsernameButton: () => cy.get("[data-cy=button-set-username]"),
   usernamePresentInput: () => cy.get("[data-cy=input-profile-username-present]"),
 
+  // display tab
+  displayTabButton: () => cy.get("[data-testid=tab-display]"),
```
asnaith, line 28: Going by what @Tbaut mentioned here: https://github.com/ChainSafe/ui-monorepo/pull/2122#discussion_r868362304 

We should use data-cy instead of testid on elements like buttons and only use testid if data-cy doesn't work. Data-cy should work for this button.

packages/files-ui/cypress/support/page-objects/settingsPage.ts
```diff
@@ -24,6 +24,12 @@ export const settingsPage = {
...
+  displayTabHeader: () => cy.get("[data-cy=label-display-header]"),
+  darkThemeLabel: () => cy.get("[data-testid=radio-input-dark-theme]"),
+  lightThemeLabel: () => cy.get("[data-testid=radio-input-light-theme]"),
```
asnaith, line 31: 👍 These ones make sense though as I've never successfully got data-cy working with radio inputs

packages/files-ui/cypress/support/page-objects/settingsPage.ts
juans-chainsafe: Sadly didn't work with data-cy :(

asnaith: Ah ok, well that's fair enough then. It's valid usage 👍

asnaith (approved):
Awesome Juan, nice to have test coverage for this functionality :)

Tbaut (approved):
great one :black_flag:

PR #2164, opened 2022-06-02
Allow the upload of empty folder on drop

### LINKED ISSUE #2104: Empty directories are not uploaded.
**Describe the bug**
The app ignores empty directories, and will not upload them.

**To Reproduce**
1. Try to upload an empty directory.

Tbaut:
closes #2104

it's for Files for now, I'll do it for Storage as well if all goes well.


juans-chainsafe (changes requested):
So, the test related with upload a file are failing, because the file is not being uploaded.
Here is a video of what is happening:

https://user-images.githubusercontent.com/104365141/171713869-65cf9a4f-895d-4a0f-8755-334a2af450c2.mov

Basically, after you select a file nothing happens, the file is not added in the modal.
Drag and drop a file is working correctly, the problem is when you try to select a file from your machine.

Tbaut:
Oh, I completely forgot this case, I didn't expect to break it actually, I focused on the drag/drop. Thanks, I'll check it out.

packages/common-components/src/FileInput/FileInput.tsx
```diff
@@ -118,6 +119,7 @@ interface IFileInputProps extends DropzoneOptions {
     error?: string
   }
   onFileNumberChange: (filesNumber: number) => void
+  onEmptyFolderChange?: (emptyFolderPaths: string[]) => void
```
tanmoyAtb: Can we name this `onEmptyFolderPathsChange`

Tbaut:
Upload by clicking and selecting files should work as well now too.
I tested with a mix between drag (with empty folders) and manual selection, and it also working.
We def. need to test this thoroughly because there's a lot going on here.

packages/files-ui/src/Components/Modules/FileBrowsers/CSFFileBrowser.tsx
```diff
@@ -184,9 +183,15 @@ const CSFFileBrowser: React.FC<IFileBrowserModuleProps> = () => {
...
-    const flattenedFiles = await getFilesFromDataTransferItems(fileItems)
-    await uploadFiles(bucket, flattenedFiles, path)
-  }, [bucket, accountRestricted, storageSummary, addToast, uploadFiles])
+    const flattenedFiles = await getFilesAndEmptyDirFromDataTransferItems(fileItems)
+    await uploadFiles(bucket, flattenedFiles?.files || [], path)
+
+    //create empty dir
+    flattenedFiles?.emptyDirPaths?.forEach(async (folderPath) => {
+      await filesApiClient.addBucketDirectory(bucket.id, { path: getPathWithFile(currentPath, folderPath) })
+
```
tanmoyAtb:
```suggestion
```

packages/files-ui/src/Components/Modules/FileBrowsers/SharedFileBrowser.tsx
```diff
@@ -224,9 +232,15 @@ const SharedFileBrowser = () => {
...
-    const flattenedFiles = await getFilesFromDataTransferItems(fileItems)
-    await uploadFiles(bucket, flattenedFiles, path)
-  }, [bucket, accountRestricted, storageSummary, addToast, uploadFiles])
+    const flattenedFiles = await getFilesAndEmptyDirFromDataTransferItems(fileItems)
+    await uploadFiles(bucket, flattenedFiles?.files || [], path)
+
+    // create any empty dir
+    flattenedFiles?.emptyDirPaths?.forEach(async (folderPath) => {
+      await filesApiClient.addBucketDirectory(bucket.id, { path: getPathWithFile(currentPath, folderPath) })
+
```
tanmoyAtb, line 244:
```suggestion
```

packages/files-ui/src/Components/Modules/FileBrowsers/UploadFileModal.tsx
```diff
@@ -108,17 +112,24 @@ const UploadFileModule = ({ modalOpen, close }: IUploadFileModuleProps) => {
     try {
       close()
       await uploadFiles(bucket, values.files, currentPath)
+
+      //create empty folders
+      emptyFolderToCreate.forEach(async (folderPath) => {
+        await filesApiClient.addBucketDirectory(bucket.id, { path: getPathWithFile(currentPath, folderPath) })
+
```
tanmoyAtb, line 116:
```suggestion
```

tanmoyAtb (commented):
Amazing work, Its working well with a combination of cases. 

Two things we can improve. 

1. If we simply add an empty folder, the toast says `encrypting and uploading 0 files`, better not to show toast if we are not uploading files and simply adding folders

2. We need a refresh bucket call after the folder creation API calls, otherwise after the uploads call, there will be no refresh after the folder creation calls and we wouldn't see the folders created.

Adding a recording here.

https://user-images.githubusercontent.com/10713353/171831385-8cf32011-d4bf-4250-a5d2-3b76f6d22db7.mov

tanmoyAtb:
BTW if we want to add this into storage in this PR, we'd better merge in #2165 before, Folder uploads have been added in that PR.

Tbaut:
Yup, saw it yesterday, I'd add it after

juans-chainsafe:
Nice, now is working on files (upload file by selecting it), but... is not working on Storage, but that was working fine 4 hours ago, seems that has to be with the merge maybe? but well this is the error: 
"After you select a file nothing happens, the file is not added in the modal."

Also, I agree with Tanmoy, will be great to see the folder added at the moment (if you import an empty folder), so you don't need to refresh the page

Tbaut:
Thanks for the tests y'all. @tanmoyAtb great suggestions, it should be in now, a refresh happens after the folder creation now and we don't show the dialog if all there is, is an empty directory.

Tbaut:
@juans-chainsafe I guess it's related to what we merged indeed, it uses the same components. Also this PR will conflict most probably with the NFT upload one, it's touching the same components.

juans-chainsafe:
@Tbaut now is refreshing correctly the empty folder when is imported, nice!

As well, the error in Storage was fixed, but we have a new error in Storage, when we upload files from the file upload modal, is not automatically refreshing the table, so the files doesn't appear instantly (you need to refresh the page to see the files). Video below:

https://user-images.githubusercontent.com/104365141/171880698-3634d265-df9d-471a-b303-1281e3574f38.mov

Tbaut:
Yup, saw it, working on it

Tbaut:
let's see how much the NFT will break stuff. We should merge it before this one.

juans-chainsafe:
Seems to be working as expected now in Storage! let's wait for the NFT PR to be merged and I found something:

Something that I noticed testing this (is not from this PR) is that:

1. Create new folder and delete it
2. Create new folder with the same name and add some files
3. Delete that folder and go to Bin
In Bin, the content will go to the first folder that was deleted (because the second one that was deleted has the name {folderName} (1) ) so if I keep creating folders with the same name and delete them, the content inside the folder will go to the older folder.

I just wanted to comment on this behavior, maybe you've seen it before @Tbaut @asnaith @FSM1

Tbaut:
Ah good one Juan, I'll open an issue for that, it's a bug

asnaith:
>    In Bin, the content will go to the first folder that was deleted (because the second one that was deleted has the name {folderName} (1) ) so if I keep creating folders with the same name and delete them, the content inside the folder will go to the older folder.

@juans-chainsafe I wasn't aware of this bug, good discovery!

tanmoyAtb (approved):
Looks amazing on both storage and files !

Tbaut:
I wouldn't be against a last test on file upload since the conflicts weren't super small. It still looks good from my test with drag or manual upload.

juans-chainsafe (approved):
I reviewed again and all seems to be working perfect! I didn't see anything strange or failing, so nice work!

PR #2179, opened 2022-06-09
Bucket delete confirm modal 

### LINKED ISSUE #2170: [Storage] Add a modal confirmation for bucket deletion
Right now when deleting a bucket, it's done right away. This is obviously a little dangerous since it's a very destructive action, we should double check with the user that it's not a mis-click.

tanmoyAtb:
closes #2170


juans-chainsafe (approved):
Thanks for adding this! feel safer, I just have a suggestion, I'm not so sure about the text of the modal "You are about to delete bucket - {0}" It feels weird, but maybe it's me, I was thinking more of "You are about to delete the Bucket {0}." but let's see what the rest thinks

tanmoyAtb:
> https://storage-ui-stage-pr-2179.onrender.com

sounds better to me too. I'll update that.

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -261,7 +264,7 @@ const BucketsPage = () => {
             >
               <Trans>Size</Trans>
             </TableHeadCell>
-            <TableHeadCell>{/* Menu */}</TableHeadCell>
+            <TableHeadCell />
```
Tbaut: I find it useful to understand that this column title is left empty for the menu, otherwise it's not clear why we'd have such an empty one.
```suggestion
            <TableHeadCell>{/* Menu */}</TableHeadCell>
```

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
```diff
@@ -370,6 +374,26 @@ const BucketsPage = () => {
...
+        accept={async () => {
+          if (!bucketToRemove) return
+          setIsRemovingBucket(true)
+          removeBucket(bucketToRemove.id)
+            .catch(console.error)
+            .finally(() => {
+              setBucketToRemove(undefined)
+              setIsRemovingBucket(false)
+            })
+        }}
```
Tbaut: This is a big function, can we extract it to an external one with `useCallback` to have the JSX easier to read?

Tbaut (commented):
Nice one, I just realized we don't make an efficient use of some testId. I'll make a PR because it touches a lot of files

edit: it's here #2184

packages/storage-ui/src/Components/Pages/BucketsPage.tsx
tanmoyAtb: updates are in !

tanmoyAtb:
Updates are in :)

Tbaut (approved):
Thanks a lot!

PR #2182, opened 2022-06-10
[Files] Sort by name, date uploaded or size test

### LINKED ISSUE #2144: Add sort columns in file browser test
As mentioned, add test coverage in FILES and STORAGE for:

- Sort by Name
- Sort by Date uploaded (only Files)
- Sort by Size (check if is worth it)

juans-chainsafe:
closes #2144


asnaith (approved):
This is nice and easy to understand 👍

I can see this is very much dependent on the default state and will break if that changes but that is probably a good thing as we don't want that to change without us realizing it.

packages/files-ui/cypress/tests/file-management-spec.ts
```diff
@@ -559,5 +559,52 @@ describe("File management", () => {
...
+      homePage.fileItemName().eq(1).should("have.text", logoImage)
+      homePage.fileItemName().eq(2).should("have.text", textFile)
+
+      // ensure that sort by date uploaded in descending order (oldest file first)
+      homePage.fileDateUploadedColumnHeader().click()
+      homePage.fileItemName().eq(0).should("have.text", textFile)
+      homePage.fileItemName().eq(1).should("have.text", logoImage)
+      homePage.fileItemName().eq(2).should("have.text", fileZip)
+
+      // ensure that sort by size in ascending order (largest file first)
```
Tbaut, line 597: For this test, it would be interesting to have the `logoImage` have either the largest or the smallest size, because so far, in all tests the logoImage is in the between all the others. What do you think?

juans-chainsafe: sure, make sense, I can remove the fileZip and replace it with other png file that we have (chainsafe.png) in that case chainsafe.png will be the bigger file, next to logoImage and finally the txt file.

Or I can remove the fileZip and just use two files (text-file.txt and logo.png). What do you think?

Tbaut: I think the more files, the better, so 3 is good, I just thought we could be sure that the sorting is actually accurate if we end up in all possible cases, for example by name 1-2-3, reverse 3-2-1, by size 2-1-3, revers 3-1-2.

In these tests, we only have 1-2-3 and 3-2-1 if that makes sense

juans-chainsafe: oh sorry I misunderstood, got it, thanks

juans-chainsafe: done @Tbaut I change the order in which files are imported, in that way, when I sort by date uploaded I have a different order for each file

Tbaut: nice thanks

Tbaut (approved):
:rocket:

PR #2184, opened 2022-06-13
Remove duplicate testid for dialog buttons

Tbaut:
Small refactor of our Dialog `testId` handling, removing some unnecessary duplicate identifiers.


tanmoyAtb (approved):
looks good and tests are green !

PR #2189, opened 2022-06-14
Update test suite for Cypress 10

### LINKED ISSUE #2181: Migrate tests to use Cypress 10
As described, it's best for us to be on the latest supported version. 

This version bump is quite significant (compared to other upgrades) and requires some code changes on our side.

asnaith:
closes #2181

---

I've updated our test suite configuration to Cypress 10 and followed the latest updates/recommendations listed here: 
https://docs.cypress.io/guides/references/changelog#10-0-0

Everything is running well, but it's worth noting that more clicks are required when running the cypress UI locally for testing. 

- Every time you launch with `yarn test:files-ui` you now have to select the type of testing (e2e or component) and then select the browser, it never remembers your choices and there is no configuration option we can set yet (discussed here: https://github.com/cypress-io/cypress/issues/22152). 

- Also, there is no longer a "run all specs" button (discussed here: https://github.com/cypress-io/cypress/discussions/21628)


Tbaut (approved):
This is ready to review isn't it?
All in all I bet this will be a little more annoying for the QA teams, but as a dev there are only minor changes and the couple more clicks are ok I feel. I hope they'll improve with time but the addition of component tests sounds like a big deal, and they provided a migration path afaiu which is also pretty welcome I guess.

asnaith:
> This is ready to review isn't it? All in all I bet this will be a little more annoying for the QA teams, but as a dev there are only minor changes and the couple more clicks are ok I feel. I hope they'll improve with time but the addition of component tests sounds like a big deal, and they provided a migration path afaiu which is also pretty welcome I guess.

@Tbaut 👍 yep, was just making sure the tests were running in CI as expected. 

I agree the extra clicking is a minor annoyance for QA during test development but I think this will be improved by the cypress team and isn't something that will affect the dev team/ci.

Tbaut (approved):

juans-chainsafe (approved):
The additional clicks are a little annoying but is fine, they will probably fix it.
Test that are failing are because an error in the API, so not because this PR.
Looks good! thanks for updating it.

PR #2192, opened 2022-06-15
[Files] Change language test

### LINKED ISSUE #2171: [Files] Change language test
As described make coverage for:

- Change to Spanish
- Change to French
- Change to Norwegian
- Change back to English

juans-chainsafe:
closes #2171


juans-chainsafe:
Some subscription tests failed but not because of this PR, I already reported to API team

asnaith:
Thanks for adding this Juan 👍

I know we usually don't check element text values in tests but for a change language test I think it makes sense and what you have used seems very unlikely to change so I'm ok with it for this purpose. 

I pushed a few additional commits to update identifiers, simplify assertions, and increased the test reliability by waiting on the last api response (after language select + page refresh) before the assertions. The latter helped avoid "detached from DOM" errors.

asnaith (approved):

Tbaut (approved):
nice one!

Tbaut:
Merging since this has nothing to do with the failing tests that is being taken care of

juans-chainsafe:
> Thanks for adding this Juan 👍
> 
> I know we usually don't check element text values in tests but for a change language test I think it makes sense and what you have used seems very unlikely to change so I'm ok with it for this purpose.
> 
> I pushed a few additional commits to update identifiers, simplify assertions, and increased the test reliability by waiting on the last api response (after language select + page refresh) before the assertions. The latter helped avoid "detached from DOM" errors.

Thanks for the updates! now I know that I don't need the exact element to get the text, I can get it finding the parent element (I think in Selenium is not like that, because of that I didn't know that Cypress can do it).

And using the intercept for wait to finish is awesome too, Didn't think about that

PR #2196, opened 2022-06-20
storage navigation refactor

### LINKED ISSUE #2091: [Storage] Main Navigation Refactor
- Move API Keys into main navigation
- Handle the Subscription screen in main navigation (hide based on soft-launch flag)
- Leave current Settings route in the code, but hide Nav element (for now)

tanmoyAtb:
closes #2091

We now have Api Keys and Billing nav items in place.
Settings are kept hidden.


tanmoyAtb:
The settings tests are failing. Going to add the updates.

asnaith:
Working well and looks good for the most part but I'm a little bit undecided on whether this spacing makes sense now: 

Screen Shot 2022-06-20 at 3 15 03 PM

Tbaut (approved):
The code looks good. I have no strong opinion regarding the spacing, I'm undecided.

packages/storage-ui/src/Components/Layouts/AppNav.tsx
```diff
@@ -13,17 +13,20 @@ import {
   formatBytes,
   ChainsafeLogo,
   FolderSvg,
-  SettingSvg,
+  // SettingSvg,
```
juans-chainsafe: this can be removed I think

juans-chainsafe (commented):
Same as Tbaut, I'm not so sure about the spacing, maybe @serenaho can help? 

Regarding the changes, looks good! waiting for the fix of the tests to approve it

serenaho:
Included some options [here](https://www.figma.com/file/alES1NMSWR0ismBtjfYo9N/ChainSafe-Storage?node-id=503%3A1028) on Figma!

Options **A, B,** and **C** are pretty straightforward but options **D** and **E** have Docs and Billing separated as buttons because they don't quite fit under the category as a menu item like the others. 

Happy to discuss this during the planning meeting tmr :)

Tbaut:
This could be for another ticket but:
- I think the docs being an external link should indeed be separate
- I think we prob don't need headers or lines, there's already enough text
- Billing is a wrong name, I think we use "Subscription" in Files

My suggestion would be all together but Docs at the bottom of the nav bar

juans-chainsafe (approved):
Subscription test were failing in Files, but was a problem with the API that was fixed, I rerun the tests and now they all passed.

Everything fine now! I guess we leave the spacing improvement for another ticket

PR #2199, opened 2022-06-21
[Storage] Show folder's cid

### LINKED ISSUE #2197: [Storage] Show fodler CID
Folders have a cid that we should be able to show.

Tbaut:
closes #2197


juans-chainsafe (dismissed):
Working really good with ipfs folder! but is allowing to copy the CID of a Chainsafe folder and after you pin that folder, it remains forever with the status "pinning" so we should remove the CID for Chainsafe file system type, as the ticket said, is only for IPFS system type.

Screen Shot 2022-06-21 at 13 35 48

Tbaut:
Folders for Chainsafe File system are also a thing though, and although it may be less useful, I don't think we should dismiss them. The fact that you can't pin them may be another issue with the api.

edit: I had edited the original issue.

juans-chainsafe:
> Folders for Chainsafe File system are also a thing though, and although it may be less useful, I don't think we should dismiss them. The fact that you can't pin them may be another issue with the api.
> 
> edit: I had edited the original issue.

oh got it, I though that this can't work with chainsafe type, but yeah was confusion with the ticket description. I don't see anything strange in the api response, there is no errors there. I also tried with an empty folder

juans-chainsafe:
Some updates on this:

- Pin Chainsafe file 👍 
- Pin Chainsafe folder ❌ 
- Pin IPFS File 👍 
- Pin IPFS folder 👍 

Now (1 day later) I the status of the Chainsafe folder that I pinned is "failed" with this message in the api response: 
`Post "http://20.80.20.221:5001/api/v0/pin/add?arg=%2Fipfs%2Fbafkreigcqj77fo6fk7nxqfj25hr76uelpd4tchjp4lbungm3aeoovenr4u&recursive=true": context deadline exceeded`

FSM1 (approved):
Looks good my side

tanmoyAtb (approved):
Good to go !

Tbaut:
Thanks @juans-chainsafe for the thorough test. I just hid the cid for chainsafe file system. Can't think of a good way to explain why to our users but I can't think of a use-case where having that cid would make sense either.

juans-chainsafe:
> Thanks @juans-chainsafe for the thorough test. I just hid the cid for chainsafe file system. Can't think of a good way to explain why to our users but I can't think of a use-case where having that cid would make sense either.

Agree, this solve the issue, if we want to allow pin a chainsafe folder, we can create another ticket in that case. Thanks!

Issue #2198, opened 2022-06-21
Tests related to cross-origin navigation fail


Tbaut:
Some specific Storage tests have been failing recently with the following error, see [here](https://github.com/ChainSafe/ui-monorepo/runs/6982901126?check_suite_focus=true)

```
|   1) Main Navigation
 |        desktop
 |          can navigate to the docs page:
 |      CypressError: Timed out after waiting `60000ms` for your remote page to load on origin(s):
 | 
 | - `http://localhost:3000`
 | 
 | A cross-origin request for `[https://docs.storage.chainsafe.io/`](https://docs.storage.chainsafe.io/%60) was detected.
 | 
 | A command that triggers cross-origin navigation must be immediately followed by a `cy.origin()` command:
 | 
 | `cy.origin('https://chainsafe.io/', () => {`
 | `  <commands targeting https://docs.storage.chainsafe.io/ go here>`
 | `})`
 | 
 | If the cross-origin request was an intermediary state, you can try increasing the `pageLoadTimeout` value in `cypress.config.ts` to wait longer.
 | 
 | Browsers will not fire the `load` event until all stylesheets and scripts are done downloading.
 | 
 | When this `load` event occurs, Cypress will continue running commands.
```

PR #2200, opened 2022-06-21
[Storage] Skip navigate to the docs test

juans-chainsafe:
Skipping this test until we can see why sometimes fails.

Related to: #2198


asnaith (approved):

asnaith:
👍 Good idea to take this out of the regular suite until it's more reliable

Tbaut (approved):

PR #2207, opened 2022-06-28
sort Buckets table test

### LINKED ISSUE #2202: [Storage] Sort columns in Buckets view test
As described add coverage to sort columns in Buckets:

- By Name
- By File system
- By Size

juans-chainsafe:
closes #2202 

Very similar to the test of sorting Files table.


packages/storage-ui/cypress/support/page-objects/bucketsPage.ts
```diff
@@ -13,6 +14,7 @@ export const bucketsPage = {
   // bucket browser row elements
   bucketItemRow: () => cy.get("[data-cy=row-bucket-item]"),
   nameTableHeader: () => cy.get("[data-cy=table-header-name]"),
+  fileSystemTableHeader: () => cy.get("[data-cy=table-header-file-system]"),
```
tanmoyAtb: `bucketsTableHeader` sounds like a better name to me.
May want to change the data-cy on the HTML as well.

tanmoyAtb (approved):
Looks good, quick comment about the table header name and data-cy

packages/storage-ui/cypress/support/page-objects/bucketsPage.ts
juans-chainsafe: Done!

FSM1 (approved):

packages/storage-ui/cypress/support/page-objects/bucketsPage.ts
```diff
@@ -22,11 +24,19 @@ export const bucketsPage = {
   deleteBucketMenuOption: () => cy.get("[data-cy=menu-delete-bucket]"),
 
   // helpers and convenience functions
-  createBucket(bucketName: string) {
+  createBucket(bucketName: string, bucketType: FILE_SYSTEM_TYPES) {
```
Tbaut, line 27: I checked after merging, but this is somewhat of an outlier. It's working as you want, but we usually use type in a different way, without using enum, and a little less wordy, but effectively achieving the same, with type safety, this would be something like
```
type BucketType = "ipfs" | "chainsafe"

createBucket(bucketName: string, bucketType: BucketType) {
// ...
case "ipfs"
// ...
case "chainsafe"
```

juans-chainsafe: thanks for that Tbaut! I forget about type, but I remember that I use that to replace the enum 3 years ago haha, I will make this change in the "rename folder and file test" PR

PR #2211, opened 2022-07-01
Use personal_sign any connected wallet beside the known ones

### LINKED ISSUE #2209: Unable to sign message using EIP1193 Wallet
User has identified an issue that prevents authentication using Brave or other 3rd party wallet in the browser. 

Investigate why the Signature request fails to work when using the native wallet built in to Brave

Tbaut:
closes #2209

### DIFF
--- a/packages/files-ui/src/Components/Modules/LoginModule/InitialScreen.tsx
+++ b/packages/files-ui/src/Components/Modules/LoginModule/InitialScreen.tsx
@@ -254,6 +254,8 @@ const InitialScreen = ({ className }: IInitialScreen) => {
     try {
       await login(loginType)
     } catch (error: any) {
+      console.error(error)
+
       let errorMessage = t`There was an error authenticating`
 
       // Invalid signature, or contract wallet not deployed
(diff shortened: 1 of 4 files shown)


juans-chainsafe (approved):
Seems to be working correctly, I tried with BlockWallet in Brave and Chrome, as well Metamask in all browsers to make sure that keeps working.

I didn't find any other wallet that I can connect via "Connected" I tried with polkadot{.js} and Phantom in Firefox, but they are not detected by the app

Tbaut:
We only support Ethereum based wallets. So pjs and phantom aren't expected to work indeed.

FSM1:
Seems to be working well here, with the exception of the actual message being signed getting reduced to a hash: 
image

Not too sure if there is anything we can really do about that though.

FSM1 (approved):

Tbaut:
The hash is something that wallet probably chose to show this way unfortunately, since what we sign is the exact same api token than with `eth_sign`.

achnir97:
Iam not being able to connect the chainsafe files to Metamask wallet while running on localnet. If any body had the same issue and had solved the problem, I would highly appreciate for the solution.

Tbaut:
Thanks for your comment. Since this PR is not related to metamask and has been merged would you mind opening an issue and describe the steps that lead to the issue?

achnir97:
Sorry. Have deleted my comment.

PR #2212, opened 2022-07-04
Make sure newly created bucket names are differents

Tbaut:
By having the name in a shared variable, they ended up not being different/unique, and conflicts have appeared between tests.

### DIFF
--- a/packages/storage-ui/cypress/tests/bucket-management.cy.ts
+++ b/packages/storage-ui/cypress/tests/bucket-management.cy.ts
@@ -63,7 +63,7 @@ describe("Bucket management", () => {
     })
 
     it("can create, upload file and delete an ipfs bucket", () => {
-      const ipfsBucketName = `ipfs bucket ${Date.now()}
+      const ipfsBucketName = `ipfs bucket ${Date.now()}`
 
       cy.web3Login({ clearPins: true, deleteFpsBuckets: true })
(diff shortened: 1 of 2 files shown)


packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -62,6 +63,7 @@ describe("Bucket management", () => {
     })
 
     it("can create, upload file and delete an ipfs bucket", () => {
+      const ipfsBucketName = `ipfs bucket ${Date.now()}`
```
Tbaut, line 66:
```suggestion
      const ipfsBucketName = `ipfs bucket ${Date.now()}

```

packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -62,6 +63,8 @@ describe("Bucket management", () => {
     })
 
     it("can create, upload file and delete an ipfs bucket", () => {
+      const ipfsBucketName = `ipfs bucket ${Date.now()}
```
Tbaut:
```suggestion
      const ipfsBucketName = `ipfs bucket ${Date.now()}`
```

FSM1 (approved):
Thanks for sorting this out 🥇

juans-chainsafe (approved):
Thanks for this fix, nice catch!

PR #2213, opened 2022-07-05
Feat/nft grid view 2206

### LINKED ISSUE #2206: [Storage] Display NFTs that have been uploaded
@FSM1 commented on [Wed Jun 22 2022](https://github.com/ChainSafe/files-design/issues/132)

When a user stores NFT data using the storage API, the API creates a folder that contains the NFT JSON as well as any binary files referenced by the JSON.

Items that have been uploaded by a user in the `NFT bucket` should be listed as in the screenshot below

image

To Be Confirmed: 
- What info is available when an `ls` request is made for the NFT bucket? What information can we extract from the folder names? What info would we have to get from the NFT json file? Based on the outcome of this investigation, we might need to update the designs slightly (eg. remove date if this is unavailable at the folder level).

- Context menu on the NFT item can be left out for now, until we figure out what context menu options are available (Delete?, Details)

tanmoyAtb:
closes #2206

Feels like a straight forward solution so far. Reading on the `directory` CID gives the JSON with necessary data to display the NFT.


packages/storage-ui/src/Components/Modules/NFTsList/NFTItem.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const classes = useStyles()
+  const [isCopied, setIsCopied] = useState(false)
+  const [NFTData, setNFTData] = useState<NFTData>()
+
+  useEffect(() => {
+    axios.get(`${trimChar(IPFS_GATEWAY, "/")}/${CID}`)
+      .then(({ data }) => { setNFTData({
+        CID,
+        name: data.name,
+        image: data.image?.replace("ipfs://", `${trimChar(IPFS_GATEWAY, "/")}/`)
```
tanmoyAtb, line 83: Not very confident about this implementation. The JSON returns image URL in this format
`ipfs://${CID}`,
I've replaced the `ipfs://` with our gateway URL here.

packages/storage-ui/src/Contexts/StorageContext.tsx
```diff
@@ -221,7 +223,10 @@ const StorageProvider = ({ children }: StorageContextProps) => {
 
   const refreshBuckets = useCallback(() => {
     storageApiClient.listBuckets()
-      .then((buckets) => setStorageBuckets(buckets.filter(b => b.type === "fps")))
+      .then((buckets) => {
+        setNFTBucket(buckets.find(b => b.type === "nft" as any))
```
tanmoyAtb: Will remove the `"nft" as any` once we get it added to the spec

packages/storage-ui/src/Components/Modules/NFTsList/NFTItem.tsx
FSM1: thats exactly the way I would have gone about it

packages/storage-ui/src/Contexts/StorageContext.tsx
FSM1: Update to the API Spec is here: https://github.com/ChainSafe/files-api-spec/pull/164/commits/06b96d57ccfabd3cf0da1a2ac8cb2289b2a3da4e

packages/storage-ui/src/Components/Modules/NFTsList/NFTItem.tsx
```diff
@@ -0,0 +1,150 @@
...
+  Typography,
+  Button
+} from "@chainsafe/common-components"
+import { CSSTheme } from "../../../Themes/types"
+import { Trans } from "@lingui/macro"
+import clsx from "clsx"
+import axios from "axios"
+import { trimChar } from "../../../Utils/pathUtils"
+
+const IPFS_GATEWAY = process.env.REACT_APP_IPFS_GATEWAY || "https://ipfs.io/ipfs/"
```
Tbaut: I believe this is already defined in the `CidRow.ts`. We should ideally just have it once, imported from there.

packages/storage-ui/src/Components/Modules/NFTsList/NFTItem.tsx
```diff
@@ -0,0 +1,150 @@
...
+interface NFTData {
+  name: string
+  CID: string
+  image?: string
+}
+
+const NFTItem = ({ CID }: {CID: string}) => {
+  const classes = useStyles()
+  const [isCopied, setIsCopied] = useState(false)
+  const [NFTData, setNFTData] = useState<NFTData>()
```
Tbaut: I'm very surprised that TS doesn't complain here, did I miss something? I guess it should be 
```suggestion
  const [NFTData, setNFTData] = useState<NFTData | undefined>()
```

packages/storage-ui/src/Components/Pages/NFTsPage.tsx
```diff
@@ -0,0 +1,77 @@
...
+    if (dismissedFlag === undefined || dismissedFlag === null) {
+      localStorageSet(DISMISSED_SURVEY_KEY, "false")
+    }
+  }, [localStorageGet, localStorageSet])
+
+  const refreshContents = useCallback((showLoading?: boolean) => {
+    if (!NFTBucket) return
+    showLoading && setLoadingCurrentPath(true)
+    storageApiClient.getBucketObjectChildrenList(NFTBucket.id, { path: "/" })
+      .then(async (newContents) => {
```
Tbaut: I don't see any `await`
```suggestion
      .then((newContents) => {
```

Tbaut (approved):
Had a gap and took a look at the code, just found some nits

juans-chainsafe:
Test failed before because we had an issue with the download endpoint, was already fixed by the API team.
Looks good so far! but @tanmoyAtb or @FSM1 is there any way that I can mock nfts to the gallery? to see how it looks like, because if is empty, doesn't say much.

Also, I guess we already have a ticket to add this, but it will be nice to have something like "No NFTs to show" like we have when a folder is empty:
Screen Shot 2022-07-20 at 15 22 33

FSM1:
> Test failed before because we had an issue with the download endpoint, was already fixed by the API team. Looks good so far! but @tanmoyAtb or @FSM1 is there any way that I can mock nfts to the gallery? to see how it looks like, because if is empty, doesn't say much.

No easy way to mock this, but it should be trivial to just create some on your account.

> Also, I guess we already have a ticket to add this, but it will be nice to have something like "No NFTs to show" like we have when a folder is empty

This should be added in the current pr

juans-chainsafe:
> > Test failed before because we had an issue with the download endpoint, was already fixed by the API team. Looks good so far! but @tanmoyAtb or @FSM1 is there any way that I can mock nfts to the gallery? to see how it looks like, because if is empty, doesn't say much.
> 
> No easy way to mock this, but it should be trivial to just create some on your account.
> 
> > Also, I guess we already have a ticket to add this, but it will be nice to have something like "No NFTs to show" like we have when a folder is empty
> 
> This should be added in the current pr

hmm I tried to add a NFT, but the upload button doesn't do anything, maybe I'm doing  something wrong?

Screen Shot 2022-07-20 at 17 45 45

FSM1:
> hmm I tried to add a NFT, but the upload button doesn't do anything, maybe I'm doing something wrong?

Thanks for catching this issue @juans-chainsafe. This was a bug introduced by the latest version of the API client. I have since resolved this and the create form should now be working. There is no redirect after the creation (the creation screen is still a WIP), so you should manually go back to the NFTs page to view the newly created item.

juans-chainsafe:
> > hmm I tried to add a NFT, but the upload button doesn't do anything, maybe I'm doing something wrong?
> 
> Thanks for catching this issue @juans-chainsafe. This was a bug introduced by the latest version of the API client. I have since resolved this and the create form should now be working. There is no redirect after the creation (the creation screen is still a WIP), so you should manually go back to the NFTs page to view the newly created item.

I tried again but still can't do it, maybe it has to be with the compilation error in storage?
Screen Shot 2022-07-21 at 09 35 00
Also, the error is displayed in the storage tests: https://github.com/ChainSafe/ui-monorepo/runs/7439125401?check_suite_focus=true

packages/storage-ui/src/Components/Modules/NFTsList/NFTItem.tsx
```diff
@@ -0,0 +1,149 @@
...
+    axios.get(`${trimChar(IPFS_GATEWAY, "/")}/${CID}`)
+      .then(({ data }) => { setNFTData({
+        CID,
+        name: data.name,
+        image: data.image?.replace("ipfs://", `${trimChar(IPFS_GATEWAY, "/")}/`)
+      }) })
+      .catch(console.error)
+  }, [CID])
+
+  if (!NFTData) return null
```
juans-chainsafe, line 88: @tanmoyAtb I think we can add here the "No NFTs to show" (not necessary needs to be this message, is the one that occurs to me) section like we have in FilesList.tsx

tanmoyAtb: Yeap, adding this in !

juans-chainsafe:
Issue was fixed by Tanmoy, thanks! and the "No NFTs to show! is great as well!

But I have a different error now, after uploading a NFT (response is 200), when I go back to NFTs view the endpoint https://ipfs.chainsafe.io/ipfs/{cid} is returning 504 error, I upload 2 NFTs and the UI tries to bring both, but the request fails
Screen Shot 2022-07-22 at 09 20 43

EDIT: After 5-10 minutes the request worked and I can see the nfts now! @tanmoyAtb @FSM1 I guess this is expected for now? but seems that needs some time to upload and then be able to see them without that 504 error

FSM1:
> Issue was fixed by Tanmoy, thanks! and the "No NFTs to show! is great as well!
> 
> But I have a different error now, after uploading a NFT (response is 200), when I go back to NFTs view the endpoint [https://ipfs.chainsafe.io/ipfs/{cid}](https://ipfs.chainsafe.io/ipfs/%7Bcid%7D) is returning 504 error, I upload 2 NFTs and the UI tries to bring both, but the request fails Screen Shot 2022-07-22 at 09 20 43
> 
> EDIT: After 5-10 minutes the request worked and I can see the nfts now! @tanmoyAtb @FSM1 I guess this is expected for now? but seems that needs some time to upload and then be able to see them without that 504 error

This all makes sense, and is working correctly. Mainly this is caused by us uploading to the staging IPFS node, which is not exposed to the outside world as a gateway. All retrieval happens from the production IPFS node, which obviously has to retrieve the data from the staging IPFS node.

Once this code is in PROD this will all be handled by the same IPFS node, and as such will be much smoother experience.

juans-chainsafe:
Awesome! then I have 2 finals questions:
I think we can add a loading state while the request is pending like we have when you enter a bucket and show the message "One sec, getting NFTs ready..." @tanmoyAtb 

I tried to pin the CID of the NFT and that works! but the thing that I see is the json, for example: https://ipfs.chainsafe.io/ipfs/QmafgT8UrDXNKiQsYsaVPoy8UNdVxJpHkC3917wRyne9R4 and not the image, but I guess that is expected @FSM1
Also, when I try to unpin the CID of the NFT, the status is stuck in 'queued' (at least I did it 20 minutes ago and is still in queued)

FSM1:
> Awesome! then I have 2 finals questions: I think we can add a loading state while the request is pending like we have when you enter a bucket and show the message "One sec, getting files ready..." @tanmoyAtb

This would be a nice addition.

> I tried to pin the CID of the NFT and that works! but the thing that I see is the json, for example: https://ipfs.chainsafe.io/ipfs/QmafgT8UrDXNKiQsYsaVPoy8UNdVxJpHkC3917wRyne9R4 and not the image, but I guess that is expected 

That is entirely expected.

> @FSM1 Also, when I try to unpin the CID of the NFT, the status is stuck in 'queued' (at least I did it 20 minutes ago and is still in queued)

This is out of scope of the issue as currently we are not allowing delete/unpin operations on the NFT bucket. If the content is pinned separately it should be possible to unpin it, but once again, I do not foresee this as a very important use case yet.

tanmoyAtb:
> Awesome! then I have 2 finals questions: I think we can add a loading state while the request is pending like we have when you enter a bucket and show the message "One sec, getting NFTs ready..." @tanmoyAtb
> 
This is in 👍 
> I tried to pin the CID of the NFT and that works! but the thing that I see is the json, for example: https://ipfs.chainsafe.io/ipfs/QmafgT8UrDXNKiQsYsaVPoy8UNdVxJpHkC3917wRyne9R4 and not the image, but I guess that is expected @FSM1 Also, when I try to unpin the CID of the NFT, the status is stuck in 'queued' (at least I did it 20 minutes ago and is still in queued)

juans-chainsafe (approved):
Awesome! thanks for the effort, this is a great start with NFT. Of course I will review the design and "Upload NFT" in other tickets

PR #2220, opened 2022-07-11
[Storage] Download file test

### LINKED ISSUE #2205: [Storage] Add download file test
As described add coverage to be able to download a file in Storage

juans-chainsafe:
closes #2205


FSM1 (approved):
lgtm 🥇

packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -143,7 +144,6 @@ describe("Bucket management", () => {
       bucketsPage.bucketItemName().eq(0).should("have.text", chainSafeBucketName)
       bucketsPage.bucketItemName().eq(1).should("have.text", ipfsBucketName)
     })
-
```
asnaith: I think this was accidental

packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -289,5 +289,89 @@ describe("Bucket management", () => {
...
+      const fileName = "text-file.txt"
+      const downloadsFolder = Cypress.config("downloadsFolder")
+      const fileFixturePath = `uploadedFiles/${fileName}`
+
+      cy.web3Login({ deleteFpsBuckets: true })
+
+      // upload a file and store file content
+      navigationMenu.bucketsNavButton().click()
+
+      bucketsPage.createBucket(chainSafeBucketName, "chainsafe")
```
asnaith: Something for the future / separate ticket perhaps but when we need to create buckets as requisites for tests (but we're not testing bucket creation itself) it would probably be better to expand the `apiTestHelper` to add this ability. This way we can create the bucket programmatically and have it there at the beginning of the test which would be more efficient than setting it up in the UI. 

Our FE developers would be best able to advise on the feasibility of doing this but I think it should be possible.

packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -289,5 +289,89 @@ describe("Bucket management", () => {
...
+      // intercept POST to ensure the request was successful
+      cy.intercept("POST", "**/bucket/*/download")
+        .as("downloadRequest")
+        .then(() => {
+          bucketContentsPage.downloadMenuOption().click()
+
+          cy.wait("@downloadRequest").should((download) => {
+            expect(download.response).to.have.property("statusCode", 200)
+          })
+        })
```
asnaith, line 351: Nicely done! 🤩

packages/storage-ui/cypress/tests/bucket-management.cy.ts
juans-chainsafe: I think I can try to do this now, sounds fun

FSM1: Its definitely doable. Shout if you have any issues @juans-chainsafe

asnaith (dismissed):

packages/storage-ui/cypress/tests/bucket-management.cy.ts
juans-chainsafe: thanks @FSM1 ! I just pushed the changes, I did it based on how we create folder in Files using the apiTestHelper @asnaith

packages/storage-ui/cypress/support/utils/apiTestHelper.ts
```diff
@@ -14,6 +16,23 @@ const getApiClient = () => {
...
+        cy.log("creating bucket", name)
+        const tokens = await apiClient.getRefreshToken({ refresh: win.localStorage.getItem(REFRESH_TOKEN_KEY) || "" })
+        apiClient.setToken(tokens.access_token.token)
+        await apiClient.createBucket({ name: name, file_system_type: fileSystemType, type: "fps", encryption_key: "" })
+        cy.log("done with creating bucket")
+      })
+      navigationMenu.cidsNavButton().click()
+      navigationMenu.bucketsNavButton().click()
+      bucketsPage.awaitBucketRefresh()
+      bucketsPage.bucketItemRow().should("have.length", 1)
```
asnaith: The navigation clicks that move away and then back to the buckets page are just to ensure that the UI is updated after creating the required buckets via the api, so I don't think we should be doing this before `resolve()`? 

We should also catch any errors that happen during bucket creation so that we know where the problem is if/when it occurs. Otherwise, the problem would be obscured and it will result in a failure of the next test step (which it shouldn't have got to) and this could cause confusion when debugging failing tests. 

If we do this, we can then we can omit `bucketsPage.bucketItemRow().should("have.length", 1)` which I think makes sense, we shouldn't be using assertions outside of the test layer and this would also fail if more than one bucket was created anyway.

packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -186,11 +183,10 @@ describe("Bucket management", () => {
       const folderName = `folder ${Date.now()}`
       const newFolderName = `new folder name ${Date.now()}`
 
-      cy.web3Login({ deleteFpsBuckets: true })
+      cy.web3Login({ deleteFpsBuckets: true, createFpsBuckets: [{ name: chainsafeBucketName, type: "chainsafe" }] })
       navigationMenu.bucketsNavButton().click()
 
       // create a new bucket and go inside the bucket
```
asnaith: This comment can be removed now as the bucket is already created by this point

packages/storage-ui/cypress/support/utils/apiTestHelper.ts
juans-chainsafe: You are right that should be after the resolve, buuut we need that because we are using the api to create the bucket, so the UI is not updated with that bucket without refreshing the page (I change the go to cids and then go to buckets, with a cy.refresh() I think has more sence).

Added as well the try and catch logic and removed the validation

asnaith: Sounds good. We have seen "detached from DOM" issues when using cy.reload() in the past but if it's working reliably here then that's fine with me :)

asnaith (commented):
Thanks for this @juans-chainsafe!  

I think it would be a good idea to get a second review from @FSM1 on this before merging because these apiTestHelper changes arrived after the original approval.

asnaith (approved):

packages/storage-ui/cypress/support/utils/apiTestHelper.ts
juans-chainsafe: I will take a look at that, and if happens, I will go to the original implementation of going to cids and then bucket page

FSM1:
> Thanks for this @juans-chainsafe!
> 
> I think it would be a good idea to get a second review from @FSM1 on this before merging because these apiTestHelper changes arrived after the original approval.

Checked through the bucket creation code and it looks good to me. Happy with this being merged @juans-chainsafe @asnaith

PR #2224, opened 2022-07-14
[Storage] Validate errors on create bucket modal test

### LINKED ISSUE #2221: [Storage] Add validate errors when trying to create a bucket test
As described add coverage when creating a bucket:

- User cannot create a bucket with no name
- User cannot create a bucket with disallowed characters (eg space only)

juans-chainsafe:
closes #2221

### DIFF
--- a/packages/storage-ui/cypress/tests/bucket-management.cy.ts
+++ b/packages/storage-ui/cypress/tests/bucket-management.cy.ts
@@ -21,15 +21,31 @@ describe("Bucket management", () => {
       createBucketModal.cancelButton().click()
       createBucketModal.body().should("not.exist")
 
-      // create a bucket and see it in the bucket table
+      // go to create bucket modal
       bucketsPage.createBucketButton().click()
       createBucketModal.body().should("be.visible")
+      // ensure can't create an empty bucket
+      createBucketModal.submitButton().click()
+      createBucketModal.bucketNameInput().should("have.class", "error")
+      // ensure can't create a bucket with only spaces
+      createBucketModal.bucketNameInput().type("  ")
+      createBucketModal.submitButton().click()
+      createBucketModal.bucketNameInput().should("have.class", "error")
+      // create a bucket and see it in the bucket table
       createBucketModal.bucketNameInput().type(chainSafeBucketName)
+      createBucketModal.bucketNameInput().should("not.have.class", "error")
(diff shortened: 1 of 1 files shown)


packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -21,16 +21,32 @@ describe("Bucket management", () => {
       createBucketModal.cancelButton().click()
       createBucketModal.body().should("not.exist")
 
-      // create a bucket and see it in the bucket table
+      // go to create bucket modal
       bucketsPage.createBucketButton().click()
       createBucketModal.body().should("be.visible")
+      // ensure can't create an empty bucket
```
asnaith, line 28: Minor thing but I find that leaving a whitespace before each comment / individual check in the flow of the test helps readability, it's not done here but is elsewhere in the test.

packages/storage-ui/cypress/tests/bucket-management.cy.ts
```diff
@@ -21,16 +21,32 @@ describe("Bucket management", () => {
...
+      createBucketModal.submitButton().click()
+      createBucketModal.bucketNameInput().should("have.class", "error")
```
asnaith, line 30: Interesting, I've never thought about using the class name before like this. I would usually avoid doing that (as I'd be concerned about it being too generic and potentially picking up a false match from elsewhere on the page)...but you've scoped it to within the bucketNameInput element so I guess that's not a valid concern, nice! 🤩

Tbaut (approved):
:+1:

packages/storage-ui/cypress/tests/bucket-management.cy.ts
juans-chainsafe: yep! it helps, done!

packages/storage-ui/cypress/tests/bucket-management.cy.ts
juans-chainsafe: yes! the other way around is checking that the modal is still present and validating the text... but we don't want to add validations that have hardcoded text, so I ended up doing that inside the input, so we shouldn't have any problem. Also, I added a validation when the input is correct that the class 'error' should not be present

asnaith (approved):
Looks good! Thanks for this.

PR #2232, opened 2022-07-20
Remove Posthog

FSM1:
Removes Posthog from Files & Storage


Tbaut (approved):
Looks good, I just found a `cookieBanner` variable in the theme that's not used anymore. Approving still so as not to prevent you from merging quickly

tanmoyAtb (approved):

PR #2234, opened 2022-07-25
[Storage] pin same cid twice test

### LINKED ISSUE #2230: [Storage] Add pin the same cid twice test
We already have the test "can see a warning when attempting to pin the same CID twice" automated, the idea here is to complete that flow pinning the cid and see it in the CID table twice.

juans-chainsafe:
closes #2230 

We already had the test that check the warning message if we try to pin the same CID, I just completed the flow, adding that cid again, and validate that is present in the table twice


packages/storage-ui/cypress/tests/cid-management.cy.ts
```diff
@@ -42,7 +42,7 @@ describe("CID management", () => {
       cidsPage.cidItemRow().should("contain.text", "queued")
     })
 
-    it("can see a warning when attempting to pin the same CID twice", () => {
+    it("can pin the same CID twice and see the warning message", () => {
```
asnaith:
```suggestion
    it("can pin the same CID twice following initial warning", () => {
```

asnaith (approved):
Left a minor suggestion for readability but looks good :)

packages/storage-ui/cypress/tests/cid-management.cy.ts
juans-chainsafe: yeaah! I had my doubts about this line, because sounded like the warning appears after pin the same cid, thanks!

FSM1 (approved):

PR #2239, opened 2022-08-04
Disable IPFS MFS filesystem creation

FSM1:
Submission checklist: 

#### Layout

#### Theme

### DIFF
--- a/packages/storage-ui/cypress/tests/bucket-management.cy.ts
+++ b/packages/storage-ui/cypress/tests/bucket-management.cy.ts
@@ -180,41 +180,41 @@ describe("Bucket management", () => {
       bucketsPage.bucketItemName().eq(1).should("have.text", ipfsBucketName)
     })
 
-    it("can rename a folder inside the ipfs bucket", () => {
-      const ipfsBucketName = `ipfs bucket ${Date.now()}`
-      const folderName = `folder ${Date.now()}`
-      const newFolderName = `new folder name ${Date.now()}`
+    // it("can rename a folder inside the ipfs bucket", () => {
+    //   const ipfsBucketName = `ipfs bucket ${Date.now()}`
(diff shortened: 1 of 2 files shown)


juans-chainsafe:
@FSM1 Hi! what need to be tested here? do we have a ticket for this with more information?

FSM1:
> @FSM1 Hi! what need to be tested here? do we have a ticket for this with more information?

Unfortunately there is not a ticket for this change. Due to the performance issues with the IPFS (MFS) filesystem, we have taken a decision to disable this bucket type from creation. I see there are some storage tests failing, which I will go through and make sure that none of the tests depend on this file-system type.

tanmoyAtb (approved):
For the purpose at hand, looks good !

juans-chainsafe:
@FSM1 got it! I checked the UI and is ok, buuut I saw that you only commented one IPFS test, but we have a lot more, the thing on the others, is that they create an IPFS buckets using the API, so they are able to create the bucket and the tests passed. All of this in bucket-management.cy.ts I can comment those test as well if you want, but again, they are not failing because they use the API to create the ipfs bucket

FSM1:
> others, is that they create an IPFS buckets using the API, so they are able to create the bucket and the tests passed. All of this in bucket-management.cy.ts I can comment those test as well if you want, but again, they are not failing because they use the API to create the ipfs bucket

Im happy to have all those tests commented out for now.

juans-chainsafe:
> > others, is that they create an IPFS buckets using the API, so they are able to create the bucket and the tests passed. All of this in bucket-management.cy.ts I can comment those test as well if you want, but again, they are not failing because they use the API to create the ipfs bucket
> 
> Im happy to have all those tests commented out for now.

Done! tests related to IPFS were commented

juans-chainsafe (approved):

PR #2244, opened 2022-08-10
Hide NFTs page by default

FSM1:
Hid the NFT page from the nav bar - this is a temporary measure until the NFT code is ready for prime time. Pages for listing and creation are accessible if the url is entered manually.

### DIFF
--- a/packages/storage-ui/src/Components/Layouts/AppNav.tsx
+++ b/packages/storage-ui/src/Components/Layouts/AppNav.tsx
@@ -18,7 +18,7 @@ import {
   useLocation,
   KeySvg,
   CreditCardOutlinedSvg,
-  FileWithImageSvg
+  // FileWithImageSvg
 } from "@chainsafe/common-components"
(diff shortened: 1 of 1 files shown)


packages/storage-ui/src/Components/Layouts/AppNav.tsx
```diff
@@ -18,7 +18,7 @@ import {
   useLocation,
   KeySvg,
   CreditCardOutlinedSvg,
```
Tbaut: To make lint happy
```suggestion
  CreditCardOutlinedSvg
```

Tbaut (approved):
If you need a quick one, then it looks good, just a comment to fix linting

packages/storage-ui/src/Components/Layouts/AppNav.tsx
FSM1: thanks for spotting this. Fixed ✨

juans-chainsafe (approved):
Looks good! and working well

PR #2254, opened 2022-10-05
Fix "view folder content from search result" test

juans-chainsafe:
Test "can view folder content from search result" was failing all the time in the pipeline and sometimes when you test it locally.

I manage to fix it locally adding the refresh for the bucket when you go from bin to home after creating a folder via API (this was implemented in order to be able to see the folder in home).

I also added a validation to be sure that when you perform a search, you are in the correct url


packages/files-ui/cypress/support/utils/apiTestHelper.ts
```diff
@@ -135,8 +135,12 @@ export const apiTestHelper = {
...
         }
 
-        navigationMenu.binNavButton().click()
-        navigationMenu.homeNavButton().click()
+        cy.intercept("POST", "**/bucket/*/ls").as("refresh").then(() => {
+          navigationMenu.binNavButton().click()
+          cy.wait("@refresh")
+          navigationMenu.homeNavButton().click()
+          cy.wait("@refresh")
+        })
```
asnaith, line 143: Nice reliability improvement 👍

asnaith (approved):
Thanks for this @juans-chainsafe :)

FSM1 (approved):

PR #2259, opened 2023-01-10
Release - 2023-01-10

FSM1:
* add cloudflare deploy actions

### DIFF
--- a/.github/workflows/cf-deploy-files-ui-prod.yaml
+++ b/.github/workflows/cf-deploy-files-ui-prod.yaml
@@ -1,0 +1,40 @@
+name: CloudFlare Deploy - Files-UI PROD
+on:
+  push:
+    branches: ["prod"]
+  pull_request:
+    branches: ["prod"]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      deployments: write
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          cache: yarn
+          node-version: "16"
+      - run: yarn install --frozen-lockfile
+      - run: yarn build:files-ui
+        env:
+          REACT_APP_API_URL: "https://api.chainsafe.io/api/v1"
+          REACT_APP_AUTH0_CLIENT_ID: "kkdDcP8aE1O3qGdwjAV9ybu75nl1afVv"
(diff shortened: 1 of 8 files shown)


tanmoyAtb (approved):
all good !

asnaith:
@FSM1 I think the render bot needs to be replaced with something that posts the Cloudflare links for Files and Storage too?

FSM1:
> @FSM1 I think the render bot needs to be replaced with something that posts the Cloudflare links for Files and Storage too?

The Render message will not be displayed any more once that is disabled.

The Cloudflare preview URL is already attached to the PR
image

asnaith:
@FSM1 Ah ok, thanks. I didn't realize it was there. That does raise a couple of points: 

- I only see one URL for storage there and none generated Files. Perhaps it's only showing the last deployment? 
- Also perhaps it's worth calling the deployment environment "public staging" or something, rather than "production"?

FSM1:
> @FSM1 Ah ok, thanks. I didn't realize it was there. That does raise a couple of points:
> 
> * I only see one URL for storage there and none generated Files. Perhaps it's only showing the last deployment?

Yup unfortunately that notification message will only show 1 of the deployments (usually the first one). The actual addresses can be checked by clicking `details` on the relevant action run, and then inspecting the summary: https://github.com/ChainSafe/ui-monorepo/actions/runs/3884495642

One thing we could try is to add a step to the action yaml file to post a comment on the PR? Improving the messaging around where previews are deployed is definitely not something that should stop the deployment right now though.

> * Also perhaps it's worth calling the deployment environment "public staging" or something, rather than "production"?

Unfortunately Cloudfare does not allow renaming these environments. It regards the deployment's main branch as the production branch, and all others as preview branches.

asnaith:
> Yup unfortunately that notification message will only show 1 of the deployments (usually the first one). The actual addresses can be checked by clicking `details` on the relevant action run, and then inspecting the summary: https://github.com/ChainSafe/ui-monorepo/actions/runs/3884495642

@FSM1 Ah ok as long as we can get there, all good. cc @juans-chainsafe 

> One thing we could try is to add a step to the action yaml file to post a comment on the PR? Improving the messaging around where previews are deployed is definitely not something that should stop the deployment right now though.

For sure, just wanted to mention this at the point of the change 👍 

> Unfortunately Cloudfare does not allow renaming these environments. It regards the deployment's main branch as the production branch, and all others as preview branches.

Oh that's too bad, ok. It could confuse new people so will try and remember this.

PR #2261, opened 2023-02-17
update service worker and dependencies

FSM1:
This PR resolves an issue where users were unable to log in to Files using OAuth (In Brave, FF and Safari).

Updated service worker and redirect.html to latest version resolved the issue. Additionally a temporary workaround that we had been using for staging (using a quiknode ropsten node) is no longer necessary as the team have migrated all required infra to goerli.


Tbaut (approved):
Thank you!

juans-chainsafe (approved):
Test "can view folder content from search result" was skipped due to random issues when is running on the pipeline CC @asnaith 

PR looks good !

asnaith:
> Test "can view folder content from search result" was skipped due to random issues when is running on the pipeline CC @asnaith
> 
> PR looks good !

Ah yes! If it's not reliable and causing too much noise then it's a good call to take it out of the regular execution 👍. Thanks @juans-chainsafe

PR #2262, opened 2023-02-20
Release - 2023-02-20

FSM1:
closes #

---
Submission checklist: 

   

#### Layout
- [ ] Change looks good in the desktop web ui
- [ ] Change looks good in the mobile web ui

#### Theme
- [ ] Components / elements inspected in light mode
- [ ] Components / elements inspected in dark mode

#### Billing
- [ ] Change applied to both files-ui and storage-ui


asnaith:
Hey @FSM1, IIRC there’s no way to validate the OAuth stuff on staging URL’s but I don't think that should hold anything up as we know production is already having OAuth issues and there are no other changes that should impact anything else. 

It would be diligent to do some post-deployment validation though so I can do that once it goes live.

FSM1:
> Hey @FSM1, IIRC there’s no way to validate the OAuth stuff on staging URL’s but I don't think that should hold anything up as we know production is already having OAuth issues and there are no other changes that should impact anything else.
> 
> It would be diligent to do some post-deployment validation though so I can do that once it goes live.

The OAuth was tested in staging successfully [here](https://files-ui-stage.pages.dev). The prod preview will not work though, and yes we should definitely test this out once the PROD deployment goes through.

juans-chainsafe (approved):
Yes, I tested the OAuth on the PR that was merge into stage, I checked Firefox, Chrome and Brave for OAuth login with Google and Github

Tbaut (approved):

